PR #125, opened 2016-04-28
add initial version of the C++ API

mislavn:


swig/cpp/src/Sysrepo.h
```diff
@@ -0,0 +1,66 @@
...
+extern "C" {
+#include "sysrepo.h"
+}
+
+class Error
+{
+public:
+    size_t cnt;
+    const sr_error_info_t *info;
+    void log_stderr(sr_log_level_t log_level);
```
lukasmacko: I think this could be the class method, the Error instance is not needed to alter the log level

swig/cpp/src/Sysrepo.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+    if  (_val)
+    sr_free_val(_val);
+    _val = val;
+    cnt = 0;
+}
+
+sr_val_t **Value::Get()
+{
+    return &_val;
```
lukasmacko: This allows to modify `_val` without calling the code in setter

swig/cpp/src/Sysrepo.cpp
```diff
@@ -0,0 +1,214 @@
...
+    if (!_val) {
+        throw runtime_error(sr_strerror(SR_ERR_NOMEM));
+    }
+    return _val->dflt;
+}
+
+char *Value::get_binary()
+{
+    if (!_val || _val && _val->type != SR_BINARY_T) {
+        throw runtime_error(sr_strerror(SR_ERR_NOMEM));
```
lukasmacko: Not enough memory error is misleading in case of retrieving invalid type

swig/cpp/src/Session.cpp
```diff
@@ -0,0 +1,281 @@
...
+int Session::get_item(const char *xpath, Value* value)
+{
+    int ret = sr_get_item(_sess, xpath, value->Get());
+    if (ret != SR_ERR_OK) {
+        throw_excption(ret);
+    }
+    return ret;
+}
+
+int Session::get_items(const char *xpath, Value& value)
```
lukasmacko: I am not sure if I understad how should I access the values, if the function returns more than one value. Could you explain, how is it supposed to be used?

swig/cpp/src/Session.cpp
```diff
@@ -0,0 +1,281 @@
...
+	       	const char *submodule_name,  sr_schema_format_t format)
+{
+    int ret = sr_get_schema(_sess, module_name, revision, submodule_name, format, &schema.content);
+    if (ret != SR_ERR_OK) {
+        throw_excption(ret);
+    }
+    return ret;
+}
+
+int Session::get_item(const char *xpath, Value* value)
```
lukasmacko: Why has this method an argument of type "pointer to the Value" while the others have "a reference to Value"?

swig/cpp/src/Session.cpp
```diff
@@ -0,0 +1,281 @@
...
+#include "Session.h"
+#include "Sysrepo.h"
+
+extern "C" {
+#include "sysrepo.h"
+}
+
+using namespace std;
+
+void Session::throw_excption(int error)
```
lukasmacko: there is a typo: throw_excEption

swig/cpp/src/Connection.cpp
```diff
@@ -0,0 +1,42 @@
...
+extern "C" {
+#include "sysrepo.h"
+}
+
+using namespace std;
+
+Connection::Connection(const char *app_name, const sr_conn_options_t opts)
+{
+    int rc;
+    _app_name = (char *) app_name;
```
rastislavszabo: There is no need to store app_name (it only needs to be passed to sr_connect). Anyway, if you want to store it for any reason, you should duplicate the string, since it may get invalid though the lifetime of the Connection.

swig/cpp/src/Session.cpp
```diff
@@ -0,0 +1,281 @@
...
+        rc = sr_session_start_user(sys.get_conn(), user_name, _datastore, _opts, &_sess);
+        if (SR_ERR_OK != rc) {
+            goto cleanup;
+        }
+    }
+
+    return;
+
+cleanup:
+    throw_excption(SR_ERR_INIT_FAILED);
```
rastislavszabo: You should rather throw the exception matching with error stored in rc.

swig/lua/libsysrepoLua51.i
```diff
@@ -0,0 +1,77 @@
...
+#include "Sysrepo.h"
+
+%}
+
+/* Parse the header file to generate wrappers */
+%include "Connection.h"
+%include "Session.h"
+%include "Sysrepo.h"
+
+typedef enum sr_type_e {
```
rastislavszabo: Just a question: Is there any way to avoid the duplication of these enums, already defined in sysrepo.h? With such a duplication, the maintainability of the code drops and we risk that someone who will update it in sysrepo.h forgets to update it here.

swig/cpp/src/Session.cpp
```diff
@@ -0,0 +1,281 @@
...
+int Session::session_stop()
+{
+    int ret = sr_session_stop(_sess);
+    if (ret != SR_ERR_OK) {
+        throw_excption(ret);
+    }
+    return ret;
+}
+
+int Session::get_last_error(Error& err)
```
rastislavszabo: Just an idea: Isn't returning of of the error code unnecessary in methods like this, given that the method will throw an exception if the wrapped function won't succeed? Methods like this could return the output argument instead (the Error object in this case), which would also simplify the API.

swig/cpp/src/Session.cpp
```diff
@@ -0,0 +1,281 @@
...
+int Session::get_last_errors(Error& err)
+{
+    int ret = sr_get_last_errors(_sess, &err.info, &err.cnt);
+    if (ret != SR_ERR_OK) {
+        throw_excption(ret);
+    }
+    return ret;
+}
+
+int Session::list_schemas(Schema& schema)
```
rastislavszabo: Just an idea: Isn't returning of of the error code unnecessary in methods like this, given that the method will throw an exception if the wrapped function won't succeed? Methods like this could return the output argument instead (the Schema object in this case), which would also simplify the API.

swig/cpp/examples/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+add_executable(get_item_example get_item_example.cpp)
+add_executable(set_item_example set_item_example.cpp)
+add_executable(delete_item_example delete_item_example.cpp)
+
+target_link_libraries(get_item_example Sysrepo_cpp)
+target_link_libraries(set_item_example Sysrepo_cpp)
+target_link_libraries(delete_item_example Sysrepo_cpp)
+
+macro(INSTALL_EXAMPLE_YANG MODULE_NAME REVISION)
```
rastislavszabo: There is no need to install the YANG models and data again, since these are already installed once by the rules in /exeamples/ directory if examples are enabled.

swig/cpp/examples/yang/ietf-interfaces.data.xml
```diff
@@ -0,0 +1,30 @@
+<interfaces xmlns="urn:ietf:params:xml:ns:yang:ietf-interfaces">
```
rastislavszabo: There is no need for adding the YANG files and data files already present in /examples/yang/ directory

swig/cpp/examples/CMakeLists.txt
mislavn: Sorry I forgot about that, I meant to ask should the C and C++ examples be separate.
I will remove them now.

rastislavszabo:
please use devel-cpp branch for this pull request

PR #137, opened 2016-05-16
Add codecov.yml

stevepeak:
> RE https://github.com/codecov/support/issues/193

I'll comment back here when we can merge 👍


stevepeak:
Actually this will be default... so you don't need to merge 👍

PR #153, opened 2016-05-23
Sysrepocfg

milanlenco:
An initial implementation of the sysrepocfg tool.


src/executables/sysrepocfg.c
```diff
@@ -0,0 +1,775 @@
...
+    return;
+}
+
+/**
+ * @brief Reads complete content of a file referenced by the descriptor 'fd' into the memory.
+ * Caller is responsible for deallocation of the memory block returned through the output argument 'out'.
+ * Returns SR_ERR_OK in case of success, SR_ERR_INTERNAL otherwise.
+ */
+static int
+srcfg_read_file_content(int fd, char **out)
```
rastislavszabo: rather than loading whole file into memory, yo should use mmap

src/executables/sysrepocfg.c
```diff
@@ -0,0 +1,775 @@
...
+
+    buffer[cur] = '\0';
+    *out = buffer;
+    return SR_ERR_OK;
+}
+/**
+ * @brief Initializes libyang ctx with all schemas installed for specified module in sysrepo.
+ */
+static int
+srcfg_ly_init(struct ly_ctx **ly_ctx, const char *module_name)
```
rastislavszabo: Please add NULL checks for all arguments in all functions. It is a good practice to have them in each function - it can be very helpful by debugging. Have a look at macros in common/sr_helpers.h

src/executables/sysrepocfg.c
```diff
@@ -0,0 +1,775 @@
...
+#include <sys/wait.h>
+#include <fcntl.h>
+#include <ctype.h>
+#include <libyang/libyang.h>
+
+#include "sr_common.h"
+#include "client_library.h"
+
+#define EXPECTED_MAX_INPUT_FILE_SIZE  4096
+#define DEFAULT_TEXT_EDITOR           "vi"
```
rastislavszabo: When no editor is specified via env. variable or an argument, print an error message suggesting how it could be specified rather than opening in default editor.

PR #268, opened 2016-08-04
Event notifications tree API

milanlenco:


milanlenco:
Added unit tests for submodule dependencies (bugs fixed in libyang) + merged with "Ly ctx per module".
Note: update your libyang installation to get the unit tests to pass.

PR #301, opened 2016-08-18
Experimental memory management

milanlenco:
Do not merge this pull request, it is only for the feedback!

The pull request implements a custom memory management, targeted to avoid duplication of strings (xpath, string values) during the conversions to/from GPB messages. Currently it is fully integrated only into RPC processing path, and partly integrated into the rest of sysrepo. Also, as a part of this work, a new public API for more user-friendly construction of values and trees is introduced (see the second half of sr_experimental.h). It is backward compatible with the old way though (the two methods just should not be mixed). The benefits of the new memory management, however, are in effect on the client side only if the new API is used. Sysrepo will use the new API only.

The basic idea was to allocate memory values, trees and GPB messages from a linked-list of larger possibly pre-allocated memory blocks (so called sysrepo memory context). When a new object is created (object may be: GPB message, sysrepo value, array of sysrepo values, sysrepo tree, array of sysrepo trees) a new memory context is created (or a freed old one is reused). The object maintains a pointer to this context. The context maintains an object counter. When values/trees are converted to/from GPB message only a shallow copy is performed and object counter is increased. The two objects will share the same context. Since the context is not thread-safe, they should not be manipulated at the same time by different threads!!! When an object is de-allocated (for GPB one has to use sr_gpb_free instead of sr__gpb__free_unpacked), the counter is decreased by one. Once the counter drops to zero, the context is freed -- it may be kept in the thread-private space if the limits for the free space are not exceeded or completely destroyed otherwise.

Results of the performance testing (for RPC):

- the number of allocations dropped by 60% for values and 70% for trees
- the total amount of newly allocated memory dropped by 96.5% for values and 97% for trees
- the runtime decreased by 12% for values and 15% for trees (the new mem. management doesn't use any inter-thread synchronization primitives).
- generally, the tree API is faster (10% shorter runtime) than values API, but uses more allocations (15% more) -- this is the same regardless of memory management
- the memory footprint as measured by valgrind is two times higher that the average with the old memory management. This is because some freed memory is actually kept in thread-specific space for reuse. On the other hand, the footprint is very stable (not a series of peaks as it is currently) and changes only if there is a noticeable change in the average size of requests/responses (the amount of free memory to keep is calculated dynamically based on the recent history). Realistically, the OS memory management also maintains some freed blocks, suffers from internal fragmentation and doesn't pass freed memory block from one thread to another instantly to reduce the frequency of inter-thread synchronizations. Valgrind doesn't take into account any of this.

The final implementation (if approved) will not contain helper macros for counting allocations (added to sysrepo.h) and will be in something like sr_mem_mgmt.c and not sr_experimental.c as is now.


milanlenco:
Btw. there are methods to duplicate value(s) and tree(s) if one really needs to have a copy in a separate context.

inc/sysrepo.h
```diff
@@ -128,6 +154,9 @@ typedef union sr_data_u {
  * @brief Structure that contains value of an data element stored in the sysrepo datastore.
  */
 typedef struct sr_val_s {
+    /** Memory context used by this value. */
+    sr_mem_ctx_t *sr_mem;
```
rastislavszabo: I would name this as `_sr_mem` (with a `_` in the front) and explicitly state that this is an internal member (as in the .proto file)

inc/sysrepo.h
```diff
@@ -158,6 +187,9 @@ typedef struct sr_val_s {
  * than to an actual xpath.
  */
 typedef struct sr_node_s {
+    /** Memory context used by this node. */
+    sr_mem_ctx_t *sr_mem;
```
rastislavszabo: same as above - I would name this as `_sr_mem` (with a `_` in the front) and explicitly state that this is an internal member (as in the .proto file)

inc/sysrepo.h
```diff
@@ -172,15 +204,24 @@ typedef struct sr_node_s {
...
+    /** Pointer to the parent node (NULL in case of root node). */
+    struct sr_node_s *parent;
+
+    /** Pointer to the next sibling node (NULL if there is no one). */
+    struct sr_node_s *next;
+
+    /** Pointer to the previous sibling node (NULL if there is no one). */
+    struct sr_node_s *prev;
+
+    /** Pointer to the first child node (NULL if this is a leaf). */
```
rastislavszabo: maybe we could also provide some public iterator macro for convenient iterating over the children?

milanlenco:
Ready for merge.

PR #304, opened 2016-08-22
Shared ptr

mislavn:
Hi,

1) I had set the Session methods get_item, get_items, get_items_iter and get_item_next to return a new class so it is has a python like API as Lukas suggested it.

2) The method get_change_next returns a shared_ptr<Operation> which is a wrapper for sr_change_oper_t, this is an overkill in C++ but it generates better python bindings.

3) In the get_change_next and get_item_next methods i used the following lines:
    if (SR_ERR_OK == ret) {
        shared_ptr<> oper... //allocated class
        return oper; //return class
    } else if (SR_ERR_NOT_FOUND == ret) {
        return NULL;
    } else {
        throw_exception(ret);
        return NULL;
    }
}

4) I had to create a new Cmake in swig for other bindings, so now cmake generates binary Sysrepo_cpp for C++ API and Sysrepo_bindings for other bindings. Changes between the two are minimal and contained to callbacks functionality, they are visible with "ifdef SWIG".

 There are no memory leaks in C++ and python bindings work like C++ bindings.

Best regards,
Mislav Novakovic


lukasmacko:
Hi Mislav,

apply point 3 also to the function `sr_get_item` and `sr_get_items` to make get calls consistent
```C
if (SR_ERR_OK == ret) {
   shared_ptr<> oper... //allocated class
   return oper; //return class
} else if (SR_ERR_NOT_FOUND == ret) {
   return NULL;
} else {
   throw_exception(ret);
   return NULL;
}
````

Regards,
Lukas

File: swig/cpp/src/Session.h [hunk 1]
```diff
@@ -50,11 +51,11 @@ class Session:public Iter_Value, public Values
     void list_schemas(Schema& schema);
     void get_schema(Schema& schema, const char *module_name, const char *revision,
```
lukasmacko: It would be better if the function returned a string. I think filling the Schema object is misleading here. Could you change it?

mislavn: I am working on this, the only problem is that SWIG does not tolerate shared_ptr<char> and the caller has to free the memory so I had to create a new class (Schema_Content) for this.

With the shared_ptr functionality I will be able to simplify some functions:
`void get_last_error(Errors& err);` -> `shared_ptr<Errors> get_last_error();`

callbacks will now return classes instead sysrepo C structs.

I wanted to ask you is it better to have:
`shared_ptr<Schema_Content> get_schema(const char *module_name, const char *revision, const char *submodule_name, sr_schema_format_t format);`
or
`shared_ptr<Schema_Content> get_schema(const char *module_name, sr_schema_format_t format, const char *revision = NULL, const char *submodule_name = NULL);`

With the second API in python we would have:
`schema = sess.get_schema("ietf-interfaces", sr.SR_SCHEMA_YANG)`
instead
`schema = sess.get_schema("ietf-interfaces", None, None, sr.SR_SCHEMA_YANG)`

I will be done with the refactoring tomorrow, so the shared_ptr will be used where it is possible.
The other part I wanted to mention is that I will probably wrap the C structs in sysrepo.h to C++ objects because the C struct wrapping is inconsistent among binding languages in SWIG.

lukasmacko: I would suggest `shared_ptr<Schema_Content> get_schema(const char *module_name, const char *revision = NULL, const char *submodule_name = NULL, sr_schema_format_t format = SR_SCHEMA_YANG);`

mislavn:
Hi Lukas,
I did the changes as we discussed via email.

Please take a look at the API in C++ and python. I tried to make the API as consistent and logical as possible. If you have some suggestions please let me know.

The main difference now is that the classes are stateless, for every struct or arry the class wrapper will return a new class via shared pointers. This way it is easier to implement new C structs and update the API.

I did not yet managed to have shared_ptr<class> in the callabck, I got some wired behaviour but I think it is possible. I want to add this feature because then the code would be cleaner and more safe. I will probably do it tommorow.

Best regards,
Mislav Novakovic

swig/cpp/examples/cpp_application_changes_example.cpp
```diff
@@ -147,16 +148,13 @@ module_change_cb(sr_session_ctx_t *session, const char *module_name, sr_notif_ev
...
-            try {
-                sr_change_oper_t oper = subscribe.get_change_next(&it, &old_value, &new_value);
-                print_change(oper, &old_value, &new_value);
-            } catch( const std::exception& e ) {
-                break;
-            }
+        it = subscribe.get_changes_iter(&change_path[0]);
+
+        while (oper = subscribe.get_change_next(it, old_value, new_value)) {
+		if (old_value->val() && new_value->val())
```
lukasmacko: This will work only for SR_OP_MODIFIED. If a change operation is created old value is NULL, similarly when if it is deleted new value is NULL.

swig/cpp/src/Struct.cpp
```diff
@@ -0,0 +1,539 @@
...
+Uint16_val::~Uint16_val() {return;}
+Uint32_val::Uint32_val(int32_t data) {_data = data;}
+Uint32_val::~Uint32_val() {return;}
+Uint64_val::Uint64_val(int64_t data) {_data = data;}
+Uint64_val::~Uint64_val() {return;}
+
+// Data
+Data::Data(sr_data_t data, sr_type_t type) {_d = data; _t = type;}
+Data::~Data() {return;}
+shared_ptr<Char_val> Data::get_binary() {
```
lukasmacko: I am not sure if we misunderstood each other here or this was necessary because of memory control. I don't see benefit in having a class for each data type.  
I expected that:
- if the type matches (value type is `SR_BINARY_T` and `get_binary` method is used) that "primitive" value is returned. IMHO there is redundant `->get()` currently.
- if the incorrect type is requested an exception is thrown

However, if it is necessary for some reason or you find it more suitable I am OK with that.

swig/cpp/examples/cpp_application_changes_example.cpp
mislavn: I will fix this in the examples.

lukasmacko:
Hi Mislav,

Apart from two comments inline please take a look on OS X build. Once these things are resolved, it is ok to be merged.

Regards,
Lukas

swig/cpp/src/Struct.cpp
mislavn: I add this to C++ API so it would be more consistent. The main idea is that every getter in the API that tries to return a non existing data will return NULL/None. I expect that users will find it easier to debug variable of type NULL/None than a thrown exception.
But if you want I can change this to the old API, it is not a problem.

lukasmacko: I would prefer throwing exception over returning NULL because IMHO it can be misleading in some sitation. NULL might be incorrectly interpreted as 'not set' instead of incorrect method used. 

Previous solution is also less verbose ``value.get_uint8()` instead of `value.data().get_uint8().get()`

mislavn: OK, I will change this.

Do you have any suggestions on how to replicate a local build of OS X or how to locally build Travis. I tried to cross compile to llvm/gcc but I did not yet managed to get it working. It is hard to fix this without access to the compiler and linker.

After this pull request I will add some simple C++ unit tests based on your unit tests "tests/cl_test.c". I was thinking of writing simple unit tests to test memory leaks and API. At the moment I am not sure what unit test library to use.
If you have any opinions on C++ unit tests please let me know.

Based on C++ unit tests one could easily write python unit tests. The syntax is quite similar.

mislavn:
Hi Lukas,

please can you merge the pull request now. I did not manage to fix the clang warning for now.
I will fix it in the next pull request. 

Regards,
Mislav

PR #309, opened 2016-09-04
Callback

mislavn:


mislavn:
Hi Rastislav,

can you take a look at this pull request. I have a few questions/suggestions.

1) Python3 bindings are ready, I need to edit cmake for this. The problem is that cmake does support very well loading packages with multiple versions (Python2 and Python3 for an example). So I would like to have for every language a cmake variable. Something like:
set(SYSREPO_PYTHON "2").
Is this OK with you?

2) I am not sure what is the difference between tree and node, it uses the same struct. Do I need two classes for this?

3) I made a wrapper class for sr_xpath_ctx_t. Do you have any other plans with this struct besides described in the tests?

4) I add an rpc_subscribe example in python2 and C++. Do you know how to locally test this without installing the yang module "test-module".

swig/cpp/src/Tree.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    shared_ptr<Tree> node(new Tree(_node));
+    return node;
+}
+shared_ptr<Tree> Tree::parent() {
+    if (_node->parent == NULL)
+        return NULL;
+
+    shared_ptr<Tree> node(new Tree(_node->parent));
+    return NULL;
```
milanlenco: I think this should be "return node;", similarly for next, prev, first_child and last_child member functions.

swig/cpp/src/Tree.cpp
```diff
@@ -0,0 +1,299 @@
...
+        sr_free_tree(_node);
+}
+shared_ptr<Tree> Tree::dup() {
+    sr_node_t *tree_dup = NULL;
+    if (_node == NULL) return NULL;
+
+    int ret = sr_dup_tree(_node, &tree_dup);
+    if (ret != SR_ERR_OK) throw_exception(ret);
+
+    shared_ptr<Tree> dup(new Tree(tree_dup));
```
milanlenco: Duplicate should be freed as well as the original node, i.e.: 
shared_ptr<Tree> dup(new Tree(tree_dup), true);

swig/cpp/src/Tree.cpp
```diff
@@ -0,0 +1,299 @@
...
+    if (ret != SR_ERR_OK)
+        throw_exception(ret);
+
+    _free = true;
+    _node = node;
+}
+Tree::Tree(sr_node_t *tree, bool free) {_node = tree; _free = free;}
+Tree::~Tree() {
+    if (_node != NULL && _free)
+        sr_free_tree(_node);
```
milanlenco: The drawback of this design is that if the root node goes out of the scope while the tree is being iterated through, the internal (C) tree gets deleted and the iteration process will thus end with an exception. It is not likely, but the user might do something like:
tree = tree.first_child();  // original shared pointer gets deleted and the internal tree with it 
tree = tree.next();          // this will throw an exception (_node is invalid)

Keeping a (shared) pointer on the parent (C++) node might do the trick (then you wouldn't also need to create a new instance of the Tree struct in the parent method). Alternatively, you could implement your own reference counter for the internal tree as a whole and call sr_free_tree (on the root node!) when the counter drops to zero.

swig/cpp/src/Tree.h
```diff
@@ -0,0 +1,99 @@
...
+
+class Tree:public Throw_Exception
+{
+public:
+    Tree(const char *root_name, const char *root_module_name);
+    Tree(sr_node_t *tree, bool free = false);
+    shared_ptr<Tree> dup();
+    shared_ptr<Tree> node();
+    sr_node_t *tree() {return _node;};
+    sr_mem_ctx_t *_sr_mem() {return _node->_sr_mem;};
```
milanlenco: _sr_mem is meant to be and described as an internal member already in the C API so I wouldn't make it accessible through the C++ API.

swig/cpp/src/Xpath.h
```diff
@@ -0,0 +1,62 @@
...
+
+using namespace std;
+
+class Xpath_Ctx:public Throw_Exception
+{
+public:
+    Xpath_Ctx();
+    Xpath_Ctx(sr_xpath_ctx_t *state);
+    ~Xpath_Ctx();
+    char *next_node(char *xpath) {return sr_xpath_next_node(xpath, _state);};
```
lukasmacko: I think it would be nice if xpath arguments had default value = NULL and even better if it xpath was passed to constructor and then to the first call of Xpath_Ctx instance functions. 
```c++
Xpath_Ctx(char *xpath);
char *next_node() {return sr_xpath_next_node(get_xpath_internal(), _state);};
private:
char *get_xpath_internal() /* returns xpath provided to constructur for the first call, NULL for subsequent */
```
This would apply to all functions that has `xpath` and `state` argument.

swig/cpp/src/Tree.h
```diff
@@ -0,0 +1,99 @@
...
+#include "sysrepo/trees.h"
+}
+
+using namespace std;
+
+class Tree:public Throw_Exception
+{
+public:
+    Tree(const char *root_name, const char *root_module_name);
+    Tree(sr_node_t *tree, bool free = false);
```
milanlenco: This method should be private, we don't want to expose C structs from the C++ wrapper class. Similarly for the get() method.

swig/cpp/src/Tree.h
```diff
@@ -0,0 +1,99 @@
...
+    sr_node_t *_node;
+    bool _free;
+};
+
+class Trees:public Throw_Exception
+{
+public:
+    Trees();
+    Trees(size_t n);
+    Trees(sr_node_t **trees, size_t *cnt, size_t n);
```
milanlenco: Like for the Tree class, methods that expose C structs should be private, i.e. this constructor (which seems unnecessary), the constructor below, p_trees() and trees() methods.

swig/cpp/examples/rpc_example.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+#include "Session.h"
+
+#define MAX_LEN 100
+
+using namespace std;
+
+void
+print_tree(shared_ptr<Tree> tree)
+{
```
milanlenco: While not necessary in this particular example, trees in general may consist of multiple nodes, so for an example application it would be more appropriate to show how to print (and iterate) trees in their entirety, i.e. for example iterate recursively through all nodes in certain order and print some diagram.

lukasmacko:
Hi Mislav,

Rastislav is on PTO, he will reply later this week. 

For now, just briefly:
3) Yes, the tests should cover all intend usage of sr_xpath_ctx_t. Basicly that helpers were designed for developers integrating their apps with sysrepo aiming to ease the processing of xpath.

Regards,
Lukas

swig/cpp/examples/rpc_example.cpp
```diff
@@ -0,0 +1,277 @@
...
+        cout << "= " << value->data()->get_binary() << endl;
+        break;
+    default:
+        cout << "(unprintable)" << endl;
+    }
+    return;
+}
+
+int test_rpc_cb(const char *xpath, const sr_val_t *input, const size_t input_cnt, sr_val_t **output,\
+                size_t *output_cnt, void *private_ctx) {
```
milanlenco: I think we should create a C++ counterpart for rpc/event-notification callback types (both for values and trees) so that the user won't manipulate with sr_val_t and sr_node_t in a C++ program at all, but rather with a Val(s) and Tree(s). 
Also, in my opinion, it is more appropriate to implement callbacks as virtual methods in C++, but this is up for discussion.

milanlenco:
Hi Mislav,

I will answer 2):

Node is a structure representing a single node in a tree. It is recursive structure in the sense that is contains pointers to its children, siblings and also to its parent (for convenience only).
By "Tree" we understand a set of nodes connected through pointers forming a tree. So there is no separate structure for a tree.
Some functions manipulate with individual nodes (such as the function for setting a node name), these are prefixed with "sr_node_" , while some manipulate with complete trees (i.e. access pointers leading to children, siblings etc.), these are prefixed with "sr_tree_" and under all normal circumstances you would pass them pointer to a root node.
So your C++ wrapper code is correct, I have just added some comments regarding minor issues.

Thanks,
Milan

swig/cpp/src/Tree.cpp
mislavn: Fixed it.

swig/cpp/src/Tree.cpp
mislavn: I too was thinking of this. I wanted to know the relation between tree and node before I implemented it. I will use std::enable_shared_from_this to get the shared_ptr of the parent and pass it to children.

swig/cpp/src/Tree.h
mislavn: Fixed it.

swig/cpp/examples/rpc_example.cpp
mislavn: I tried to implement this in Python but I had some trouble with premature destructor call. I suspect the problem is in the SWIG permissions.
PyObject *s = SWIG_NewPointerObj(sess, SWIGTYPE_std__shared_ptrT_Session_t, 0 | SWIG_POINTER_ OWN);
This should be possible to fix.

I had in plan, when the new python callbacks work, to implement something similar in C++. So the API is the same in C++ and Python.
Is this OK for you? Or should the C++ and python API diverge already.

We can add virtual functions too, the only problem I have with them Is that you can have only one callback per class and I wanted to have similar API for C++/python and other bindings.

mislavn:
Hi,

I have some more questions about the callbacks in python/C++,  C++ constructor and destructors and the general API design.
Could we have a group chat tomorrow. I will be available after 9 pm.

Best regards,
Mislav Novakovic

lukasmacko:
Hi Mislav,

take a look into static analysis issues related to bindings

thanks,
Lukas

mislavn:
Hi,

I added the two commits we talked about:
1) how to remove Val_Holder class
2) How to create a class in the python callback which needs size allocation input from the user.

Regards,
Mislav

milanlenco:
Regarding destruction of objects, I think we should be able to do without using the "free" flag in the constructor. Shared pointer (or a custom usage counter) could be used with the internal C-object. This way, all C++ wrapper instances of the same value(s)/tree(s) would share the ownership rather than leaving the destruction up to one of them. This would prevent also from a potential error scenario in which the owner instance is destroyed, while some other wrapper instances remained.

Also regarding public/private access modifiers: The internal C structs should be completely hidden from the C++ user. Unless there is no way around it, I would make private all the constructors and methods with internal structs as arguments.

swig/cpp/src/Session.cpp
```diff
@@ -427,12 +427,14 @@ shared_ptr<Iter_Change> Subscribe::get_changes_iter(const char *xpath)
     }
 }
 
-shared_ptr<Operation> Subscribe::get_change_next(shared_ptr<Iter_Change> iter, shared_ptr<Val_Holder> new_value,\
-                                            shared_ptr<Val_Holder> old_value)
+shared_ptr<Operation> Subscribe::get_change_next(shared_ptr<Iter_Change> iter, shared_ptr<Val> new_value,\
```
lukasmacko: I was thinking about this and I believe that the cleanest solution would be to slightly modify the Operation class (renaming it to Change might also be suitable). Basically, I was thinking about adding  shared_ptr<Val> new_val and old_val into the Change class. 

In this case all output would be returned using return value and user wouldn't need to provide "output-like arguments".

```C++
shared_ptr<Change> Subscribe::get_change_next(shared_ptr<Iter_Change> iter)
````
Would do you think? Would it cause any issues?

Comment (mislavn) [inline swig/cpp/src/Session.cpp]: NO, This would be simple to implement. So the Changes class would return two Val classes as I understand?

Comment (lukasmacko) [inline swig/cpp/src/Session.cpp]: Yes, all output arguments in C api operation, new_val, old_val would be members of Change class.

File: swig/python2/examples/rpc_example.py [hunk 11]
```diff
@@ -62,15 +62,14 @@ def print_value(value):
...
-        in_vals = sr.Vals(input_vals, input_cnt)
-        out_vals = sr.Vals(output_vals, output_cnt, 3)
-
         if (in_vals == None or out_vals == None):
             return
 
+        out_vals.allocate(3)
+
         for n in range(in_vals.val_cnt()):
             print_value(in_vals.val(n))
```
milanlenco: I think that for Python a more suitable solution would be to return the output using the return statement and to avoid using output-like arguments. So something like:

def test_rpc_cb(xpath, in_vals, private_ctx)"
    out_vals = Vals(3)
    out_vals.val(0).set("/ietf-interfaces:interfaces/interface[name='eth0']", SR_LIST_T)
    out_vals.val(1).set("/ietf-interfaces:interfaces/interface[name='eth0']", SR_LIST_T)
    out_vals.val(2).set("/ietf-interfaces:interfaces/interface[name='eth0']", SR_LIST_T)

    return out_vals

Would it be possible to achieve it using C++ & SWIG?

mislavn: I am looking into this, but i think it should be possible.

mislavn: Regarding the counter in the C structs, what should I do for now?
Wait for the counter in the structs or implement the method with the "free" flag.

I agree with the philosophy of keeping the Constructors private if they have an sysrepo C struct. The user should be able to create a new class or an empty one.

milanlenco: _Regarding the counter in the C structs, what should I do for now?
Wait for the counter in the structs or implement the method with the "free" flag._

I didn't mean to make a usage counter inside the internal structs, but to add some between them and the wrapper class, i.e. on the C++ level. So you could use shared_ptr for the task, or, if you experience any issues, you could possibly create your own usage counter tailor made for the C++ to Python conversion using SWIG.

mislavn:
Hi guys,

I have made the changes as you requested.
The sr_node_t and sr_val_t use a custom Counter class. Now the memory allocation is decided before the class is created if it constructor accepts a pointer to the class.

All of the callbacks do not return a sysrepo struct but a class wrapper, this is for C++ and python.

For the shared_ptr<class> I used a define because of inconsistent support for std::shared_ptr in SWIG, some have better support for boost library. It will be helpful for me when I do the Lua bindings.

I still have some memory leaks, it is a little hard to debug now with the changes from libyang. I will fix the memory leaks when the library stabilizes. For now I was not able to test all edge cases.

Best regards,
Mislav Novakovic

milanlenco:
Hi Mislav,

I find the Counter class very misleading. What I had in mind was to wrap the internal C objects with a shared pointer OR with a custom access counter if there are any issued with the shared pointer in SWIG. From the first glance at the code I though that you did both: wrapped the C object with your Counter class and then the Counter class with the shared pointer. But it turned out that the Counter is not actually counter but a custom destructor for the object passed to shared_ptr. Cleaner solution would be to use Deleter with shared_ptr:

template< class Y, class Deleter >
shared_ptr( Y* ptr, Deleter d );

and to name the deleter accordingly.

Also you still haven't modified callbacks to return values/trees through the return statement but kept the use of the "allocate" method on the input argument which will not look very Python-ish once translated with SWIG. Are there any issued with returning objects using "return" in SWIG?

Thanks,
Milan

mislavn:
Hi Milan,

OK, I can create a Deleter class.

For the callbacks I did not find any solution which would satisfy you. The Python callback can return a pointer to the C++ bindings but because it is a C callback wrapper it must not free it self when the shared_ptr goes out of scope. That means there must be a constructor which will allocate memory but not free itself for Vals(size_t n) and Trees(size_t n). Second option is to have a class method which will return a duplicate of the class but which will not free the memory (return Vals.duplicate() )
Third option is to have the method allocate(size_t n) method or a new wrapper class in the callback signature.
I am not sure if Python will allow to return a PyObject's if not I will have to cast void pointers to class pointer and I think that would lead to some segfaults in the C++ bindings if the user is not careful in the return statement.

So for me the only two options is to have a method allocate(size_t n) or to have wrapper class for Vals and Trees. The second best thing I did was to make the class methods rpc_send() and rpc_send_tree() to return a shared_ptr<> object.

I am not sure if you got my point.


Regards,
Mislav Novakovic

PR #323, opened 2016-09-13
SWIG: add python3 bindings

mislavn:


mislavn:
Hi,

here are the python3 bindings with the new cmake changes.

I did only superficial tests because netopeer2 is out of sync with the current libyang devel branch. I can run the python3 examples but I did not tests will it properly show changes in the datastore but it should work. When netopeer2 compiles with the current libyang devel branch I will properly test it.

Regards,
Mislav

PR #330, opened 2016-09-19
SWIG: add lua 5.2 bindings

mislavn:

### DIFF
--- a/swig/lua52/CMakeLists.txt
+++ b/swig/lua52/CMakeLists.txt
@@ -1,0 +1,45 @@
+add_definitions(-DSWIG)
+add_definitions(-DSWIGLUA)
+
+project(Sysrepo_Lua)
+
+add_library(Sysrepo_Lua SHARED ${SYSREPO_CPP_SOURCES})
+
+# include custom Modules
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/CMakeModules/")
+
+include_directories(${CMAKE_SOURCE_DIR})
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src)
+target_link_libraries(Sysrepo_Lua sysrepo)
+
+if(NOT BIN_INSTALL_DIR)
+	set(BIN_INSTALL_DIR ${CMAKE_INSTALL_PREFIX}/bin)
+endif()
+
+set(LUA_SWIG_BINDING_52 libsysrepoLua52)
+
+if (SWIG_FOUND)
+    include(${SWIG_USE_FILE})
+    find_package(Lua 5.2 REQUIRED)
+endif()
+
+if (SWIG_FOUND AND LUA_FOUND)
+    include(${SWIG_USE_FILE})
+    include_directories(${LUA_INCLUDE_DIR})
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+    set(CMAKE_SWIG_FLAGS "-c++")
+    set(CMAKE_SWIG_OUTDIR ${CMAKE_BINARY_DIR})
+    set(CMAKE_SWIG_FLAGS "-I${PROJECT_SOURCE_DIR}/inc")
+    set(CMAKE_SWIG_OUTDIR ${CMAKE_CURRENT_BINARY_DIR})
+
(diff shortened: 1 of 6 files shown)


mislavn:
Hi,

the build fails because the system is missing lua 5.2.
If you are using ubuntu you can install it with:
`      apt-get update && apt-get install -y \
      default-jre \
      default-jdk
`
Regards,
Mislav

rastislavszabo:
Hi Mislav,

the build cannot fail because of missing libraries. if lua 5.2 is not present on the system, bindings generation for lua 5.2 should be skipped and build should succeed.

Thanks,
Rastislav

mislavn:
Hi Rastislav,

I will update the commit so that by default Lua 5.1 is build, as it was before this pull request.

Regards,
Mislav

swig/lua52/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+if(NOT BIN_INSTALL_DIR)
+	set(BIN_INSTALL_DIR ${CMAKE_INSTALL_PREFIX}/bin)
+endif()
+
+set(LUA_SWIG_BINDING_52 libsysrepoLua52)
+
+if (SWIG_FOUND)
+    include(${SWIG_USE_FILE})
+    find_package(Lua 5.2 REQUIRED)
```
rastislavszabo: I think that all you need to do is to remove REQUIRED from here (similarly as you have it for lua51). In that case the build will pass in case that lua52 is not found.

mislavn: I do not want that the lua5.2 bindings are built with the lua5.1 library, that might confuse the users later on when they use the bindings.

I was actually thinking of adding a check in main cmake file. Something like:
lua:
selected version:
supported versions:
available versions:

And if there is a mismatch I would change the selected version and display a message.
I could do the same for python.

What do you think?

rastislavszabo: > I do not want that the lua5.2 bindings are built with the lua5.1 library

I agree, but changing `find_package(Lua 5.2 REQUIRED)` to `find_package(Lua 5.2)` would not allow it, would it? You can also check `LUA_VERSION_STRING` or `LUA_VERSION_MAJOR` and `LUA_VERSION_MINOR` in the branch below.

> I was actually thinking of adding a check in main cmake file. Something like:
lua:
selected version:
supported versions:
available versions:
And if there is a mismatch I would change the selected version and display a message.
I could do the same for python.

Yes, that is a good idea. In any case - if there is anything missing (python2/python3/lua51/lua52), the build of sysrepo must always succeed - we should just just print a warning.

mislavn: > I agree, but changing find_package(Lua 5.2 REQUIRED) to find_package(Lua 5.2) would not allow it, would it? You can also check LUA_VERSION_STRING or LUA_VERSION_MAJOR and LUA_VERSION_MINOR in the branch below.

In theory it should not build it but I encountered on stack overflow that sometimes find_package would find another version if the specified was not found. I will remove it when I add the cmake checks for library versions because it will be obsolete.  It should be finished soon.

mislavn: Hi Rastislav,
can you take a look at the new commit with the proposed changes.
Regards,
Mislav

CMakeLists.txt
```diff
@@ -9,12 +9,64 @@ set(SYSREPO_MICRO_VERSION 10)
...
+    find_package(SWIG)
+    if(NOT SWIG_FOUND)
+        message(WARNING "SWIG library not found")
+    else()
+        include(${SWIG_USE_FILE})
+    endif()
+endif()
+
+# find Lua package
+if(GEN_LUA_BINDINGS AND SWIG_FOUND)
```
rastislavszabo: Mislav, please, just to make the main CMakeLists little bit better readable (it becomes too large): could we move all this new stuff (or better all swig-related stuff) to the `swig/CMakeLists.txt` file? The only thing that I would leave here would be variables - GEN_LANGUAGE_BINDINGS, GEN_PYTHON_BINDINGS, etc.

mislavn: At the start I did the same thing but then I moved it to the main CMakeLists because of the python bindings in them.

Now I moved it to the swig/CmakeLists file and in the main CMakeLists add find_package for its python bindings.

Regards,
Mislav

PR #332, opened 2016-09-20
FIX 	issues #331 Segmentation fault when use sysrepoctl -L parameter

ZhaoGuangYue:
FIX 	issues #331 Segmentation fault when use sysrepoctl -L parameter


rastislavszabo:
Hi,
this should be already fixed in the devel branch - will be merged into master with the next few weeks.

PR #350, opened 2016-09-30
,Java cmake

mislavn:
Hi,

I add the java bindings. I made some changes to the camke structure inside swig so to minimize the number of language specific macros, for now only LUA macro is left.

When I start the work on python testes I will test other language bindings and update them.

In the main cmake file I added the line:
set(BUILD_CPP_EXAMPLES 1 CACHE BOOL "Build C++ examples.")

Regards,
Mislavs


mislavn:
Hi Rastislav,

Can you take a look at the build, it breaks on:

Swig source
/home/travis/build/sysrepo/sysrepo/swig/cpp/src/Session.h:119: Error: Syntax error in input(3).

The line in the header is:
std::vector<S_Callback> cb_list;

I am no sure why and I can not replicate this error locally.
I used vectors in C++ before and it did not break.

Regards,
Mislav

lukasmacko:
Hi Mislav,

just guessing ...because of `#define S_Callback std::shared_ptr<Callback>` there are two angle brackets without space between them and that might be causing the problem according to the [link](http://swig.10945.n7.nabble.com/Syntax-error-in-input-3-when-using-vectors-of-complex-numbers-td13819.html)
so changing the line to be like `std::vector<S_Callback > cb_list;` might help

Regards,
Lukas

mislavn:
Hi Lukas,

thanks for the fix.
Just so you know I will add java examples to the pull request then update it.

Regards,
Mislav

mislavn:
Hi Rastislav,

I added the changes to the pull request.
At the moment I set GEN_JAVA_BINDINGS to 0 because travis breaks on:

In file included from /home/travis/build/sysrepo/sysrepo/build/swig/java/libsysrepoJavaJAVA_wrap.cxx:160:0:
/usr/lib/jvm/java-8-oracle/include/jni.h:45:20: fatal error: jni_md.h: No such file or directory
 #include "jni_md.h"
                    ^

When I fix the coverity issues in CPP bindings then I will fix this also, it is simply more convenient then.

Regards,
Mislav

Issue #366, opened 2016-10-04
segfault in swig/cpp/examples/rpc_example due to a use-after-free


jktjkt:
Looks like ``Trees`` is mixing manual memory management with ``std::shared_ptr`` which won't work. In default build this still degfaults, but sysrepo's custom memory management makes debugging harder, so this is a result of the same operation with ``USE_SR_MEM_MGMT=OFF`` and compiled with clang 3.9.0 with ASAN.

```
jkt@kolibrik ~ $ ASAN_OPTIONS=detect_odr_violation=0 LD_LIBRARY_PATH=/opt/libnetconf/lib ./work/prog/sysrepo/build/swig/cpp/examples/rpc_example
[...]
 ========== SUBSCRIBE TO RPC TREE CALL ==========

=================================================================
==23507==ERROR: AddressSanitizer: heap-use-after-free on address 0x607000027718 at pc 0x7f316d807ee6 bp 0x7fff040b7920 sp 0x7fff040b7918
WRITE of size 8 at 0x607000027718 thread T0
    #0 0x7f316d807ee5 in Tree::set(char const*, sr_type_e) /home/jkt/work/prog/sysrepo/build/../swig/cpp/src/Tree.cpp:133:29
    #1 0x5198e9 in main /home/jkt/work/prog/sysrepo/build/../swig/cpp/examples/rpc_example.cpp:245:28
    #2 0x7f316b2f6733 in __libc_start_main /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/csu/libc-start.c:289
    #3 0x41e438 in _start (/home/jkt/work/prog/sysrepo/build/swig/cpp/examples/rpc_example+0x41e438)

0x607000027718 is located 24 bytes inside of 80-byte region [0x607000027700,0x607000027750)
freed by thread T0 here:
    #0 0x4d9710 in free /var/tmp/portage/sys-devel/llvm-3.9.0/work/llvm-3.9.0.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:47
    #1 0x7f316d445a2a in sr_free_tree /home/jkt/work/prog/sysrepo/build/../src/common/sr_common.c:133:13
    #2 0x7f316d80bb6b in Counter::~Counter() /home/jkt/work/prog/sysrepo/build/../swig/cpp/src/Internal.cpp:92:20
[...]

PR #367, opened 2016-10-04
cpp: Ensure that Trees::tree reuses our reference counter

jktjkt:
The code uses Counter instances to work around the std::shared_ptr API
design where one cannot really manipulate the reference count from
outside. In this context, we however need to ensure that a
shared_ptr<Tree> created from within a Trees context reuses the same
counter, and that the underlying data structure only gets deleted when
the enclosing Trees structure is destroyed.

@mislavn , please check issue #366 and ensure that this patch doesn't make the situation worse.

### DIFF
--- a/swig/cpp/src/Tree.cpp
+++ b/swig/cpp/src/Tree.cpp
@@ -304,8 +304,7 @@ Trees::~Trees() {return;}
 S_Tree Trees::tree(size_t n) {
     if (_trees == NULL || n >= _cnt) return NULL;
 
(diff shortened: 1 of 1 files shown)


rastislavszabo:
Hi Mislav,
what about this fix: https://github.com/sysrepo/sysrepo/pull/367/commits/b04250d38f7b5b44cddec090b38853cec42cccca
Should it be merged, or does your fix make it redundant?

rastislavszabo:
fixed in https://github.com/sysrepo/sysrepo/pull/375
Thanks!

PR #375, opened 2016-10-05
Cpp update

mislavn:
Hi Rastislav,

here is the patch for the C++ bug.

For simpler maintenance I changed back to the helper class Trees_Holder and Vals_Holder which have the method allocate instead of having the method inside the classes Vals and Trees.

The classes are used in the methods rpc_cb and rpc_tree_cb.

I have updated the python2 example rpc_example.py too and some other fixes in C++.

Regards,
Mislav


rastislavszabo:
Hi Mislav,
what about this fix: b04250d
Should it be merged, or does your fix make it redundant?

mislavn:
Hi Rastislav,

The  fix: b04250d is included in the pull request.
I updated the java bindings also.

I rebased my sysrepo fork to the newest devel branch and it breaks now on:

/opt/fork/sysrepo/src/request_processor.c: In function 'rp_commit_timeout_req_process':
/opt/fork/sysrepo/src/request_processor.c:2155:54: error: 'Sr__CommitTimeoutReq' has no member named 'expired'
             msg->internal_request->commit_timeout_req->expired);
                                                      ^

Regards,
Mislav

PR #389, opened 2016-10-06
Remove C ODR violation

jktjkt:
Because both libsysrepo.so and libsysrepo-engine.so were linking objects
from COMMON, the protobuf-c's sr__value__descriptor variable was present
twice in the resulting program. That's forbidden.

Thanks to ASAN for showing this:

```
  =================================================================
  ==31698==ERROR: AddressSanitizer: odr-violation (0x7ffbf3b588a0):
    [1] size=120 'sr__value__descriptor' src/common/sysrepo.pb-c.c:3772:34
    [2] size=120 'sr__value__descriptor' src/common/sysrepo.pb-c.c:3772:34
  These globals were registered at these points:
    [1]:
      #0 0x42b18d in __asan_register_globals.part.14 /var/tmp/portage/sys-devel/llvm-3.9.0/work/llvm-3.9.0.src/projects/compiler-rt/lib/asan/asan_globals.cc:309
      #1 0x7ffbf381997d in asan.module_ctor (/opt/libnetconf/lib/libsysrepo.so+0x8097d)
      #2 0x7ffbf3bb111b in call_init /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/elf/dl-init.c:30
      #3 0x7ffbf3bb111b in _dl_init /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/elf/dl-init.c:120
      #4 0x7ffbf3ba1cb9  (/lib64/ld-linux-x86-64.so.2+0xcb9)

    [2]:
      #0 0x42b18d in __asan_register_globals.part.14 /var/tmp/portage/sys-devel/llvm-3.9.0/work/llvm-3.9.0.src/projects/compiler-rt/lib/asan/asan_globals.cc:309
      #1 0x7ffbf2d7218d in asan.module_ctor (/opt/libnetconf/lib/libsysrepo-engine.so+0xcd18d)
      #2 0x7ffbf3bb111b in call_init /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/elf/dl-init.c:30
      #3 0x7ffbf3bb111b in _dl_init /var/tmp/portage/sys-libs/glibc-2.22-r4/work/glibc-2.22/elf/dl-init.c:120
      #4 0x7ffbf3ba1cb9  (/lib64/ld-linux-x86-64.so.2+0xcb9)

  ==31698==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
  SUMMARY: AddressSanitizer: odr-violation: global 'sr__value__descriptor' at src/common/sysrepo.pb-c.c:3772:34
  ==31698==ABORTING
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -98,24 +98,16 @@ if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 endif()
 
 # sysrepo library
-add_library(sysrepo SHARED $<TARGET_OBJECTS:COMMON> $<TARGET_OBJECTS:SR_SRC>)
-add_library(sysrepo_a $<TARGET_OBJECTS:COMMON> $<TARGET_OBJECTS:SR_SRC>)
+add_library(sysrepo SHARED $<TARGET_OBJECTS:COMMON> $<TARGET_OBJECTS:SR_ENGINE> $<TARGET_OBJECTS:SR_SRC>)
+add_library(sysrepo_a $<TARGET_OBJECTS:COMMON> $<TARGET_OBJECTS:SR_ENGINE> $<TARGET_OBJECTS:SR_SRC>)
 SET_TARGET_PROPERTIES(sysrepo_a PROPERTIES
               OUTPUT_NAME sysrepo CLEAN_DIRECT_OUTPUT 1)
-target_link_libraries(sysrepo ${LINK_LIBRARIES} sysrepo-engine)
-target_link_libraries(sysrepo_a ${LINK_LIBRARIES} sysrepo-engine_a)
-
-# sysrepo-engine library
-add_library(sysrepo-engine SHARED $<TARGET_OBJECTS:COMMON> $<TARGET_OBJECTS:SR_ENGINE>)
-add_library(sysrepo-engine_a $<TARGET_OBJECTS:COMMON> $<TARGET_OBJECTS:SR_ENGINE>)
-SET_TARGET_PROPERTIES(sysrepo-engine_a PROPERTIES
-              OUTPUT_NAME sysrepo-engine CLEAN_DIRECT_OUTPUT 1)
-target_link_libraries(sysrepo-engine_a ${LINK_LIBRARIES})
-target_link_libraries(sysrepo-engine ${LINK_LIBRARIES})
+target_link_libraries(sysrepo ${LINK_LIBRARIES})
+target_link_libraries(sysrepo_a ${LINK_LIBRARIES})
 
 # sysrepo daemon
 add_executable(sysrepod ${EXECUTABLES_DIR}/sysrepod.c)
-target_link_libraries(sysrepod sysrepo-engine)
+target_link_libraries(sysrepod sysrepo)
 
 # sysrepo plugin daemon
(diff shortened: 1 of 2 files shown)


jktjkt:
Hmm, this appears to break the build for ``netopeer2-server``:

```
jkt@kolibrik ~/work/prog/netopeer2/build-server[master] $ ninja -j1
[1/5] Linking C executable netopeer2-server
FAILED: : && /usr/bin/clang  -fsanitize=address -Wall -Wextra -g -O0 -DDEBUG  -fsanitize=address CMakeFiles/serverobj.dir/ietf_netconf_server.c.o CMakeFiles/serverobj.dir/netconf_monitoring.c.o CMakeFiles/serverobj.dir/operations.c.o CMakeFiles/serverobj.dir/op_get_config.c.o CMakeFiles/serverobj.dir/op_editconfig.c.o CMakeFiles/serverobj.dir/op_copyconfig.c.o CMakeFiles/serverobj.dir/op_deleteconfig.c.o CMakeFiles/serverobj.dir/op_candidate.c.o CMakeFiles/serverobj.dir/op_validate.c.o CMakeFiles/serverobj.dir/op_un_lock.c.o CMakeFiles/serverobj.dir/op_generic.c.o CMakeFiles/serverobj.dir/log.c.o CMakeFiles/netopeer2-server.dir/main.c.o  -o netopeer2-server  -rdynamic -lpthread /opt/libnetconf/lib/libyang.so /opt/libnetconf/lib/libnetconf2.so /opt/libnetconf/lib/libsysrepo.so -Wl,-rpath,/opt/libnetconf/lib: && :
/usr/bin/x86_64-pc-linux-gnu-ld: CMakeFiles/serverobj.dir/ietf_netconf_server.c.o: undefined reference to symbol 'sr_strerror'
/opt/libnetconf/lib/libsysrepo-engine.so: error adding symbols: DSO missing from command line
x86_64-pc-linux-gnu-clang-3.9: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

Are 3rd party projects expected to link against just ``sysrepo``, or against both that and ``sysrepo-engine``?

jktjkt:
OK, fixed, I think. I won't pretend that I'm an expert on library design, so I wonder if it's OK to require all downstream consumers to either use pkg-config, or to manually link with both of these two libraries.

CESNET/netopeer2's ``FindSYSREPO.cmake`` appears to be broken because it uses pkg-config to find the library locations, but it doesn't use the returned list of libs.

rastislavszabo:
Hi,

we actually don't want 3rd party projects to link against both `sysrepo` and `sysrepo-engine`. We intended `sysrepo` to be the client library to link by 3rd party projects and `sysrepo-engine` to be the engine library linked to sysrepo deamon. But since the client library is also capable of initiating its own sysrepo engine in case that daemon is not running, `sysrepo-engine` must be linked to the `sysrepo` library :)

As it complicates things and does not bring any additional value, I think that the best option would be to create only one `sysrepo` library that would include all object files.

jktjkt:
> As it complicates things and does not bring any additional value, I think that the best option would be to create only one sysrepo library that would include all object files.

Done in 111a0884d04138a30c2062629c54d62a2d373cbd ; you might need to nuke the build dir and destdir for this, though. ``ninja clean`` was not enough for me.

PR #405, opened 2016-10-11
Continue in request processing even if state data retrieval fails

### LINKED ISSUE #395: sr_get_items_iter failed for state leaf after sr_dp_get_items_subscribe
code:
```
int main(void)
{
    sr_conn_ctx_t *sr_conn;
    sr_session_ctx_t *srs; /* SYSREPO session */
    int rc;
    sr_val_iter_t *sriter;
    sr_subscription_ctx_t *subscription = NULL;

    sr_log_set_cb(log_for_sr);

    /* connect to the sysrepo */
    rc = sr_connect("test", SR_CONN_DEFAULT, &sr_conn);
    if (rc != SR_ERR_OK) {
        printf("Unable to connect to sysrepo (%s).\n", sr_strerror(rc));
        return -1;
    }

    rc = sr_session_start(sr_conn, SR_DS_RUNNING, SR_SESS_DEFAULT, &srs);
    if (rc != SR_ERR_OK) {
        printf("Unable to create Netopeer session with sysrepod (%s).\n", sr_strerror(rc));
        return -2;
    }

    rc = sr_get_items_iter(srs, "/a:c1/f1//*", &sriter);
    if (rc != SR_ERR_OK) {
        printf("Failed to get item iter before subscribe.\n", sr_strerror(rc));
        return -3;
    }

    rc = sr_dp_get_items_subscribe(srs, "/a:c1", get_items_clb, 0, SR_SUBSCR_CTX_REUSE, &subscription);
    if (rc != SR_ERR_OK) {
        printf("Failed to subscribe '/a:c1' get-items (%s).\n", sr_strerror(rc));
        return -4;
    }

    rc = sr_get_items_iter(srs, "/a:c1/f1//*", &sriter);
    if (rc != SR_ERR_OK) {
        printf("Failed to get item iter after subscribe.\n", sr_strerror(rc));
        return -5;
    }

    sr_disconnect(sr_conn);

    printf("Succeed to do all.\n");

    return 0;
}
```
a.yang
```
module a{
  namespace "aaa:1";
  prefix "a";

  container c1{
    leaf f1{
      type uint32;
      config false;
    }
  }
}
```
I know that we should not use `sr_get_items_iter(srs, "/a:c1/f1//*", &sriter);` to get children of a leaf, however, it is not very well to return error only after subscribe, and netopeer2 did it like this when you execute `get --filter-xpath /a:c1/f1`.

lukasmacko:
fix #395

### DIFF
--- a/src/rp_dt_get.c
+++ b/src/rp_dt_get.c
@@ -802,7 +802,10 @@ rp_dt_prepare_data(rp_ctx_t *rp_ctx, rp_session_t *rp_session, const char *xpath
 
             rc = rp_dt_xpath_requests_state_data(rp_ctx, rp_session, data_info->schema, xpath, api_variant,
                     tree_depth_limit, &rp_session->state_data_ctx);
-            CHECK_RC_MSG_GOTO(rc, cleanup, "rp_dt_xpath_requests_state_data failed");
+            if (SR_ERR_OK != rc) {
+                SR_LOG_WRN_MSG("rp_dt_xpath_requests_state_data failed - state data won't be retrieved");
+                rc = SR_ERR_OK;
+            }
 
             if (NULL == rp_session->state_data_ctx.subtrees || 0 == rp_session->state_data_ctx.subtrees->count) {
                 SR_LOG_DBG("No state state data provider is asked for data because of xpath %s", xpath);


fanchanghu:
Hi,
I do't think it's a good idea to ignore error of `rp_dt_xpath_requests_state_data`, this may ignore some real error. How about making `rp_dt_xpath_requests_state_data` return `SR_ERR_OK` and `rp_session->state_data_ctx.subtrees->count = 0` when result is empty and xpath is syntactically ok?

lukasmacko:
yes you're right it can be done better see https://github.com/sysrepo/sysrepo/pull/411

PR #410, opened 2016-10-11
Swig refactor

mislavn:
Hi Rastislav,

In this pull request I refactored the swig files, remove Throw_Exceptions class and add install taregt for lua.

Regards,
Mislav


swig/cpp/src/Connection.h
```diff
@@ -31,10 +31,10 @@ extern "C" {
 #include "sysrepo.h"
 }
 
-class Connection:public Throw_Exception
+class Connection
 {
 public:
-    Connection(const char *app_name, const sr_conn_options_t opts = SESS_DEFAULT);
+    Connection(const char *app_name, const sr_conn_options_t opts = 0);
```
jktjkt: What is the motivation behind this change? As an API user, I have no idea what ``0`` means. Given that the code already requires C++11, have you considered going with scoped enums (``enum class``)? (And "something" for flags, yeah...)

mislavn: It is because of the swig bindings, if I put SR_SESS_DEFAULT python bindings would not recognize it at runtime.

I had some difficulties generating wrapper code for the C structs and at the same time using them as default values in function overloading, so I created my custom data type based on sysrepo ones.

Since almost all of the default values use the first element ( value 0) I wanted to remove my custom enums which are used only for function overloading.

@jktjkt  you do make a valid point, I could use typedef or a define just for the default value that would make it more readable. At the moment I am not sure how swig will behave I will check this tomorrow.

mislavn: The pull request has been updated.

swig/cpp/src/Sysrepo.h
```diff
@@ -82,6 +82,13 @@
     #define S_Callback         std::shared_ptr<Callback>
 #endif
 
+#define SESS_DEFAULT 0
```
jktjkt: Well, my biggest gripe with this is that this is not type-safe. I understand your point about SWIG generators and your approach where C++ is used mainly as a vehicle for other OO-oriented languages.

I'll be happy to help in designing something which works well under both SWIG and C++ once my copany-internal requests make it besides $legal, you know.

mislavn: That would be great, I'm looking forward to it.

Issue #379, opened 2016-10-05
segfault in swig/cpp/examples/rpc_example


jktjkt:
Buiilt on CentOS7 with either clang 4.0.0 from today's git, or using GCC 5.2.1 from the devtoolset-4 package. ASAN is active, custom memory management disabled:

```
$ ./swig/cpp/examples/rpc_example
Application will make an rpc call in test-module

 ========== SUBSCRIBE TO RPC CALL ==========


 ========== START RPC CALL ==========


 ========== RPC CALLED ==========

/test-module:activate-software-image/image-name = acmefw-2.3
/test-module:activate-software-image/location = /root/
ASAN:SIGSEGV
=================================================================
==2460==ERROR: AddressSanitizer: SEGV on unknown address 0x000000415c30 (pc 0x7f23e13d3abc bp 0x000000415c30 sp 0x7f23d7cdfe90 T6)

 ========== PRINT RETURN VALUE ==========

/test-module:activate-software-image/status = The image acmefw-2.3 is being installed.
/test-module:activate-software-image/version = 2.3
/test-module:activate-software-image/location = /root/
/test-module:activate-software-image/init-log (container)

 ========== SUBSCRIBE TO RPC TREE CALL ==========

    #0 0x7f23e13d3abb  (/lib64/libasan.so.2+0x21abb)
    #1 0x7f23e144a6dc in __interceptor_free (/lib64/libasan.so.2+0x986dc)
    #2 0x7f23e0bf3df2 in sr_free_val_content /home/centos/sysrepo/src/common/sr_utils.c:1609
    #3 0x7f23e0bdbdd6 in sr_free_values /home/centos/sysrepo/src/common/sr_common.c:91
    #4 0x7f23e0ca60a7 in cl_sm_rpc_process /home/centos/sysrepo/src/clientlib/cl_subscription_manager.c:912
    #5 0x7f23e0ca93ca in cl_sm_conn_msg_process /home/centos/sysrepo/src/clientlib/cl_subscription_manager.c:1019
    #6 0x7f23e0ca9f3c in cl_sm_conn_in_buff_process /home/centos/sysrepo/src/clientlib/cl_subscription_manager.c:1064
    #7 0x7f23e0cabd2b in cl_sm_fd_read_data /home/centos/sysrepo/src/clientlib/cl_subscription_manager.c:1150
[...]

PR #412, opened 2016-10-12
Refactoring of C++'s logging with one SR change

jktjkt:
This series:
- switches the C++ code to static methods (please read the commit messages),
- adds logging callbacks for C++,
- changes core SR to not skip expensive computation when a custom CB is active (see the last commit).

I have no idea whether you want the last commit or not; please feel free to remove that one.

Ccing @mislavn for the C++ part.

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -3110,7 +3110,7 @@ dm_commit_notify(dm_ctx_t *dm_ctx, dm_session_t *session, sr_notif_event_t ev, d
         }
 
         /* Log changes */
-        if (NULL != diff && (SR_LL_DBG == sr_ll_stderr || SR_LL_DBG == sr_ll_syslog)) {
+        if (NULL != diff && (SR_LL_DBG == sr_ll_stderr || SR_LL_DBG == sr_ll_syslog || sr_log_callback)) {
             while (LYD_DIFF_END != diff->type[d_cnt]) {
                 char *path = dm_get_notification_changed_xpath(diff, d_cnt);
                 SR_LOG_DBG("%s: %s", dm_get_diff_type_to_string(diff->type[d_cnt]), path);
--- a/swig/cpp/examples/get_item_example.cpp
+++ b/swig/cpp/examples/get_item_example.cpp
@@ -30,7 +30,10 @@ int
 main(int argc, char **argv)
 {
     try {
-        Logs::set_stderr(SR_LL_DBG);
+        Logs::set_stderr(SR_LL_NONE);
+        Logs::setCallback([](const sr_log_level_t logLevel, const std::string &message) {
+            std::cerr << "sysrepo log (" << logLevel << "): " << message << std::endl;
+        });
 
 	S_Connection conn(new Connection("app1"));
 
--- a/swig/cpp/examples/get_item_example.cpp
+++ b/swig/cpp/examples/get_item_example.cpp
@@ -30,8 +30,7 @@ int
 main(int argc, char **argv)
 {
     try {
-        Logs log;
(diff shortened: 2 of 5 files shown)


mislavn:
I will check why swig breaks it this pull request.

mislavn:
@jktjkt swig doesn't have out-of-the-box support for std::function so i would need to add custom code for all of the target languages.

An easier solution would be to add a virtual functions and custom wrapper code for Python and Lua, just like in the Callback class.

@jktjkt  will this be a problem for you?

If you want the C++ bindings could keep the std::function but Java, Python and Lua will use the virtual functions.

jktjkt:
> I will check why swig breaks it this pull request.

Sorry for confusion -- this patch series does not break stuff; one needs to nuke the build dir so that the build system regenerates the relevant SWIG intermediate files.

> swig doesn't have out-of-the-box support for std::function

Yes :(. I prefer to have C++ bindings using native C++ idioms; if you would like to, I can provide an extra wrapper using virtual methods on top of that for use by SWIG's target languages.

jktjkt:
I see that Travis builds fail; I'm using SWIG Version 3.0.5 locally.

mislavn:
> Yes :(. I prefer to have C++ bindings using native C++ idioms; if you would like to, I can provide an extra wrapper using virtual methods on top of that for use by SWIG's target languages.

That would be great.

> I see that Travis builds fail; I'm using SWIG Version 3.0.5 locally.
OK, I will check this.

mislavn:
@rastislavszabo can you change the swig version on the build server.

I will check now what is the minimum swig version required for successful build of the language bindings.

mislavn:
@jktjkt I can not build this locally without ignoring some methods in the Logs class.
Did you locally built swig bindings for python or lua, it breaks only on them.

rastislavszabo:
@mislavn 
> can you change the swig version on the build server

If it's really needed, you can change it here:
https://github.com/sysrepo/sysrepo/blob/master/deploy/travis/install-libs.sh
and here:
https://github.com/sysrepo/sysrepo/blob/master/deploy/travis/install-libs-osx.sh

jktjkt:
> I can not build this locally without ignoring some methods in the Logs class.

Clean build fails for me, too, so I made these methods only available when doing C++ stuff. Is this OK-ish for you? An alternative is to simply kill this feature for now; the other commits are IMHO still valuable. And yes, I'm OK with adding a virtual-method-based interface, but I'd prefer to wait until I finish legal paperwork locally (which means that I cannot give any ETA).

mislavn:
> Is this OK-ish for you? An alternative is to simply kill this feature for now; the other commits are IMHO still valuable. And yes, I'm OK with adding a virtual-method-based interface, but I'd prefer to wait until I finish legal paperwork locally (which means that I cannot give any ETA).

I can add a small patch to swig to ignore some methods from the Logs class, it will compile then.
I will send you the patch once I am finished with the issue:

https://github.com/sysrepo/sysrepo/issues/379

mislavn:
@jktjkt  can you add the line:

 %ignore Logs;

in the file swig/swig_base/cpp_classes.i anywhere after the 30-th line.

Then it should compile. When you are finished with the class and add the virtual functions I will make it compile with swig.

rastislavszabo:
@mislavn , @jktjkt What about this pull request? It hasn't been touched for 2 weeks and its build is still failing in Travis. How can we merge it / should we close it?

rastislavszabo:
Guys, it seems that this pull request cannot be merged and no one knows how to fix it. Please open a new one (or ones) if you need any changes in this PR.

PR #414, opened 2016-10-12
Double

mislavn:
Hi Rastislav,

in this pull request I fixed a problem with setting double values from lua and python.


swig/cpp/src/Struct.h
```diff
@@ -70,19 +70,19 @@ class Val
     Val(sr_val_t *val, S_Counter counter);
     Val(const char *val, sr_type_t type = SR_STRING_T);
     Val(bool bool_val, sr_type_t type = SR_BOOL_T);
-    Val(double decimal64_val, sr_type_t type = SR_DECIMAL64_T);
```
jktjkt: (I'm commenting because I'm playing with the exact same part, trying to debug SR_DECIMAL64_T in C++.)

Why is this particular ctor removed? My understanding based on the commit message is that this is some SWIG bug, is that right? One minor drawback is that it makes the C++ API a bit uglier.

mislavn: The problem was that python and lua do not recognize between int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t, uint64_t and double, it is the same type for them.

swig simply overloads the constructors and functions. So now there are two constructors/functions for numbers , one for double and one for integers (int8_t, ..). So their type signature need to be different that is why in swig there is now:

Val(double decimal64_val);
Val(uint64_t uint64_val, sr_type_t type)

As you can see swig puts a few limitations on your code.

If you have any better solution I am open to change.

mislavn: @jktjkt what do you want to change in the pull request?

I could change this in the swig file, replace the constructor with the one specified in the swig file or I can add a macro "#ifdef SWIG" in the C++ source code for SWIG specific code.

Those are our two options for now.

jktjkt: I clearly lack SWIG experience here. I was bothered a small bit by the irregularity in these constructors. If I was designing a C++-only API, I would go with abandoning the type info tag and rely on native overload resolution rules, something like this:

```c++
explicit Val(const int8_t v);
explicit Val(const int16_t v);
explicit Val(const int32_t v);
explicit Val(const uint8_t v);
explicit Val(const uint16_t v);
explicit Val(const uint32_t v);
explicit Val(const double v);
//...
```
However, I have no clue if that would play with SWIG at all, so let's ignore this nitpicking for now.

PR #423, opened 2016-10-14
Yang 1.1 actions; nested event notifications; lyd_merge; bug fixes; printer for sysrepo trees

milanlenco:
There is more than enough in this one pull request.

Originally it was supposed to be only about actions and nested event notifications, but then many bugs related to validation of rpc/action/notifications, cross module dependencies, sysrepoctl and operational data have emerged. Many bugs were found also in libyang (thank you lyd_merge) and I have waited for them to get fixed so that all new and old unit tests would pass.

I think this should go to master, since we wanted all yang 1.1 features to have supported in the latest release, plus the bug fixes certainly should be included in master.

The only extra feature I have added are pretty (I hope) printers for sysrepo trees (see sysrepo/trees.h). They will be needed in the upcoming new examples featuring sysrepo trees (will they go to master as well?) and are handy also for debugging purposes.


rastislavszabo:
will be merged into master

PR #436, opened 2016-10-19
fix crash when sr_module_change_subscribe failed in a special condition

fanchanghu:
Hi,
it will crash when `sr_module_change_subscribe` with `SR_SUBSCR_CTX_REUSE` failed in a special condition:
first, do a successful subscribe.
second, do a failed subscribe, then `sr_subscription->sm_subscription_cnt` becomes 0.
third, do another failed subscribe, then `sr_subscription` is freed. but the pointer of *subscription_p is not set to null. user's program will crash when he used the pointer later.
it looks like `cl_subscription_init` does not add new `cl_sm_subscription_ctx_t` but `cl_sr_subscription_remove_one` delete one.


fanchanghu:
it seems there are still some problems, I'll try to improve it.

Issue #438, opened 2016-10-19
subscriptions_cnt is reduced 2(expect 1) when sr_module_change_subscribe failed


fanchanghu:
`cl_subscription_close` will reduce` subscriptions_cnt` and is called twice, first in `cl_subscription_init`, second directly in` sr_module_change_subscribe`.

PR #440, opened 2016-10-20
BUGFIX sr_module_change_subscribe free error when cl_subscription_init failed

fanchanghu:
Hi,
this request is for fixing #438 , and also, `cl_sr_subscription_remove_one` need be called only when we added a `sm_subscription` to `sr_subscription` before.


milanlenco:
Thanks for the fix.

PR #441, opened 2016-10-20
Print backtrace of a crashed unit test in travis.

milanlenco:
With this pull request, a crashed unit test will generate core file in travis and the backtrace of every thread will get printed using gdb.
The only limitation is that travis will not generate more than one core file in one build, so for multiple crashed unit tests only the backtrace of the first one will get printed.
All travis build caches need to be cleared for this to take an effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ before_install:
 
 before_script:
   - ulimit -c unlimited -S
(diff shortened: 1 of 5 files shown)


milanlenco:
OK, now it is not limited to one core file only.

PR #445, opened 2016-10-20
Counter

mislavn:


mislavn:
I renamed the Counter class to Deleter and add private variable "S_Deleter _deleter;" to class Data so it resources could not be freed prematurely.

Issue #485, opened 2016-11-01
trees_test: assert failure with -DUSE_SR_MEM_MGMT=OFF


jktjkt:
```
[==========] Running 9 test(s).
[ RUN      ] sr_new_tree_test
[       OK ] sr_new_tree_test
[ RUN      ] sr_new_trees_test
[       OK ] sr_new_trees_test
[ RUN      ] sr_node_set_name_test
[       OK ] sr_node_set_name_test
[ RUN      ] sr_node_set_module_test
[       OK ] sr_node_set_module_test
[ RUN      ] sr_node_set_string_test
[       OK ] sr_node_set_string_test
[ RUN      ] sr_node_add_child_test
[       OK ] sr_node_add_child_test
[ RUN      ] sr_dup_tree_test
[       OK ] sr_dup_tree_test
[ RUN      ] sr_dup_trees_test
[       OK ] sr_dup_trees_test
[ RUN      ] sr_print_tree_test
trees_test: ../src/common/sr_utils.c:1734: void sr_free_tree_content(sr_node_t *): Assertion `0 == tree->data.int32_val' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff61f63db in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:55
55      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff61f63db in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:55
#1  0x00007ffff61f7a71 in __GI_abort () at abort.c:89
#2  0x00007ffff61eedd8 in __assert_fail_base (fmt=0x7ffff6338cc8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=assertion@entry=0x7982e0 <.str> "0 == tree->data.int32_val", file=file@entry=0x795620 <.str.21> "../src/common/sr_utils.c", line=line@entry=1734, 
    function=function@entry=0x798320 <__PRETTY_FUNCTION__.sr_free_tree_content> "void sr_free_tree_content(sr_node_t *)") at assert.c:92
#3  0x00007ffff61eee96 in __GI___assert_fail (assertion=0x7982e0 <.str> "0 == tree->data.int32_val", file=0x795620 <.str.21> "../src/common/sr_utils.c", line=1734, 
[...]

PR #486, opened 2016-11-01
Travis-CI: Also test builds without custom memory management

jktjkt:
This is currently expected to fail for (at least) #485.


jktjkt:
(I'll be force-pushing into this branch until I manage to get it right or give up :).)

PR #495, opened 2016-11-02
Install the C++ headers

jktjkt:
I'm not installing directly to include/sysrepo because the include setup
is a bit weird. There's an include/sysrepo/*.h and include/sysrepo.h,
and the include paths not necessarily reflect all that.

Long term, I would like the C++ headers to use better include paths,
such as sysrepo/Foo.h, but that would require some shuffling around and
SWIG updates, and it's too late in the evening for them now.

### DIFF
--- a/swig/CMakeLists.txt
+++ b/swig/CMakeLists.txt
@@ -178,6 +178,8 @@ endif()
 
 # install binary
 install(TARGETS Sysrepo-cpp DESTINATION ${LIB_INSTALL_DIR})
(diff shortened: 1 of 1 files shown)


rastislavszabo:
Hi,

> I'm not installing directly to include/sysrepo because the include setup is a bit weird. 

Feel free to re-organize it. For the sake of backward compatibility we would need to keep following pattern of includes:

```
#include "sysrepo.h"
```
for the main API and

```
#include "sysrepo/values.h"
```

for the optional extensions.

PR #496, opened 2016-11-02
Install systemd units

jktjkt:
Because these go into a fixed location, these have to be enabled by
configuring with ``-DWITH_SYSTEMD=ON``.


rastislavszabo:
Could you please add some information on systemd to docs?
e.g. https://github.com/sysrepo/sysrepo/blob/master/INSTALL.md#useful-cmake-options
or https://github.com/sysrepo/sysrepo/blob/master/INSTALL.md#using-sysrepo

PR #546, opened 2016-11-15
Cpp update

mislavn:
Hi Rastislav,

in this pull request I updated some C++ classes and implemented the new action API.

The only problem is that I am not sure  how to implement the functions:
sr_val_build_str_data(sr_val_t *value, sr_type_t type, const char *format, ...);
sr_node_build_str_data(sr_node_t *node, sr_type_t type, const char *format, ...);

I was not clear for me how to implement it from the examples in the unit tests.
Can you explain me how to use this API.

Regards,
Mislav

### DIFF
--- a/swig/cpp/examples/cpp_application_changes_example.cpp
+++ b/swig/cpp/examples/cpp_application_changes_example.cpp
@@ -34,74 +34,6 @@ using namespace std;
 
 volatile int exit_application = 0;
 
-/* Function for printing out values depending on their type. */
-void
-print_value(S_Val value)
-{
-    cout << value->xpath();
-    cout << " ";
-    switch (value->type()) {
-    case SR_CONTAINER_T:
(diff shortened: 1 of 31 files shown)


rastislavszabo:
Hi Mislav,
instead of doing:
```c
char string[STR_SIZE];
snprintf(string, STR_SIZE, "value-built-from-string-and-integers-%d-%d", int1, int2);
sr_val_set_str_data(value, SR_STRING_T, string);
```
you can do:
```c
sr_val_build_str_data(value, SR_STRING_T, "value-built-from-string-and-integers-%d-%d", int1, int2);
```
`sr_node_build_str_data` works similarly. 

Hope it's clear now.

mislavn:
I updated the pull request based on the comments from @milanlenco in the sysrepo-private email thread.

@milanlenco can you take a look at the language bindings examples.

PR #549, opened 2016-11-18
Search for module schema by name *recursively* (fix Netopeer issue #54)

milanlenco:
Fix for: https://github.com/CESNET/Netopeer2/issues/54


michalvasko:
Hi,
I am still seeing the error, am I doing something wrong or there is still some problem, does netopeer2 devel-server branch server install work for you (with having authd installed)?

Regards,
Michal

PR #557, opened 2016-11-22
Lua53

mislavn:
I add Lua 5.3 support to swig and refactored the Lua bindings.


mislavn:
The build failed on osx, it had the same problem as we had with swig 3.0.2, so I add a patch that sets swig to version 3.0.5 in brew.

PR #572, opened 2016-11-26
fix a bug that data in buffer is cleared when only part of a big message is received(EWOULDBLOCK occured during receiving the message).

fanchanghu:
As the current code, `buff->pos` is set to 0 when `buff_pos == 0` and `(buff_size - buff_pos) > 0`, this will clear the received data in buffer when only part of a big message is received(EWOULDBLOCK occured during receiving the message).


fanchanghu:
Hi,
I have changed it in `cl_sm_conn_in_buff_process` function too. But it is difficult to be reproduced, it only occurred accidentally in one of our test environment.

PR #574, opened 2016-11-28
Lua

mislavn:
Add a bugfix for lua and updated lua examples to use "io.write" instead "print".


mislavn:
> I believe that this commit is redundant. You are using a regular operator new which always throws std::bad_alloc when it is unable to allocate memory.

OK. I will remove this commit and at other places in code where this occurs.

jktjkt:
Oh, apparently GH comments don't apply to individual commits, but to the entire PL. I like https://github.com/sysrepo/sysrepo/pull/574/commits/58dd0c28a38d5a8d6d70dd811010484bf5232ef2 .

mislavn:
I also removed the memory check in the lua and python callbacks.

PR #583, opened 2016-11-29
Add support to sysrepo for rfc6020 uses statement

WinterZheng:
Provide support for the uses statement which is used to reference a “grouping” definition.  See https://tools.ietf.org/html/rfc6020#section-7.12 for more details.


milanlenco:
Thanks for the fix!

PR #590, opened 2016-11-30
Bindings update

mislavn:
In this pull request I add the Deleter class to Session so sr_session_stop is not called in the sr_module_change_cb function callback (this applies to all of the language bindings).

Also I add return code support for all language binding callbacks.
So now you can have "return sr.SR_ERR_NOMEM" in all of the language bindings for functions which have int in their type signature.


mislavn:
I add a new commit https://github.com/sysrepo/sysrepo/pull/590/commits/7fff89e57421c24c409033f8895ec2052b9a5305 which simplifies the lua bindings, it uses the same name for lua5.1, lua5.2 and lua5.3 bindings.

The bug Joan reported on the sysrepo mailing list should be fixed, I could not replicate his bug but it should not be possible any more with the commit: https://github.com/sysrepo/sysrepo/pull/590/commits/eec4931ae6531ac8dae932e9325c857c01492dbc

PR #598, opened 2016-12-03
Module enable can't enable sub nodes with uses statement

morganzhh:
For RFC6020 section 7.12.1 support, pull request#583 solve the issue of sysrepoctl.
Here is another issue, if the 'uses' statement is directly under a 'module', function dm_enable_module_running will ignore it and the result of this is, nodes under that 'uses' statement will not be enabled, hence can not be edited.


lukasmacko:
Thanks!

PR #606, opened 2016-12-06
Documentation updates

rastislavszabo:


jktjkt:
Thanks for the docs, they read quite well :). I was thinking about using something "less obvious" for the shell escaping and the ``system()`` call, but in the end it probably doesn't matter in this context.

PR #607, opened 2016-12-08
Let C++ callbacks also have reture values

morganzhh:
We are using Sysrepo C++ interface, in which, those c++ callbacks cannot take any return values, it would be better that callbacks could also return OK/Failed to all the change events, especially on 'Verify' event.


rastislavszabo:
@mislavn will this work with all language bindings?

rastislavszabo:
already implemented in #590 (`devel` branch)

PR #629, opened 2016-12-14
Support for subscriptions with user credentials.

milanlenco:


milanlenco:
Build fails due to the issue https://github.com/CESNET/libyang/issues/222

PR #633, opened 2016-12-16
add default nodes into empty data trees if empty data can be validated.

### LINKED ISSUE #335: Top level default nodes & empty data tree
Currently only way how to add default nodes is validation function `lyd_validate`. That is one of the reasons why validation used to be done for all data trees loaded from file system, even empty ones.

However,  validation fails if module contains a top-level mandatory node and data tree is empty. see https://github.com/sysrepo/sysrepo/issues/333 As a result of bugfix validation is skipped and default nodes are not added if data tree is empty during file loading.

Default nodes are correctly added during validaition `sr_validate()` and commit `sr_commit()`.

The aim of this issue is to add default nodes into empty data trees.

fanchanghu:
Hi, 
This request is used to fix #335 only when empty data can be validated (now, even if the empty data can be validated, get-config with 'report-all' still returns empty).

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -707,11 +707,25 @@ dm_load_data_tree_file(dm_ctx_t *dm_ctx, int fd, const char *data_filename, dm_s
     }
 
     /* if there is no data dependency validate it with of LYD_OPT_STRICT, validate it (only non-empty data trees are validated)*/
-    if (!schema_info->cross_module_data_dependency && NULL != data_tree && 0 != lyd_validate(&data_tree, LYD_OPT_STRICT | LYD_OPT_CONFIG, schema_info->ly_ctx)) {
-        SR_LOG_ERR("Loaded data tree '%s' is not valid", data_filename);
-        lyd_free_withsiblings(data_tree);
-        free(data);
-        return SR_ERR_INTERNAL;
+    if (!schema_info->cross_module_data_dependency) {
+        if (NULL != data_tree) {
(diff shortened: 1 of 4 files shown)


rastislavszabo:
Build fails due to https://github.com/CESNET/libyang/issues/222, waiting for fix in libyang.

rastislavszabo:
The build now works. It seems that the tests are not passing with this fix, we will need to look into that.

fanchanghu:
I  checked this problem, but I can not build success even if I revert my change. The error message like this:
```
-- Exec: /root/github/sysrepo/build/src/sysrepocfg --import=/root/github/sysrepo/examples/yang/ietf-interfaces.data.xml --datastore=startup --format=xml --level=0 ietf-interfaces
CMake Error at examples/cmake_install.cmake:79 (message):
  Error: 1
Call Stack (most recent call first):
  cmake_install.cmake:146 (INCLUDE)


make: *** [install] Error 1
```
what I do is like this:
1, update libyang and sysrepo to latest in branch master;
2, use `cmake .. && make install && make test` in libyang/build, it's OK;
3, use `cmake -DGEN_LANGUAGE_BINDINGS=0 .. && make install && make test` in sysrepo/build, and get the error. (If I do not use `-DGEN_LANGUAGE_BINDINGS=0`, some .cpp files failed to compile.)

fanchanghu:
I changed 2 test cases according to the CI job log, hope it can work now...

fanchanghu:
Hi,
I try to change some test cases to make them pass, however, I'm not sure it's OK to do it like that...
@rastislavszabo

Issue #644, opened 2016-12-21
C++ Session.cpp destructor


morganzhh:
In ``devel`` branch, we have introduced a ``deleter`` in Session's constructor, so should we remove the ``sr_session_stop`` from Session's destructor?

```c
Session::~Session()
{
    if (_sess) {
        int ret = sr_session_stop(_sess);
        if (ret != SR_ERR_OK) {
            throw_exception(ret);
        }
	_sess = NULL;
    }
}

```

When getting callback from module change

```c
static int module_change_cb(sr_session_ctx_t *session, const char *module_name, sr_notif_event_t event, void *private_ctx) {
    S_Session sess(new Session(session));
    Callback *wrap = (Callback*) private_ctx;
    return wrap->module_change(sess, module_name, event, wrap->private_ctx["module_change"]);
}
```

We use ``S_Session``, once  ``module_change_cb`` returned, the ``session`` get stopped, which causes any error message set by application(``wrap->module_change``) get flushed.

```c
int
sr_session_stop(sr_session_ctx_t *session)
{
    Sr__Msg *msg_req = NULL, *msg_resp = NULL;
    sr_mem_ctx_t *sr_mem = NULL;
    int rc = SR_ERR_OK;

    CHECK_NULL_ARG2(session, session->conn_ctx);

    cl_session_clear_errors(session);  ======>>>>

```

rastislavszabo:
@mislavn , @jktjkt , could you please have a look?

mislavn:
Yes that should not be there. I will make a pull request for this.

jktjkt:
My knowledge of SR's internals is limited, so while I can confirm that the code indeed calls ``sr_session_stop``, I have no idea whether it's correct or incorrect. I would have thought that it's a good thing that the session is cleaned up, otherwise there's an obvious resource leak. I have no idea whether it causes any error messages to get lost.

mislavn:
@jktjkt sr_session_stop should be called only from the Deleter class and the Deleter class is created only when a new Session class was created from the user, else when a sysrepo API returns a session, sysrepo takes care of it so it is not called from the Deleter class.

That piece of code I will remove in the pull request https://github.com/sysrepo/sysrepo/pull/646

PR #646, opened 2016-12-21
cpp: remove sr_session_stop from the Session destructor

mislavn:
sr_session_stop should be called only from the Deleter class and the
Deleter class is created only when a new Session class was created
from the user.


mislavn:
Bugfix for the issue: https://github.com/sysrepo/sysrepo/issues/644

mislavn:
> I would amend the commit to include a reference to the bug # before merging this.

Done.

PR #649, opened 2016-12-21
Event notifications store

rastislavszabo:


src/common/sr_data_structs.c
```diff
@@ -561,7 +561,7 @@ sr_free_lock_item(void *lock_item)
 {
     CHECK_NULL_ARG_VOID(lock_item);
     sr_lock_item_t *li = (sr_lock_item_t *) lock_item;
-    free(li->filename);
+    free((void*)li->filename);
```
jktjkt: Are you sure about this one? That might make it possible to assign a literal constant to ``filename`` and ``free()`` it later on.

src/common/sr_utils.c
```diff
@@ -2547,6 +2552,42 @@ sr_create_uri_for_module(const struct lys_module *module, char **uri)
...
+    char *time_str_copy = NULL, *colon = NULL;
+
+    CHECK_NULL_ARG2(time_str, time);
+
+    time_str_copy = strdup(time_str);
+    CHECK_NULL_NOMEM_RETURN(time_str_copy);
+
+    /* time str ends in '+hh:mm' but should be '+hhmm' */
+    colon = strrchr(time_str_copy, ':');
+    memmove(colon, colon + 1, 2);
```
jktjkt: Does the user have control over the string data? If yes, then this is an undefined behavior when the string doesn't contain colon.

rastislavszabo: No, this function is currently used just internally. But I'll fix it for the case we will expose it publicly later.

src/common/sr_data_structs.c
rastislavszabo: Yes, I want to have the `filename` member of the struct `sr_lock_item_s` `const char *`, not a `char *`.

BTW, assignment of a `const char *` to `char *` is allowed, so freeing a literal constant is possible either way, e.g.:
```c
char *str = "abc";
free(str);
```
compiles just fine.

PR #650, opened 2016-12-21
String val

mislavn:
Implemented C++ methods for functions sr_set_item_str() and sr_val_to_str().

@rastislavszabo I am not sure what error message to return if I get NULL from sr_val_to_str().
Now it returns SR_ERR_OPERATION_FAILED.

Also would it be a problem to create a similar functions for nodes, something like sr_node_to_str().

### DIFF
--- a/swig/cpp/src/Session.cpp
+++ b/swig/cpp/src/Session.cpp
@@ -303,6 +303,14 @@ void Session::set_item(const char *xpath, S_Val value, const sr_edit_options_t o
     }
 }
 
+void Session::set_item_str(const char *xpath, const char *value, const sr_edit_options_t opts)
+{
+    int ret = sr_set_item_str(_sess, xpath, value, opts);
+    if (ret != SR_ERR_OK) {
+        throw_exception(ret);
+    }
+}
+
 void Session::delete_item(const char *xpath, const sr_edit_options_t opts)
 {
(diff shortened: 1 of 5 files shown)


rastislavszabo:
>  I am not sure what error message to return if I get NULL from sr_val_to_str(). Now it returns SR_ERR_OPERATION_FAILED.

I'm not sure either. I actually started to think that `sr_val_to_str` should return an error code as the rest of the API functions... For now we can leave it as `SR_ERR_OPERATION_FAILED`.

rastislavszabo:
> Also would it be a problem to create a similar functions for nodes

`sr_val_to_str` can be used also used for nodes, since nodes can be safely casted to `sr_val_t`. But we can create wrappers for `sr_node_t` type too.

mislavn:
> sr_val_to_str can be used also used for nodes, since nodes can be safely casted to sr_val_t. But we can create wrappers for sr_node_t type too.

Thanks @rastislavszabo 

Also if it is not a problem, could you make a function like:
int string_to_val(sr_val_t **value, const char *xpath, const char *value);

The function would create a sr_val_t from the xpath and value.
This would be helpful with operational data.

rastislavszabo:
@mislavn: it is `sr_val_set_xpath` + `sr_val_set_str_data`

mislavn:
OK, thanks.

PR #651, opened 2016-12-21
cpp: remove exception throwing from class destructors

mislavn:
This is a temporary fix for the Coverity bug report of type:
'Error handling issues  (UNCAUGHT_EXCEPT)'

### DIFF
--- a/swig/cpp/src/Internal.cpp
+++ b/swig/cpp/src/Internal.cpp
@@ -101,7 +101,8 @@ Deleter::~Deleter() {
         if (!v._sess) break;
         int ret = sr_session_stop(v._sess);
         if (ret != SR_ERR_OK) {
-            throw_exception(ret);
+            //this exception can't be catched
+            //throw_exception(ret);
         }
 	v._sess = NULL;
     break;
--- a/swig/cpp/src/Session.cpp
+++ b/swig/cpp/src/Session.cpp
@@ -444,7 +444,8 @@ Subscribe::~Subscribe()
     if (_sub && _sess->_sess) {
         int ret = sr_unsubscribe(_sess->_sess, _sub);
         if (ret != SR_ERR_OK) {
-            throw_exception(ret);
+            //this exception can't be catched
+            //throw_exception(ret);
         }
 	_sub = NULL;
     }


mislavn:
@jktjkt , @rastislavszabo do you have any suggestion what to do instead of throwing a exception in the destructor. I can't think of anything then to ignore it.

Here is the report from Coverity Scan.

________________________________________________________________________________________________________
*** CID 158947:  Error handling issues  (UNCAUGHT_EXCEPT)
/home/travis/build/sysrepo/sysrepo/swig/cpp/src/Internal.cpp: 70 in Deleter::~Deleter()()
64         _t = SCHEMAS;
65     }
66     Deleter::Deleter(sr_session_ctx_t *sess) {
67         v._sess = sess;
68         _t = SESSION;
69     }
>>>     CID 158947:  Error handling issues  (UNCAUGHT_EXCEPT)
>>>     An exception of type "sysrepo_exception" is thrown but the throw list "throw()" doesn't allow it to be thrown. This will cause a call to unexpected() which usually calls terminate().
70     Deleter::~Deleter() {
71         switch(_t) {
72         case VAL:
73             if (v._val) sr_free_val(v._val);
74      v._val = NULL;
75         break;

** CID 158946:  Error handling issues  (UNCAUGHT_EXCEPT)
/home/travis/build/sysrepo/sysrepo/swig/cpp/src/Session.cpp: 442 in Subscribe::~Subscribe()()


________________________________________________________________________________________________________
*** CID 158946:  Error handling issues  (UNCAUGHT_EXCEPT)
/home/travis/build/sysrepo/sysrepo/swig/cpp/src/Session.cpp: 442 in Subscribe::~Subscribe()()
436         _sub = NULL;
437         _sess = sess;
438         swig_sub = _sub;
439         swig_sess = _sess;
440     }
441
>>>     CID 158946:  Error handling issues  (UNCAUGHT_EXCEPT)
>>>     An exception of type "sysrepo_exception" is thrown but the throw list "throw()" doesn't allow it to be thrown. This will cause a call to unexpected() which usually calls terminate().
442     Subscribe::~Subscribe()
443     {
444         if (_sub && _sess->_sess) {
445             int ret = sr_unsubscribe(_sess->_sess, _sub);
446             if (ret != SR_ERR_OK) {
447                 throw_exception(ret);

rastislavszabo:
> do you have any suggestion what to do instead of throwing a exception in the destructor

Nope. Perhaps just log it.

mislavn:
> Nope. Perhaps just log it?

OK. I will update the commit.

mislavn:
@rastislavszabo it is not easy to expose the sysrepo logging API to C++. For now let this be simply commented out.

rastislavszabo:
ok

PR #660, opened 2016-12-22
lua: bugfix in type signature of lua callbacks

mislavn:


mislavn:
Changed the return type in Lua callback dp_get_items from Vals to Vals_Holder also changed the type signature for Trees_Holder/Vals_Holder to SWIGTYPE_p_Trees/Vals_Holder.

PR #704, opened 2017-01-26
Port of existing Python2 tests to Python3.

antonio-paunovic:
Differences between Python versions and CMake are adjusted to work same as Python2 tests.

Tests are copied and modified to work with Python3.


swig/python3/tests/NotificationTest.py
```diff
@@ -0,0 +1,570 @@
...
+
+class NotificationTester(SysrepoTester):
+
+    def cleanup(self):
+        if self.filename and os.path.isfile(self.filename):
+            os.unlink(self.filename)
+
+    def subscribeStep(self, xpath):
+        self.filename = "notifications_test_" + str(randint(0, 9999))
+        self.process = subprocess.Popen(["notifications_test_app", xpath, self.filename])
```
lukasmacko: I think we agreed yesterday that this external app will be replaced by python (binding-native) code, correct me if I am wrong.

antonio-paunovic: Hi Lukas, 

We were talking about Lua at the time,  should I do it for Python tests also?

lukasmacko: Hi Antonio,
yes you're right. Anyway, I think it would be great if you could do it for python test, also.

antonio-paunovic: Sure, I'll get to it tomorrow. Should I reuse this pull request or make other when I'm done ?

lukasmacko: You can reuse this one, just push more commits into this branch.

PR #743, opened 2017-02-22
Change getting NACM flag from schema node

PavolVican:


mislavn:
Hi,
at the moment the libynag and sysrepo devel branches are not in sync, we will fix this in the next few days.

PavolVican:
Hi.

I think libyang is ready at current devel version. This pull request was pushed, but extension branch wasn't merged in devel.

Regards,
Pavol

rkrejci:
Tests fails for the reasons not connected with this pull request and we need this patch for other pull requests and for fixing the tests, so I'm going to aprove pull request.

PR #749, opened 2017-03-07
Remove troublesome Python binding tests.

antonio-paunovic:
Tests concerning notificatiosn and subscriptions will be removed now and
refactored later because those break most often. Plan is to make them more robust or more simple.


rkrejci:
well, some other test started to fail. But because it is running before the two removed tests, it seems that it is not connected with this change, so I'm going to merge it.

PR #762, opened 2017-04-04
C++ unit tests

antonio-paunovic:


swig/CMakeLists.txt
```diff
@@ -109,6 +109,19 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/cpp/src/" DESTINATION ${CMAKE_INS
...
+  set(CPP_TEST_MODULE "swig-test")
+  set(CPP_TEST_MODULE_REV "2017-03-09")
+  set(CPP_SWIG_DIR ${CMAKE_CURRENT_SOURCE_DIR}/cpp)
+  set(CPP_TEST_DIR ${CPP_SWIG_DIR}/tests)
+  set(CPP_YANG_DIR ${CPP_TEST_DIR}/yang)
+
+  ADD_CUSTOM_COMMAND(
+    TARGET Sysrepo-cpp
+    POST_BUILD
+    COMMAND ${CMAKE_BINARY_DIR}/src/sysrepoctl --install --yang=${CPP_YANG_DIR}/${CPP_TEST_MODULE}@${CPP_TEST_MODULE_REV}.yang
```
jktjkt: This looks like being in conflict with the change I introduced in 00025c4. Please ensure that it only runs when tests are enabled, otherwise the build will start failing again during cross-compilation.

swig/cpp/tests/changes.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+const char *
+get_xpath(char *test_name, char *node_name)
+{
+    char buf[MAX_LEN];
+    sprintf(buf, xpath_if_fmt, test_name, node_name);
+    return strdup(buf);
+}
+
+class NOPCb: public Callback {
```
jktjkt: nit: this is hard to read, consider something like NopCallback or NopCb if u must make shrtcts :)

swig/cpp/tests/changes.cpp
```diff
@@ -0,0 +1,219 @@
...
+char *
+get_test_name(int i)
+{
+    char buf[MAX_LEN] = {0};
+    sprintf(buf, "test-cpp-%d", i);
+    return strdup(buf);
+}
+
+const char *
+get_xpath(char *test_name, char *node_name)
```
jktjkt: I'm a C++ programmer, so please take this into perspective when reading this comment. This is a unit test for a C++ part of a program. Even if C code was more efficient, it doesn't matter because this is a test. In C++, this would become:

```c++
std::string get_xpath(const std::string &test_name, const std::string &node_name)
{
  return "/swig-test:cpp-operations/test-get[name='" + test_name + "']/" + node_name;
}
```

...and see a couple of comments below for usage.

swig/cpp/tests/changes.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+void init_test(S_Session sess)
+{
+    S_Subscribe subs(new Subscribe(sess));
+    S_Callback cb(new NOPCb());
+
+    subs->module_change_subscribe(module_name, cb, NULL, 0, SR_SUBSCR_DEFAULT | SR_SUBSCR_APPLY_ONLY);
+
+    for (int i = LOW_BOUND; i < HIGH_BOUND; i++) {
+        const char *xpath = (char*)get_xpath(get_test_name(i), (char *) "number");
```
jktjkt: Again, this is a matter of C++-ish preferences. These casts are not needed (or can be made not needed) especially if one follows the C++ idioms instead of the C ones.

```C++
const auto xpath = prepare_test_xpath(get_test_name(1), "number");
```

swig/cpp/tests/changes.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+            try {
+                snprintf(change_path, MAX_LEN, "/%s:*", module_name);
+                auto it = sess->get_changes_iter(&change_path[0]);
+
+                while (auto change = sess->get_change_next(it)) {
+                    assert(SR_OP_DELETED == change->oper());
+                }
+            } catch (const std::exception& e) {
+                cout << e.what() << endl;
```
jktjkt: This silently eats exceptions which is probably not something that you want to do in a unit test.

If you want to ignore an exception which is always thrown, add an assert that checks that it was thrown. If it doesn't really throw, then do not catch these exceptions and let the code crash -- it's a unit test. Crashing on unhandled exceptions is good.

swig/cpp/tests/changes.cpp
```diff
@@ -0,0 +1,219 @@
...
+};
+
+void
+test_module_change_create(S_Session sess)
+{
+    S_Subscribe subs(new Subscribe(sess));
+    S_Callback cb(new CreateCb());
+
+    init_test(sess);
+    usleep(100);
```
jktjkt: What's the purpose of these usleeps (also two more below)? How was the value determined? I'm asking because I'm getting several failures from the test suite in our internal code which uses sysrepo and multicore CI VMs. If there's any threading involved here and the test cannot be redesigned, then the timeouts must really be longer. If there's no threading, why bother sleeping?

antonio-paunovic:
Hi @jktjkt ,
thanks for quick feedback. These are first version of basic tests for language bindings in vain to get started. Idea is to have some straightforward sysrepo tests to make sure basic API works. 
This is to be done for other bindings as well (hopefully today). 

Although tests for bindings are supposed to stay small, you're welcomed to suggest more tests that you think would be useful. Your comments seem well-placed and I'll address them tomorrow.

swig/cpp/tests/changes.cpp
antonio-paunovic: I've done it like that because wrapper functions return C types. If std::string is preferable in  C++ then we'll change it like that. Do you think bindings should be updated to return std::string instead of char* in general?

swig/cpp/tests/changes.cpp
antonio-paunovic: Looks like good advice.

swig/cpp/tests/changes.cpp
antonio-paunovic: OK

swig/cpp/tests/changes.cpp
antonio-paunovic: I'll take a deeper look into this.

antonio-paunovic: I've refactored tests to use decoupled modules. 
Before, race could occur because same module could be modified by tests.
Sleeps were used for it to work sequentially, now tests can be run with ctest -j2 or more.

antonio-paunovic:
Hi @jktjkt 
I've tried to address points you've made. 
Is there anything else or is this mergeable?

Issue #671, opened 2017-01-03
sysrepoctl core dump during installing a yang file with two 'augment' syntax


WinterZheng:
The sysrepoctl crashed when installing a yang file with two 'augment' syntax. Bellow is the error logs.

> tester@Wzheng build % sysrepoctl --install --yang=/home/tester/ports.yang 
> Installing a new module from file '/home/tester/ports.yang'...
> Installing the YANG file to '/home/tester/3rd/sysrepo/build/repository/yang/ports.yang'...
> Installing data files for module 'ports'...
> [1]    10672 segmentation fault (core dumped)  sysrepoctl --install --yang=/home/tester/ports.yang

Here is the yang file ports.yang :
```YANG
module ports {
    namespace "http://example.com/ns/ports";
    prefix ports;

    container ports {
        list port {
            key name;
            leaf name {
                type string;
            }
            leaf rate {
                type uint32;
            }
        }
    }

    augment "/ports/port" {
        when "name = 'input'";
        leaf input-power {
            type int32;
        }
    }

    augment "/ports/port" {
        when "name = 'output'";
        leaf output-power {
            type int32;
        }
    }

}
```
The core dump happened in the function 'md_traverse_schema_tree' in the 'module_dependencies.c'.  

next_node:
            /* backtracking + automatically moving to the next sibling if there is any */
            if (node != root) {

The 'input-power' node's next sibling node is 'output-power', and they have the same parent node name '/ports/port'. But the address of 'node' pointer and 'root' pointer is different in this case.

lukasmacko:
fixed (unreproducible) with the latest devel branch

morganzhh:
Hi

I am afraid is's not solved, the reason for this bug is invalid data access, the test survived in my Ubuntu  VM but not in my Alpine Linux VM.

When the node is Augment, the ``lys_node.next`` is really ``lys_node_augment.when``, which don't have 'module', so when  ``LYS_MAIN_MODULE(node->next)`` get called, the behavior will be unknown.

```c
#define LYS_MAIN_MODULE(lys_elem)  \
            (lys_elem->module->type ? ((struct lys_submodule *)lys_elem->module)->belongsto : lys_elem->module)
```

```c
struct lys_node_augment {
  
    /* replaces #next and #prev members of ::lys_node */
    struct lys_when *when;           /**< when statement (optional) */
    struct lys_node *target;         /**< pointer to the target node */

};
```

```c
struct lys_when {
[...]

PR #771, opened 2017-04-05
fix traversing schema tree when exploring modules dependencies

rkrejci:
Ignore the augment nodes and go directly through them because the same
way we got into (via the child pointer), we have to get back (usually
parent, but in case of augmenting nodes we have to go through the agment
node)

Fixes sysrepo/sysrepo#671


rkrejci:
merged manually - 98315fd

PR #776, opened 2017-04-07
Fix macosx rpath warning.

antonio-paunovic:
Building on osx gives warning 

> CMake Warning (dev):
> 
>   Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
>   --help-policy CMP0042" for policy details.  Use the cmake_policy command to
>   set the policy and suppress this warning.
> 
>   MACOSX_RPATH is not specified for the following targets:
> 
>    dummy-plugin-1
>    dummy-plugin-2
>    sysrepo
>    turing-machine
> 
> This warning is for project developers.  Use -Wno-dev to suppress it.

and this would supposedly fix it.


antonio-paunovic:
Hi Radek,
in tests/CMakeLists.txt there is a comment that some tests including plugin daemon test. 
I was checking if putting that test as first of those 'stateful' tests would change anything. It passed that time but, as we know that doesn't mean it would fix the test. I'll turn that part back and try with #777.

antonio-paunovic:
You're right about version. I've put check to set variable only if version is higher than 2.8.11. 
We thought that this kind of change isn't good enough reason to rise minimal cmake version at this point.
Is this way OK?

antonio-paunovic:
OK, cmake version is bumped to 2.8.12.

Issue #800, opened 2017-04-20
warning reported by sanitizer


mislavn:
I used the latest sysrepo devel branch, compiled it with clang and `-fsanitize=thread`. 

When I run the sysrepo daemon I got the following warninng at startup:
```
root@1c33e6e10481:/opt/fork/sysrepo/build# ./src/sysrepod -d -l 3
[INF] (dm_init:1952) Initializing Data Manager, schema_search_dir=/opt/fork/sysrepo/build/repository/yang/, data_search_dir=/opt/fork/sysrepo/build/repository/data/
[INF] (nacm_init:827) Initializing NACM.
[INF] (nacm_init:878) NACM configuration was loaded from the startup datastore.
==================
WARNING: ThreadSanitizer: unlock of an unlocked mutex (or by a wrong thread) (pid=25470)
    #0 pthread_mutex_unlock <null> (sysrepod+0x00000042d26b)
    #1 ac_unset_user_identity /opt/fork/sysrepo/src/access_control.c:510 (libsysrepo.so+0x0000000a6f5f)
    #2 dm_lock_module /opt/fork/sysrepo/src/data_manager.c:1365 (libsysrepo.so+0x000000157edb)
    #3 dm_copy_config /opt/fork/sysrepo/src/data_manager.c:5068 (libsysrepo.so+0x00000019ad8a)
    #4 dm_copy_module /opt/fork/sysrepo/src/data_manager.c:5506 (libsysrepo.so+0x00000019631a)
    #5 dm_enable_module_running /opt/fork/sysrepo/src/data_manager.c:5239 (libsysrepo.so+0x000000195506)
    #6 nacm_init /opt/fork/sysrepo/src/nacm.c:885 (libsysrepo.so+0x000000209e9b)
    #7 dm_init /opt/fork/sysrepo/src/data_manager.c:2016 (libsysrepo.so+0x000000163526)
    #8 rp_init /opt/fork/sysrepo/src/request_processor.c:3861 (libsysrepo.so+0x0000000cc4d7)
    #9 cm_init /opt/fork/sysrepo/src/connection_manager.c:1884 (libsysrepo.so+0x0000000a7fd7)
    #10 main /opt/fork/sysrepo/src/executables/sysrepod.c:118 (sysrepod+0x0000004a2b58)

  Location is heap block of size 64 at 0x7d100000dfc0 allocated by main thread:
    #0 calloc <null> (sysrepod+0x00000042bb6d)
[...]

PR #801, opened 2017-04-20
BUGFIX for mutex handling

mislavn:
This patch fixes the case where ac_set_user_identity would not lock the
mutex and in ac_unset_user_identity it would get anyway unlocked which
resulted in a warning by sanitizer:

WARNING: ThreadSanitizer: unlock of an unlocked mutex (or by a wrong thread)

This is bug fix for the github issue https://github.com/sysrepo/sysrepo/issues/800.


mislavn:
@michalvasko please can you take a look at this pull request.

Issue #804, opened 2017-04-21
sanitizer warning for data race and deadlocks


mislavn:
I used the latest sysrepo devel branch, compiled it with clang and `-fsanitize=thread`.

At install time I got the following warninings.

```
-- Exec: /opt/fork/sysrepo/build/src/sysrepoctl --install --yang=/opt/fork/sysrepo/examples/yang/turing-machine.yang --permissions=666; /opt/fork/sysrepo/build/src/sysrepocfg --import=/opt/fork/sysrepo/examples/yang/turing-machine.data.xml --datastore=startup --format=xml --level=0 turing-machine
==================
WARNING: ThreadSanitizer: data race (pid=13948)
  Write of size 8 at 0x7d0c0000df10 by thread T5 (mutexes: write M27):
    #0 sr_cbuff_enqueue /opt/fork/sysrepo/src/common/sr_data_structs.c:514 (libsysrepo.so+0x00000005b730)
    #1 rp_msg_process /opt/fork/sysrepo/src/request_processor.c:4050 (libsysrepo.so+0x0000000d1103)
    #2 cm_req_process /opt/fork/sysrepo/src/connection_manager.c:858 (libsysrepo.so+0x0000000b3cff)
    #3 cm_conn_msg_process /opt/fork/sysrepo/src/connection_manager.c:1005 (libsysrepo.so+0x0000000b2c5e)
    #4 cm_conn_in_buff_process /opt/fork/sysrepo/src/connection_manager.c:1060 (libsysrepo.so+0x0000000b0cf8)
    #5 cm_conn_read_cb /opt/fork/sysrepo/src/connection_manager.c:1142 (libsysrepo.so+0x0000000aee62)
    #6 ev_invoke_pending <null> (libev.so.4+0x000000003d72)
    #7 cm_event_loop_threaded /opt/fork/sysrepo/src/connection_manager.c:1818 (libsysrepo.so+0x0000000aba2b)

  Previous read of size 8 at 0x7d0c0000df10 by thread T4:
    #0 sr_cbuff_items_in_queue /opt/fork/sysrepo/src/common/sr_data_structs.c:541 (libsysrepo.so+0x00000005bdd4)
    #1 rp_worker_thread_execute /opt/fork/sysrepo/src/request_processor.c:3659 (libsysrepo.so+0x0000000ce1c0)

  Location is heap block of size 40 at 0x7d0c0000def0 allocated by main thread:
[...]

PR #806, opened 2017-04-21
BUGFIX for data race

mislavn:


mislavn:
Bug fix for the data race in the issue https://github.com/sysrepo/sysrepo/issues/804

PR #808, opened 2017-04-22
BUGFIX do not rtn err for empty data when get nodes by schema

fanchanghu:
Hi,
The `get` operation will return an internal error when data is empty and schema has state node, such as follow module:
```
module test{
	namespace "urn:it:params:xml:ns:yang:test";
	prefix "test";
    	
	container C1{
	    presence "enable C1";
		list L1{
			key "name";
			leaf name{
				type string;
			}
			leaf value{
				type string;
				config false;
			}
		}
	}
}
```
if you do not configure any data and execute `get --filter-xpath=/test:C1` in netopeer, you will get an error like this:
```
> get --filter-xpath=/test:C1
ERROR
        type:     application
        tag:      operation-failed
        severity: error
        message:  Getting items (/test:C1//.) from sysrepo failed (Sysrepo-internal error).

> 
```

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -6571,10 +6571,14 @@ dm_get_nodes_by_schema(dm_session_t *session, const char *module_name, const str
     rc = dm_get_data_info(session->dm_ctx, session, module_name, &di);
     CHECK_RC_MSG_RETURN(rc, "Get data info failed");
 
-    *res = lyd_find_instance(di->node, node);
-    if (NULL == *res) {
-        SR_LOG_ERR("Failed to find nodes %s in module %s", node->name, module_name);
-        rc = SR_ERR_INTERNAL;
+    if (di->node == NULL) {
+        *res = ly_set_new();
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
I agree that *sysrepo* should not produce internal errors when there are no data for a module, but not only when it has some state nodes (which is what you said this solves). From the commit it seems to me it could fix this problem for all the modules, not just those with some state nodes, but maybe you know whether it does or not?

Regards,
Michal

fanchanghu:
I'm not sure whether does it produce this error when there are no state nodes in schema, but it seems this push request can only fix those with some state nodes, as the function `dm_get_nodes_by_schema` can only be called when there are some state nodes in schema.
Have I missed something?

michalvasko:
Hi,
you are right, this is only for state data, I mixed it. So it is fine, but we want to wait with the merge, because it would restart tests and they would fail now.

Regards,
Michal

Issue #798, opened 2017-04-20
Error importing data in to ietf-interfaces -  Failed to append data tree bbf-fastdsl


darshana-jk:
I'm running the devel branches of all packages (sysrepo, netopeer2, libyang and libnetconf2) and the source is up-to-date.

Initially I can load the bbf-vdsl modules  [bbf-vdsl YANG](https://github.com/BroadbandForum/yang/blob/master/standard/interface/bbf-vdsl.yang) and they are loaded and initialized successfully (I had to do a minor patch to load it: [bbf-vdsl-pointers.yang.patch.txt](https://github.com/sysrepo/sysrepo/files/941162/bbf-vdsl-pointers.yang.patch.txt)). 
```
sudo sysrepoctl -l
Sysrepo schema directory: /usr/local/etc/sysrepo/yang/
Sysrepo data directory:   /usr/local/etc/sysrepo/data/
(Do not alter contents of these directories manually)

Module Name                   | Revision   | Data Owner          | Permissions | Submodules                    | Enabled Features
---------------------------------------------------------------------------------------------------------------------------------
ietf-netconf-acm              | 2012-02-22 | root:root           | 666         |                               |
ietf-netconf-notifications    | 2012-02-06 | root:root           | 666         |                               |
ietf-netconf                  | 2011-06-01 | root:root           | 666         |                               |
nc-notifications              | 2008-07-14 | root:root           | 666         |                               |
notifications                 | 2008-07-14 | root:root           | 666         |                               |
iana-if-type                  | 2014-05-08 |                     |             |                               |
ietf-interfaces               | 2014-05-08 | root:root           | 666         |                               |
ietf-yang-types               | 2013-07-15 |                     |             |                               |
bbf-yang-types                | 2016-07-18 |                     |             |                               |
bbf-fastdsl                   | 2016-07-18 |                     |             |                               |
[...]

PR #810, opened 2017-04-24
Implemented flag changes and fixes

michalvasko:
New installed flag added.
Refs #798

### DIFF
--- a/src/sysrepo.proto
+++ b/src/sysrepo.proto
@@ -244,12 +244,13 @@ message Schema {
   required string module_name = 1;           /**< Name of the module. */
   required string ns = 2;                    /**< Namespace of the module. */
   required string prefix = 3;                /**< Prefix of he module. */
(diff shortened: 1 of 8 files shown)


src/sysrepo.proto
```diff
@@ -244,12 +244,13 @@ message Schema {
   required string module_name = 1;           /**< Name of the module. */
   required string ns = 2;                    /**< Namespace of the module. */
   required string prefix = 3;                /**< Prefix of he module. */
-  required bool implemented = 4;             /**< TRUE for explicitly installed modules. */
+  required bool installed = 4;               /**< TRUE for explicitly installed modules. */
+  required bool implemented = 5;             /**< TRUE for implemented modules (do not have to be installed). */
```
rkrejci: Please add mroe information when the not-installed module is actually implemented. Here, or/and in module_dependencies.h.

PR #824, opened 2017-04-27
Travis CI build: use protobuf's and protobuf-c's releases instead of …

rkrejci:
…rolling on git

Avoid compatibility issues between the protobuf and protobuf-c projects
and use stable packages that do not change.


smalls12:
I couldn't find any documentation regarding which version of protobuf-c to use.

I found this which highlights that Travis is pulling down tar tarball of protobuf-c v1.2.1.
I assume that means that you officially support sysrepo with protobutf-c v1.2.1.

Any plans on upgrading to protobuf-c v1.3.0+ ?

Issue #788, opened 2017-04-12
Add SOVERSION to shared lib to follow linux conventions


ykoehler:
The shared libraries shared by this package are not following Linux convention and only output a lib*.so file.  The SOVERSION target properties should be specified so that the proper lib*.so.<maj>, lib*.so.<maj>.<min> and lib*.so.<maj>.<min>.<patch> symlinks gets generated.

rkrejci:
@mislavn - is it reasonable to add also soname version into the binding libraries created by SWIG? I've added it only into the main libsysrepo.so.

mislavn:
> @mislavn - is it reasonable to add also soname version into the binding libraries created by SWIG? I've added it only into the main libsysrepo.so.

You are right, I will make a pull request.

rkrejci:
@ykoehler, could you please describe your usecase to have soname set in libsysrepo library? I mean, do you have some issue with not having soname or is it just "follow common convention"? The libsysrepo is used to communicate with the main sysrepod daemon, so it may be a bad idea to allow having multiple libsysrepo version in use and communicate with sysrepod of a different version.

ykoehler:
The issue I have with the lib missing soname is that I use this project inside Yocto, and Yocto will "split" installed files into multiple .rpm packages, and it doesn't recognise the .so file unless it follow the convention.  The file therefore is not part of the main *.rpm, and get move to the *-devel.rpm.  With the SONAME, the lib is then automatically put in the main *.rpm as expected.

rkrejci:
Ok, proposed solution:
- use soname, so app always links the correct library version. We will get this:
```
libsysrepo.so -> libsysrepo.so.0.5.11
libsysrepo.so.0.5 -> libsysrepo.so.0.5.11
libsysrepo.so.0.5.11
```
- send version number in `sr_connect()` (client side) and verify it in somewhere in the connection manager (in sysrepod process) to have the same (soname) version on both sides.

Another approach is to avoid SONAME in libsysrepo (Yocto as well as any other RPM builder should be then forced to put the .so into the correct package) and having instead a check for build-time and run-time libsysrepo version compatibility (version check in `sr_connect()` is necessary anyway). But the approach with SONAME seems more generic and standard.

Any thoughts or comments? Did I miss something?

rkrejci:
small correction, symlinks are actually
```
libsysrepo.so -> libsysrepo.so.0.5
libsysrepo.so.0.5 -> libsysrepo.so.0.5.11
[...]

PR #825, opened 2017-04-28
Versioning

rkrejci:


rkrejci:
I'm afraid I'm not getting your point. What do you think is a problem here?

Anyway, rethinking the problem again - is it a good idea to have `SONAME` in this library? Its purpose is actually to allow to have multiple versions of the library installed in the system at once. But `libsysrepo.so` is used to communicate with `sysrepod` (or to access datastore). So having multiple (soname) versions of the library in the system can cause that a client using version X can try to communicate with the sysrepod version Y. Which may cause problems. Maybe we need some other mechanism for version checking, what do you think?

antonio-paunovic:
Not exactly a problem, but it seems unconventional.

Rethinking the problem, you're right. `libsysrepo` is linked to sysrepod, sysrepocfg, etc. during the build. They will all use same version and I can't think of reason why would anybody want situation where parts of Sysrepo infrastructure (also plugin .so files) could have different versions.

rkrejci:
Well, I can imagine that it can happen that plugins/connected apps (which are not part of the sysrepo project and built at different time) can be built with different version. But then the issue is not solved by SONAME, we will need a mechanism on the client side to detect this. 

I'm going to prepare something and will wait also for feedback in #788, but currently this pull request is not ready for merge.

antonio-paunovic:
OK. Another thing I noticed is that libyang and this pull use major and minor version for SOVERSION and usually (like in `/usr/lib` on Linux) libs tend to only use major version for it. I can't find out why, since minor version also describes API.
This is what I meant by unconventional before. Maybe not very important difference at the moment.

So instead of:
`
set(SYSREPO_VERSION ${SYSREPO_MAJOR_VERSION}.${SYSREPO_MINOR_VERSION}.${SYSREPO_MICRO_VERSION})
set(SYSREPO_SOVERSION ${SYSREPO_MAJOR_VERSION}.${SYSREPO_MINOR_VERSION})
`

we would have: 
`
set(SYSREPO_VERSION ${SYSREPO_MAJOR_VERSION}.${SYSREPO_MINOR_VERSION}.${SYSREPO_MICRO_VERSION})
set(SYSREPO_SOVERSION ${SYSREPO_MAJOR_VERSION})
`

Issue #828, opened 2017-04-28
Build fails with non-existing out-of-tree REPOSITORY_LOC due to swig+cpp's POST_BUILD


jktjkt:
Here's how a from-the-scratch shared-nothing builds fails:
```
2017-04-28 16:15:19,282 [output] FAILED: swig/libSysrepo-cpp.so 
2017-04-28 16:15:19,282 [output] : 
&& /opt/rh/devtoolset-6/root/usr/bin/c++  -fPIC -Wall -Wpedantic -std=c++11 -fno-strict-aliasing -g -O0   -shared -Wl,-soname,libSysrepo-cpp.so -o swig/libSysrepo-cpp.so swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Sysrepo.cpp.o swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Connection.cpp.o swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Session.cpp.o swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Struct.cpp.o swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Xpath.cpp.o swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Tree.cpp.o swig/CMakeFiles/Sysrepo-cpp.dir/cpp/src/Internal.cpp.o  src/libsysrepo.so -lrt -lpthread /home/turbo-hipster/target/lib/libredblack.so -lev -lprotobuf-c /home/turbo-hipster/target/lib64/libyang.so -Wl,-rpath,/home/turbo-hipster/jobs/e8d4791723d1/30/230/1/check/check-czechlight-devtoolset6-el7/1cab956/sysrepo/sysrepo/build/src:/home/turbo-hipster/target/lib:/home/turbo-hipster/target/lib64: 
&& cd /home/turbo-hipster/jobs/e8d4791723d1/30/230/1/check/check-czechlight-devtoolset6-el7/1cab956/sysrepo/sysrepo/build/swig 
[...]

PR #831, opened 2017-05-02
BUGFIX for cmake swig test repository check

mislavn:
This is a bug fix for the issue #828 .


mislavn:
> I can confirm that this fixes the linked bug. Can you perhaps link to it from the commit message? That way, one doesn't have to access github comments (which are essentially opaque to git) to find the reason for this change.

Done.

PR #833, opened 2017-05-03
Version check

rkrejci:
- add SONAME to libraries and bump to version 0.6.0 following the versioning policy
- check version compatibility when client connects to the sysrepod
- add missing dependency in libSysrepo-cpp build


inc/sysrepo.h
```diff
@@ -241,6 +241,7 @@ typedef enum sr_error_e {
     SR_ERR_LOCKED,             /**< Requested resource is already locked. */
     SR_ERR_TIME_OUT,           /**< Time out has expired. */
     SR_ERR_RESTART_NEEDED,     /**< Sysrepo Engine restart is needed. */
+    SR_ERR_VERSION_MISSMATCH,  /**< Incompatible client library used to communicate with sysrepo. */
```
antonio-paunovic: This could be typo. Missmatch is spelled mismatch.

rkrejci:
It is because `POST_BUILD` command connected with `Sysrepo-cpp` target uses sysrepoctl and it happened that sysrepoctl was not yet built and it all failed, like in [this Travis build](https://travis-ci.org/sysrepo/sysrepo/jobs/228247686). As I now look at `libsysrepoLua` target, yes, the dependency is supposed to be added there too, I'm going to add it.

PR #838, opened 2017-05-04
Fix verifying version on sysrepo socket connections

rkrejci:
Remove optimization of not sending version-verify request on local
(intra-process) communication. It is not able to get this information
surely on both the communication sides so the version-verify request
is now sent/expected whenever a new connection is created.

Fixes sysrepo/sysrepo#836


fabioantonini:
It worked fine. Thanks

PR #846, opened 2017-05-11
sysrepocfg not-strict flag

michalvasko:
### Description
In sysrepocfg import operation the user in majority of the cases will use a file with only the data they want to import - silently skipping unknown elements can hide a problem. In case it is intentional, it can be specified by a flag.


rkrejci:
Before the merge I have moved version on devel from 0.6.0 to 0.6.1 to mark the change in tool's functionality.

PR #857, opened 2017-05-19
Overriding the debug/release build detection

jktjkt:
The old system piggy-backed on the ``CMAKE_BUILD_TYPE`` variable in order to guess whether this is a systemwide installation or a developer's build. I am aware of this (I modified it before to support ``ReleaseWithDebInfo``), yet I lost some time re-flashing a custom board with a firmware that has a broken installation due to a use of ``-DCMAKE_BUILD_TYPE=Debug``.
    
That is a valid use case, CMake's build type is *the* way of configuring whether to use debug symbols or not. The choice of debug symbols (and asserts, for that matter) is orthogonal to the choice of where to install the repository data, and where to create PID files.
    
This patch at least adds an override which makes it possible to force a production installation even when building with the Debug CMake build type. I'm not changing the default even though I think that we might be slowly moving to a point where it makes more sense to pester the developers with a custom CMake option rather than letting the packagers implement that.


jktjkt:
Done, 6a80df2.

PR #871, opened 2017-05-29
Honor the option CALL_TARGET_BINS_DIRECTLY

ykoehler:
In yocto we compile the project for a different architecture, as such executing any TARGET binary on the HOST is bad since it may not be CPU compatible. An option in the root CMakeList was created to call those binaries or not at install time based on the option value, but the examples/CMakeList.txt was not honoring it.


ykoehler:
Wait before merging, I think this is broken, the install-yang.sh doesn't seems to contain the instructions for the examples.

jktjkt:
@ykoehler , I've added some options which make it possible to call host's binaries during build, https://github.com/sysrepo/sysrepo/pull/946 . It's been merged into devel earlier today.

Issue #839, opened 2017-05-05
Getting compiliation error  with the latest code


rosy8:
Hi,

I am getting the following compilation error with the latest code.

[ 61%] Building C object src/CMakeFiles/SR_ENGINE.dir/rp_dt_filter.c.o
[ 64%] Building C object src/CMakeFiles/SR_ENGINE.dir/data_manager.c.o
gitcode_devel/sysrepo/src/data_manager.c: In function ‘dm_release_tmp_ly_ctx’:
<b>gitcode_devel/sysrepo/src/data_manager.c:567:20: error: ‘LY_INTERNAL_MODULE_COUNT’ undeclared (first use in this function) </b>
     uint32_t idx = LY_INTERNAL_MODULE_COUNT;
                    ^
gitcode_devel/sysrepo/src/data_manager.c:567:20: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [src/CMakeFiles/SR_ENGINE.dir/data_manager.c.o] Error 1
make[1]: *** [src/CMakeFiles

I just made the below changes 
//uint32_t idx = LY_INTERNAL_MODULE_COUNT;
 uint32_t idx = 5;  (As specified in side libyang.h file    #define LY_INTERNAL_MODULE_COUNT 5 )

With this compilation was successful but getting the following error during installation..
[root@18-88 lib64]# /usr/bin/sysrepoctl --install --yang=/etc/libnetconf/yang-models/ietf-inet-types@2013-07-15.yang
Installing a new module from file '/etc/libnetconf/yang-models/ietf-inet-types@2013-07-15.yang'...
Installing the YANG file to '/etc/sysrepo/yang/ietf-inet-types@2013-07-15.yang'...
<b>[ERR] (srctl_ly_log_cb:487) libyang: Schema node not found.</b>
[ERR] (md_lyd_new_path:413) Failed to set installed flag for module 'ietf-inet-types@2013-07-15': Schema node not found.
Error: Unable to insert the module into the dependency graph.
Reverting the install operation...
Deleted the schema file '/etc/sysrepo/yang/ietf-inet-types@2013-07-15.yang'.
Install operation failed.
[root@18-88 lib64]#

Please let me know how to overcome this issue.
Thanks in advance.

Regards,
Rosy

michalvasko:
Hi Rosy,
no modifications are necessary, just update *libyang*.

Regards,
Michal

rosy8:
Hi Michal,

[...]

PR #875, opened 2017-06-02
Insert mod fix

rkrejci:
### Description
When a module A is installed that imports and uses a grouping from B and in the grouping there is a leafref to module C (imported in B), A is implemented, B is imported, and C should also be implemented, but it was not (it was assumed that dependencies of imported modules will always be only imported).

Also, `sysrepoctl` installed and printed information about data files for only imported modules, which is wrong.

Replaces #872

### Closure
Refs #839

### DIFF
--- a/deploy/travis/install-libs.sh
+++ b/deploy/travis/install-libs.sh
@@ -21,7 +21,7 @@ if [ ! -d "$INSTALL_PREFIX_DIR/lib" ]; then
     git clone git://git.cryptomilk.org/projects/cmocka.git
     cd cmocka ; mkdir build; cd build
     cmake -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX_DIR -DCMAKE_C_FLAGS="-DUNIT_TESTING_DEBUG" ..
-    make -j2 && make install
+    make -j2 > /dev/null && make install
     cd ../..
 
     # protobuf 
@@ -29,7 +29,7 @@ if [ ! -d "$INSTALL_PREFIX_DIR/lib" ]; then
     tar -xzf v3.2.0.tar.gz
     cd protobuf-3.2.0
     ./autogen.sh && ./configure --prefix=$INSTALL_PREFIX_DIR 
-    make -j2 && make install
+    make -j2 > /dev/null && make install
     cd ..
 
     # protobuf-c
@@ -37,7 +37,7 @@ if [ ! -d "$INSTALL_PREFIX_DIR/lib" ]; then
     tar -xzf v1.2.1.tar.gz
     cd protobuf-c-1.2.1
     ./autogen.sh && ./configure --prefix=$INSTALL_PREFIX_DIR
-    make -j2 && make install
+    make -j2 > /dev/null && make install
     cd ..
 
 else
@@ -54,6 +54,6 @@ else
 fi
 cd libyang ; mkdir build ; cd build
 cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_BUILD_TESTS=OFF ..
-make -j2 && sudo make install
+make -j2 > /dev/null && sudo make install
 cd ../..
 
--- a/src/executables/sysrepoctl.c
+++ b/src/executables/sysrepoctl.c
@@ -181,6 +181,7 @@ srctl_list_modules()
         for (size_t i = 0; i < schema_cnt; i++) {
             printf("%-30s| %-11s| ", schemas[i].module_name,
                     (NULL == schemas[i].revision.revision ? "" : schemas[i].revision.revision));
+
             /* print conformance */
             if (schemas[i].installed) {
                 printf("Installed   | ");
@@ -189,18 +190,30 @@ srctl_list_modules()
             } else {
                 printf("Imported    | ");
             }
+
             /* print owner */
-            srctl_print_module_owner(schemas[i].module_name, buff);
+            if (schemas[i].implemented) {
+                srctl_print_module_owner(schemas[i].module_name, buff);
+            } else {
+                buff[0] = '\0';
+            }
             printf("%-20s| ", buff);
+
             /* print permissions */
-            srctl_print_module_permissions(schemas[i].module_name, buff);
+            if (schemas[i].implemented) {
+                srctl_print_module_permissions(schemas[i].module_name, buff);
+            } else {
+                buff[0] = '\0';
+            }
             printf("%-12s| ", buff);
+
             /* print submodules */
             size_t printed = 0;
             for (size_t j = 0; j < schemas[i].submodule_count; j++) {
                 printed += printf(" %s", schemas[i].submodules[j].submodule_name);
             }
             for (size_t j = printed; j < 30; j++) printf(" ");
+
             /* print enabled features */
             printf("|");
             for (size_t j = 0; j < schemas[i].enabled_feature_cnt; j++) {
--- a/src/executables/sysrepoctl.c
+++ b/src/executables/sysrepoctl.c
@@ -848,7 +848,7 @@ srctl_data_install(const struct lys_module *module, const char *owner, const cha
     int ret = 0, rc = SR_ERR_OK;
 
     /* install data files only if module can contain any data */
-    if (sr_lys_module_has_data(module)) {
+    if (module->implemented && sr_lys_module_has_data(module)) {
         printf("Installing data files for module '%s'...\n", module->name);
         ret = srctl_data_files_apply(module->name, srctl_file_create, NULL, false);
         if (0 != ret) {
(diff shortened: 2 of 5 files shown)


fabioantonini:
HI Radek

I have just pulled the last changes from sysrepo (pull request #875) and 
now sysrepod fails to start

# sysrepod -d -l4
[DBG] (sr_log_stderr:141) Setting log level for stderr logs to 4.
[DBG] (cm_init:1910) Connection Manager init started.
[DBG] (sr_cbuff_init:463) Initiating circular buffer for 10 elements.
[DBG] (sm_init:290) Session Manager initialized successfully, ctx=0x9e5138.
[DBG] (cm_init:1935) Connection Manager sm_init started.
[DBG] (cm_server_init:163) Initializing sysrepo server at 
socket=/tmp/sysrepod.sock
[DBG] (cm_init:1944) Connection Manager cm_server_init started.
[DBG] (rp_init:3821) Request Processor init started.
[DBG] (sr_cbuff_init:463) Initiating circular buffer for 10 elements.
[DBG] (rp_init:3857) initializing Notification Processor
[DBG] (sr_cbuff_enqueue:522) Circular buffer enqueue to position=0, 
current count=1.
[DBG] (np_setup_notif_store_cleanup_timer:710) Notification store 
cleanup timer set up for 600 seconds.
[DBG] (np_init:803) Notification Processor initialized successfully.
[INF] (dm_init:1958) Initializing Data Manager, 
schema_search_dir=/etc/configs/sysrepo/current/yang/, 
data_search_dir=/etc/configs/sysrepo/current/data/
[INF] (nacm_init:827) Initializing NACM.
[DBG] (pm_load_data_tree:231) Persist data file 
'/etc/configs/sysrepo/current/data/ietf-netconf-acm.persist' does not exist.
[DBG] (pm_load_data_tree:233) No persistent data for module 
'ietf-netconf-acm' will be loaded.
[WRN] (dm_apply_persist_data_for_model:881) Persist file for module 
ietf-netconf-acm does not exist.
*[ERR] (nacm_load_config:488) Unable to open the NACM startup datastore 
('/etc/configs/sysrepo/current/data/ietf-netconf-acm.startup'): No such 
file or directory.**
**[ERR] (dm_init:2023) Failed to initialize NACM context.**
**[ERR] (rp_init:3876) Data Manager initialization failed.*
[DBG] (np_cleanup:818) Notification Processor cleanup requested.
[DBG] (np_notification_store_cleanup:1866) Notification store cleanup 
requested.
[DBG] (np_get_all_notification_files:597) Listing notification 
directories in '/etc/configs/sysrepo/current/data/notifications/'.
[DBG] (np_get_all_notification_files:614) Listing notification directory 
'ietf-netconf-notifications'.
[DBG] (np_get_notification_files:551) Listing notification data files 
for 'ietf-netconf-notifications' modified from '1970-01-01 00:00' to 
'1970-01-12 02:41'.
[ERR] (cm_init:1969) Cannot initialize Request Processor.
[DBG] (rp_cleanup:3920) Request Processor cleanup started, requesting 
cancel of each worker thread.
[DBG] (rp_cleanup:3956) Request Processor cleanup finished.
[DBG] (sm_cleanup:303) Session Manager cleanup requested, ctx=0x9e5138.
[DBG] (sr_cbuff_dequeue:538) Circular buffer dequeue, new buffer head=1, 
count=0.
[INF] (cm_cleanup:2023) Connection Manager successfully destroyed.
[ERR] (main:119) Unable to initialize Connection Manager: Input/output 
error.
[INF] (cm_cleanup:2023) Connection Manager successfully destroyed.
[INF] (main:144) Sysrepo daemon terminated.

Here below the sysrepoctl output

Module Name                   | Revision   | Conformance | Data 
Owner          | Permissions | Submodules                    | Enabled 
Features
-----------------------------------------------------------------------------------------------------------------------------------------------
ietf-netconf                  | 2011-06-01 | Installed   | 
root:root           | 777         | |
*ietf-netconf-acm              | 2012-02-22 | Imported* 
|                     | |                               |
ietf-netconf-notifications    | 2012-02-06 | Installed   | 
root:root           | 777         | |
ietf-yang-types               | 2013-07-15 | Installed 
|                     | |                               |
ietf-inet-types               | 2013-07-15 | Installed 
|                     | |                               |
iana-if-type                  | 2014-05-08 | Installed 
|                     | |                               |
ietf-interfaces               | 2014-05-08 | Implemented | 
root:root           | 777         | |
ietf-ip                       | 2014-06-16 | Installed 
|                     | |                               |
ietf-netconf-server           | 2016-11-02 | Installed   | 
root:root           | 777         | | listen ssh-listen tls-listen 
call-home ssh-call-home tls-call-home
ietf-x509-cert-to-name        | 2014-12-10 | Imported 
|                     | |                               |
ietf-ssh-server               | 2016-11-02 | Imported 
|                     | |                               |
ietf-keystore                 | 2016-10-31 | Implemented | 
root:root           | 777         | |
ietf-tls-server               | 2016-11-02 | Imported 
|                     | |                               |
ietf-system                   | 2014-08-06 | Installed   | 
root:root           | 777         | | authentication local-users

If I try to install ietf-netconf-acm as the first yang module  then it 
appears as 'installed' in sysrepoctl's output and everything starts fine

Is it required an order of installation for yang modules? Previously the 
pull request everything worked fine in my opinion.

regards
fabio

Il 06/06/2017 11:38, Radek Krejci ha scritto:

rkrejci:
Hi Fabio, did you just updated the tools and kept the repository? Probably it will be needed to completely initiate sysrepo repository files (not only the data repository) by removing the repository directory and rebuild and reinstall the sysrepo.

fabioantonini:
Hi Radek
I have removed everything under my repository directory and reinstalled everything from the beginning as I usually do with a new build of the package. In other word I have removed completely data/yang directories and all the persist/startup/running files and deployed the package.
I want to remark that if I try to install ietf-netconf-acm as first yang module everything works fine. In that case ietf-netconf-acm appears as 'installed'. When it is 'imported' the problem occurs.
thanks a lot
fabio

rkrejci:
I'm not sure what do you mean by "I try to install ietf-netconf-acm". It should be installed automatically by the cmake (as part of make). The schema list is the result if you do (with no repository directory)
```
$ cmake ..
$ make
```
?

rkrejci:
It shouldn't be necessary to have a specific order, but since you seem to have enabled NACM, the ietf-netconf-acm` must be explicitely installed (not just imported). But it should be done as part of the make (for the local repository in build directory)

fabioantonini:
I'm working in a cross-environment so the yang modules installation occurs on target when the package is deployed. I have a script that list all the ietf yang modules and tries to install them. The first yang file is ietf-netconf that imports ietf-netconf-acm.yang. Then the script lists ietf-netconf-acm.yang but it has been already imported under the yang directory so its installation is skipped. The assumption is that if a yang module has been imported it doesn't need to be installed. So far the script worked fine, but probably we need to change the script logic.

jktjkt:
@fabioantonini I added the ``CALL_TARGET_BINS_DIRECTLY`` CMake option for this purpose. Do you use it? Does it offer the functionality that you require?

fabioantonini:
Hi Jan
Thanks a lot for your advise. I noticed your CMake option but I didn't have the time to go more in depth. 
It could offer what I require. I'll keep you informed
regards
fabio

rkrejci:
libyang differs between the implemented and imported schemas and it is better reflected in sysrepo now - the data (,rpcs and notifications) are ignored (hidden for subscription) in the imported modules, so such modules are used only as a source of groupings, typedefs or identities for the modules which import it. In some circumstances, sysrepo allowed you to subscribe even for the data from not implemented (imported) schemas. This should be fixed and, so it is necessary to better distinguish between these types now, there are actualy 3 of them (the note is also in man page):
* imported - implicitely inserted as import of another installed schema
* implemented - implicitely inserted, but it is affected by other installed module (e.g. by augmenting) to be also implemented
* installed - explicitely inserted, so it is implemented (some subscription is allowed/expected runtime)

Note that `ietf-netconf` is, by default, just imported, so if you want to replicate repository state in your environment, you are not supposed to install it explicitely (it will be added implicitely by the module which needs it). On the other hand, `ietf-netconf-acm` is installed so it is supposed to be installed explicitely.

In your case you can filter only the installed modules and install them via sysrepoctl in any order. The rest schemas should be inserted into the sysrepo database automatically.

fabioantonini:
Hi Radek/Jan
I have adopted the advised approach and it worked really fine. Now we are using the built script install-yang.sh to install the yang files.
We have noticed that the same approach has not been adopted under sysrepo/example. Could it be an improvemement?
thanks a lot

fabio

jktjkt:
There's a WIP at https://github.com/sysrepo/sysrepo/pull/871; I haven't checked it yet.

PR #889, opened 2017-07-04
libyang api changes and xpath fixes

### LINKED ISSUE #888: Recent changes in libyang cause the sysrepo build to fail
It looks like https://github.com/CESNET/libyang/commit/15104310a3e68368cce3b45f18090363849b37ae introduced the following failure in sysrepo's make install:
```
-- Exec: /opt/dev/sysrepo/build/src/sysrepoctl --install --yang=/opt/dev/sysrepo/examples/yang/turing-machine.yang --permissions=666; /opt/dev/sysrepo/build/src/sysrepocfg --import=/opt/dev/sysrepo/examples/yang/turing-machine.data.xml --datastore=startup --format=xml --level=0 turing-machine
CMake Error at examples/cmake_install.cmake:39 (message):
  Error: 1
Call Stack (most recent call first):
  cmake_install.cmake:170 (include)



make: *** [install] Error 1
```

A later commit https://github.com/CESNET/libyang/commit/9ff79aa3c9303c6eb6945286653a0d34ddc07623 also causes make to fail with:
```
[  3%] Building C object src/CMakeFiles/COMMON.dir/common/sr_utils.c.o
/opt/dev/sysrepo/src/common/sr_utils.c: In function 'sr_find_schema_node':
/opt/dev/sysrepo/src/common/sr_utils.c:2327:47: warning: passing argument 2 of 'lys_find_xpath' from incompatible pointer type [-Wincompatible-pointer-types]
     struct ly_set *set = lys_find_xpath(NULL, node, expr, options);
                                               ^
In file included from /usr/local/include/libyang/libyang.h:20:0,
                 from /opt/dev/sysrepo/src/common/sr_utils.c:36:
/usr/local/include/libyang/tree_schema.h:2117:16: note: expected 'const char *' but argument is of type 'const struct lys_node *'
 struct ly_set *lys_find_xpath(const struct lys_node *ctx_node, const char *expr, int options);
                ^
/opt/dev/sysrepo/src/common/sr_utils.c:2327:53: warning: passing argument 3 of 'lys_find_xpath' makes integer from pointer without a cast [-Wint-conversion]
     struct ly_set *set = lys_find_xpath(NULL, node, expr, options);
                                                     ^
In file included from /usr/local/include/libyang/libyang.h:20:0,
                 from /opt/dev/sysrepo/src/common/sr_utils.c:36:
/usr/local/include/libyang/tree_schema.h:2117:16: note: expected 'int' but argument is of type 'const char *'
 struct ly_set *lys_find_xpath(const struct lys_node *ctx_node, const char *expr, int options);
                ^
/opt/dev/sysrepo/src/common/sr_utils.c:2327:26: error: too many arguments to function 'lys_find_xpath'
     struct ly_set *set = lys_find_xpath(NULL, node, expr, options);
                          ^
In file included from /usr/local/include/libyang/libyang.h:20:0,
                 from /opt/dev/sysrepo/src/common/sr_utils.c:36:
/usr/local/include/libyang/tree_schema.h:2117:16: note: declared here
 struct ly_set *lys_find_xpath(const struct lys_node *ctx_node, const char *expr, int options);
                ^
src/CMakeFiles/COMMON.dir/build.make:117: recipe for target 'src/CMakeFiles/COMMON.dir/common/sr_utils.c.o' failed
make[2]: *** [src/CMakeFiles/COMMON.dir/common/sr_utils.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:138: recipe for target 'src/CMakeFiles/COMMON.dir/all' failed
make[1]: *** [src/CMakeFiles/COMMON.dir/all] Error 2
make: *** [all] Error 2
Makefile:127: recipe for target 'all' failed
```

Make and make install are both fine at https://github.com/CESNET/libyang/commit/41cb7c3546bc714b56b87bf1093eda563d0b3027

michalvasko:
### Description
Recent libyang fixes corrected some logical errors in XPath paths and addressing, which were also present both indirectly and directly in sysrepo. It was incorrectly implemented and understood, nodes in XPath do not inherit the local module from their parent, but always consider the "context" module to be local. For data, context module is the module of the context node (more about various context nodes in RFC 7950) and for schema, context module is the module where the path is defined.

Also, one libyang function API changed.

### Test case
Affected tests modified and fixed.

### Closure
Fixes #888

### DIFF
--- a/src/common/sr_utils.c
+++ b/src/common/sr_utils.c
@@ -1443,7 +1443,7 @@ sr_api_variant_from_str(const char *api_variant_str)
 /**
  * @brief Copy and convert content of a libyang node and its descendands into a sysrepo tree.
  *
- * @param [in] parent Parent node.
+ * @param [in] top_parent Top-level parent node.
  * @param [in] node libyang node.
  * @param [in] depth Depth of the node relative to the root node.
  * @param [in] slice_offset Number of child nodes of the chunk root to skip.
@@ -1453,7 +1453,7 @@ sr_api_variant_from_str(const char *api_variant_str)
  * @param [out] sr_tree Returned sysrepo tree.
  */
 static int
-sr_copy_node_to_tree_internal(const struct lyd_node *parent, const struct lyd_node *node, size_t depth,
+sr_copy_node_to_tree_internal(const struct lyd_node *top_parent, const struct lyd_node *node, size_t depth,
          size_t slice_offset, size_t slice_width, size_t child_limit, size_t depth_limit,
          sr_tree_pruning_cb pruning_cb, void *pruning_ctx, sr_node_t *sr_tree)
 {
@@ -1505,7 +1505,7 @@ sr_copy_node_to_tree_internal(const struct lyd_node *parent, const struct lyd_no
     sr_tree->dflt = node->dflt;
 
     /* set module_name */
-    if (NULL == parent || lyd_node_module(parent) != lyd_node_module(node)) {
+    if (NULL == top_parent || lyd_node_module(top_parent) != lyd_node_module(node)) {
         rc = sr_node_set_module(sr_tree, lyd_node_module(node)->name);
         CHECK_RC_MSG_GOTO(rc, cleanup, "Failed to set module of a sysrepo node.");
     }
@@ -1532,7 +1532,7 @@ sr_copy_node_to_tree_internal(const struct lyd_node *parent, const struct lyd_no
                 if (SR_ERR_OK != rc) {
                     goto cleanup;
                 }
-                rc = sr_copy_node_to_tree_internal(node, child, depth + 1, slice_offset, slice_width,
+                rc = sr_copy_node_to_tree_internal(top_parent ? top_parent : node, child, depth + 1, slice_offset, slice_width,
                         child_limit, depth_limit, pruning_cb, pruning_ctx, sr_subtree);
                 if (SR_ERR_OK != rc) {
                     goto cleanup;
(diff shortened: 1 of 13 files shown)


fanchanghu:
Hi,
It seems the mail system of my organization does not support pure-text mail, my reply about your mail seems does not display well in the mail-list, so I post my opinion here~~

I still think that the unprefix node in sysrepo API should use its parent's prefix. In fact, there are 3 cases using xpath:
1, xpath used in schema files, such as path of leafref, should use the module's prefix as default, this is required by RFC 7950
2, xpath used in xml files, such as value of instance-identifier, has no default prefix, all prefix must be specified, this is required by RFC 7950
3, xpath used in sysrepo API, such as the subscription interface, it is very special, I found it was called "node id" in the API, it should follow the JSON specification. RFC 7951 does not define common style about xpath, but it says something about the value of instance-identifier:

https://tools.ietf.org/html/rfc7951#section-6.11
```
   For example,

   /ietf-interfaces:interfaces/interface[name='eth0']/ietf-ip:ipv4/ip

   is a valid instance-identifier value because the data nodes
   "interfaces", "interface", and "name" are defined in the module
   "ietf-interfaces", whereas "ipv4" and "ip" are defined in "ietf-ip".
```

I think we should follow this format in the sysrepo API, so it is correct of the current implement.
@michalvasko @rkrejci

rkrejci:
Hi, thanks for this feedback! We have national holidays now and Michal will be back in office on next Monday. We had this issue with XPaths in schemas (must, when, leafrefs, augments) and it's possible that we are wrongly mixing schema and data xpaths (because of the current libyang API design). Lets discuss it later with Michal.

PR #900, opened 2017-07-14
tests: Do not assume a particular build directory layout

jktjkt:
CMake is quite happy to support an arbitrary location of source and build directories. This code assumed that we're always building within a one-level deep subdirectory (e.g., `build/`) of the top-level source directory. That's a wrong assumption; it's better to hard-code these paths into test binaries via CMake.


michalvasko:
Hi,
there are much more problems with the tests and even though it would be great if we could fix them all, it will not be anytime soon. Nevertheless, this is a step in the right direction, thanks.

Regards,
Michal

PR #923, opened 2017-08-23
Support handling of pending events with a custom event loop

jktjkt:
When sysrepo is running with its own event loop, it ensures that all pending events have been processed when the subscription manager gets destroyed. This patch adds an equivalent functionality for scenarios where the application provides its own event loop.

This functionality requires a real blocking call, so it is not enough to reuse the existing event handling that is based on asynchronous writes to a notification pipe. A new callback function was needed.

I had to add both a new global static to `client_library.c` and an extra member to the `cl_sm_ctx_t` because `sr_fd_watcher_init` and `cl_sm_cleanup` are defined in separate translation units.

Technically speaking, it's still possible to call `sr_fd_watcher_init` with a NULL synchronizing callback. However, doing so results in some events not being handled if, e.g., the event loop runs in a different thread and the subscription manager gets destroyed "too soon", prior to the event loop having handled "everything".

Cc: @TrinityCoder

### DIFF
--- a/examples/application_fd_watcher_example.c
+++ b/examples/application_fd_watcher_example.c
@@ -188,6 +188,12 @@ event_loop()
     } while ((SR_ERR_OK == rc) && !exit_application);
 }
 
+void
+flush_all_events()
+{
+    printf("We are single-threaded, no reason to wait for all pending events to be processed\n");
+}
+
 int
 main(int argc, char **argv)
 {
@@ -202,7 +208,7 @@ main(int argc, char **argv)
     }
 
     /* init app-local fd watcher */
-    rc = sr_fd_watcher_init(&poll_fd_set[0].fd);
+    rc = sr_fd_watcher_init(&poll_fd_set[0].fd, &flush_all_events);
     poll_fd_set[0].events = POLLIN;
     poll_fd_cnt = 1;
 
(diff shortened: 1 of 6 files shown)


michalvasko:
Hi,
firstly, sorry it took me this long to review your patch. It looks fine and will be merged, but we would like to wait around a week until we merge it. The reason is that we want to release a new sysrepo next week with only bugfixes, no API changes. Hope it is not a problem.

Regards,
Michal

jktjkt:
@michalvasko , no problem, and thanks for looking into this.

This actually required a follow-up fix, check the other commit I just pushed. Thanks to @TrinityCoder for the initial debugging.

I wonder if I took a reasonable approach here. What confuses me is that apparently the lifetime of the event loop is tied to the existance of any **subscription**, yet the FD handlers for custom watching are notified about any FD, including those for a **connection**, not just subscriptions. I would have expected the code to only tear down the event loop afetr the last connection gets closed, but my understanding of SR's internals is kinda small :).

michalvasko:
Hi,
unfortunately I do not know much about the low-level mechanisms either, so you are on your own here for now. However, it is possible I will look into it later, before the next major *sysrepo* release, and I will then consider what you are saying.

Regards,
Michal

PR #928, opened 2017-09-13
BUGFIX whole-module subscriptions did not match any notifications

### LINKED ISSUE #922: Unclear on how to listen for several notifications
After the recent change which unified the XPath formats, I no longer know how to subscribe to any notifications from a module. None of the following formats appears to work:

* `/module:*//.` which used to work, but now raises an error message `libyang error: Unexpected character(s) '` at registration time
* `/module:*//` (same error message)
* `/module:*` which succeeds in registering for the notifications, but fails to actually receive them at runtime

This is with latest libyang devel and sysrepo #921. Stuff works with CESNET/libyang@441f134 and sysrepo/sysrepo@d9f6a9d.

michalvasko:
### Description
Subscriptions for whole modules use a special format, which was not recognized and subscriptions were not delivered even when they should have been.

### Closure
Fixes #922

### DIFF
--- a/src/persistence_manager.c
+++ b/src/persistence_manager.c
@@ -1104,12 +1104,17 @@ pm_add_subscription(pm_ctx_t *pm_ctx, const ac_ucred_t *user_cred, const char *m
         SR_LOG_DBG("Removing all existing %s subscriptions from '%s' persist data tree.",
                 sr_subscription_type_gpb_to_str(subscription->type), module_name);
 
-        /* make sure there will be no illegal quotes */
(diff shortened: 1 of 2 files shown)


jktjkt:
I had to change our code from subscribing to `/myModule:*//.` which no longer works, to use just `myModule`. Is that by design?

michalvasko:
Hi,
yes, the path was changed because libyang no longer supports special XPath paths like `*` or `.`, documentation was also updated.

Regards,
Michal

PR #946, opened 2017-10-03
cross-building: Workaround calls to sysrepoctl/sysrepocfg

jktjkt:
Sysrepo has always wanted to call binaries which it has just produced. That works well when building natively, but it's a no-go when doing cross-compilation.

I've added a workaround earlier in commit 00025c49. One could use two options, `CALL_TARGET_BINS_DIRECTLY` and `INDIRECT_YANG_INSTALL_DIR` to produce a shell script which could be used to initialize `/etc/sysrepo` later on, perhaps on the target system. However, that is still not a complete fix for scenarios which intend to, e.g., produce a self-contained disk image to be flashed to an embedded device.

I'm using Buildroot which makes it super-easy to build a package twice -- once as a native, host package, and the second time as a cross-built package targeting another system with a possibly different architecture. This patch uses that feature. One can first build a special "throw-away" sysrepo which will only be used for its `sysrepoctl` and `sysrepocfg`. This host-side-only sysrepo should be built with an appropriate `REPOSITORY_LOC` to point to the target's `/etc/sysrepo`.

When building the second sysrepo that is supposed to run on the target, proceed as normally, but pass two additional arguments, `CALL_SYSREPOCTL_BIN` and `CALL_SYSREPOCFG_BIN` which point to the throwaway sysrepo. That way, these tools can be used for executing host's native code which however installs modules into the target's `/etc/sysrepo`. These tools can be also used for other packages such as Netopeer2.


CMakeLists.txt
```diff
@@ -42,6 +42,8 @@ set(CMAKE_C_FLAGS_DEBUG   "-g -O0")
 
 set(CALL_TARGET_BINS_DIRECTLY ON CACHE BOOL "Enable calling sysrepoctl/sysrepocfg directly at install time")
 set(INDIRECT_YANG_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/yang/" CACHE PATH "Where to store YANG files when not executing sysrepoctl directly")
+set(CALL_SYSREPOCTL_BIN "${CMAKE_BINARY_DIR}/src/sysrepoctl" CACHE PATH "Path to the sysrepoctl binary to be used at install time")
+set(CALL_SYSREPOCFG_BIN "${CMAKE_BINARY_DIR}/src/sysrepoctl" CACHE PATH "Path to the sysrepocfg binary to be used at install time")
```
michalvasko: You made a typo, this must be sysrepocfg binary, not sysrepoctl. That is why the tests failed.

michalvasko:
Hi,
seems useful, thanks. Please address what I wrote in the review and then we can merge it.

Regards,
Michal

jktjkt:
Fixed both. I've actually changed these in INSTALL.md because that's where most of the instructions are used, anyway.

BTW, it seems that netopeer2's build system picks these up without any extra work, at least with Buildroot, because the host's `path/to/sdk/bin` is in `$PATH`, and therefore it gets picked by CMake's `find_XXX`. It then fails because the build happens as non-root and there are some modules which want to be `root:root` and `0600`, but that's just an easy patch away :).

PR #949, opened 2017-10-04
Stop sysrepo-plugind from auto-launching the daemon, and fix systemd integration

jktjkt:
Check commit messages for details. In a nutshell, this stops `sysrepo-plugind` from auto-launching the main `sysrepod` daemon because that's something which is better handled by the init system.

### DIFF
--- a/src/executables/sysrepo-plugind.c
+++ b/src/executables/sysrepo-plugind.c
@@ -96,6 +96,9 @@ typedef struct sr_pd_ctx_s {
     ev_timer init_retry_timer;     /**< Initialization retry timer. */
 } sr_pd_ctx_t;
 
+/** @brief Options for sr-plugind's call to sr_connect */
+static sr_conn_options_t connect_options = SR_CONN_DAEMON_REQUIRED;
+
 /**
  * @brief Callback called by the event loop watcher when a signal is caught.
  */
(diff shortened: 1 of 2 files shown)


src/executables/sysrepo-plugind.c
```diff
@@ -314,7 +314,7 @@ sr_pd_session_check(sr_pd_ctx_t *ctx)
         ctx->connection = NULL;
 
         /* reconnect */
-        rc = sr_connect("sysrepo-plugind", SR_CONN_DAEMON_REQUIRED | SR_CONN_DAEMON_START, &ctx->connection);
```
michalvasko: It would probably be best if there were both options available. So, please add a flag for sysrepo-plugind which would add the flag SR_CONN_DAEMON_START to this call. Meaning, the default will be without it - what you proposed.

jktjkt: I can do that, but I wonder what is your reason for asking for this. Do you have a particular use case in mind where an auto-start is actually beneficial?

michalvasko: Not sure if when normally deployed, it is better then to have it all properly configured like you said. But when trying out some plugins or simply for testing purposes it may find its use.

jktjkt: I've modified the code according to your request, but please note that `sysrepod` is still auto-started only if `sysrepo-plugind` runs as root. That's why I doubt someone is going to use this feature for testing.

If this persuades you, just pick the older commit, otherwise please feel free to go with the new version f040b14003c89124b95739e6b1bd441afc179148 which adds the `-D` option to follow the old behavior.

PR #951, opened 2017-10-06
Identref fix

### LINKED ISSUE #950: identityref cannot be resolved (when referenced indirectly)
Hi,
please have a look at this commit:
https://github.com/frimpler/sysrepo/commit/f97d2f0043896188c58036661d96772f35e2fb44

It contains a test case which causes a failure in identityref resolution

Problem description:
--------------------------
The module "set_items-mod1" has a leaf which refers to a certain interface instance using the type if:interface-ref of ietf-interfaces. The test case creates an interface having the type iana-if-type:ethernetCsmacd and tries further to add a list instance of "/set_items-mod1:cont/list". This does not work, because the type iana-if-type:ethernetCsmacd cannot be found in resolve_identref().

Reasons:
------------
There are two error scenarios:
1.)
If the module "set_items-mod1" will be used as visible in the commit, the module iana-if-type is not in the module list of ly_ctx, because it has not been imported explicitly by "set_items-mod1", but resolve_identref() iterates through this list.
2.) 
If the module "set_items-mod1" imports iana-if-type (which is IMHO not necessary), the resolution fails anyway - now with the module in the list - because it's not flagged as "implemented".

Btw.: because the resolution fails, we have to deal with crashes when the ident-pointer of lyd_val will be accessed (e.g. in lyv_data_content()). Unfortunately I couldn't catch the crash with my test case, but the root cause of it is addressed, anyway.

Regards,
Frank

michalvasko:
### Description
Fix of a problem when foreign identity is used in an identityref that is being duplicated into another context, use libyang data callback.

### Closure
Fixes #950

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -1834,6 +1834,10 @@ dm_append_data_tree(dm_ctx_t *dm_ctx, dm_session_t *session, dm_data_info_t *dat
 
         if (NULL == data_info->node) {
             data_info->node = sr_dup_datatree_to_ctx(di->node, data_info->schema->ly_ctx);
+            if (NULL == data_info->node) {
+                SR_LOG_ERR("Failed to duplicate %s data tree into another context", di->schema->module->name);
+                return SR_ERR_INTERNAL;
+            }
         } else {
             ret = lyd_merge_to_ctx(&data_info->node, di->node, LYD_OPT_EXPLICIT, data_info->schema->ly_ctx);
             CHECK_ZERO_LOG_RETURN(ret, SR_ERR_INTERNAL, "Failed to merge %s data tree", di->schema->module->name);
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -1830,6 +1830,8 @@ dm_append_data_tree(dm_ctx_t *dm_ctx, dm_session_t *session, dm_data_info_t *dat
 
     /* transform data from one ctx to another */
     if (NULL != di->node) {
+        ly_ctx_set_module_data_clb(data_info->schema->ly_ctx, dm_module_clb, dm_ctx);
+
         if (NULL == data_info->node) {
             data_info->node = sr_dup_datatree_to_ctx(di->node, data_info->schema->ly_ctx);
         } else {
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -4893,14 +4893,10 @@ dm_install_module(dm_ctx_t *dm_ctx, dm_session_t *session, const char *module_na
         ll_node = module->deps->first;
         while (ll_node) {
             dep = (md_dep_t *)ll_node->data;
-            if (dep->type == MD_DEP_EXTENSION || dep->type == MD_DEP_DATA) {
-                /* Note: imports are automatically loaded by libyang */
-                rc = dm_load_schema_file(dm_ctx, dep->dest->filepath, true, &si);
-                CHECK_RC_LOG_GOTO(rc, unlock, "Loading of %s was not successfull", dep->dest->name);
-            }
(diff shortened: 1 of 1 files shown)


frimpler:
Hi Michal,

to me it seems, that the callback will be used in different places in order to overcome problems with missing modules in ly_ctx instances. I'm afraid that this has to be done again and again, whenever a new, similar problem occurs. I'm mentioning it, because I'm currently working at a different error and it's again a missing module which breaks a correct augmentation of a module. (This time I could put it in moveto_resolve_model(), for example)
My question: 
Isn't it better to initialize ly_ctx instances e.g. in ly_ctx_new()? Of course, one has to think of a way to make use of a callback from the library "consumer" like dm_module_clb() and a way to update the context, when new modules have been loaded...
I'm asking this mainly, because I want to understand, how you preferred to solve such kind of problems at best.

Kind regards,
Frank

michalvasko:
Hi Frank,
it depends which solution to use. For *libyang*, the idea generally is that when you are parsing data, there are situations when you may require modules you could not possibly know you require before parsing itself. Then the data callback gets called and gives a chance to load the missing module. As for *sysrepo*, it uses this mechanism whenever the situation is such that it was not possible to know beforehand.

Such a case is this `identityref` because there can be any module that defines identities that were derived from `ietf-interfaces` base interface identity, which then gets used as a value in data and *sysrepo* does not know beforehand (it tries to load into a context only models really needed, so, for example, loading all models that extend all identities in models already in the context is not preferred).

So, best would you create a new issue and we will see what exactly is the problem and how to solve it.

Regards,
Michal

frimpler:
Hi Michal,

thanks for your answer. It reflects somehow, what I have seen/experienced already in the code. Apart from the the identityref case above, would you say that all modules, which are "visible" by an include/import hierarchy are expected to be loaded before parsing the tree itself? Or if not, why not. Performance? (Btw. I'm asking exactly this, because this is the problem in the issue for which I'm currently preparing a test case)
Concerning writing of issues. My employer wants me to support you, because we are still facing blocking issues which needs to be "formulated" as a regression test case. And as soon as a test exist I will of course first try to propose a solution by myself. I may become each time better in it..:-)
In this context: 
I can imagine that I will create always a pull request with a test case which triggers a misbehaviour. But due to the fact, that a pull request has a CI pipeline attached, I'm afraid that such pull requests have trouble to be accepted. So I'm looking for a better way to deliver test cases. Any suggestions?

Regards,
Frank

michalvasko:
Hi Frank,
this issue is fairly complicated because *sysrepo* tried from the start to be as effective as possible and not load redundant models (so I am often encountering opposite problems, required models are not being loaded). For this reason each model has stored its dependencies (`repository/data/internal/sysrepo-module-dependencies.xml`) which are used to load all the required models (but this list may not be complete like I said before, sometimes you simply cannot know). To sum up, all imports/includes of a model are loaded together with additional dependencies, which are also model augmenting this model and with the last commit I added recursive loading of dependencies in case there is a data dependency on another model (which means it needs another model data for validation - foreign leafrefs/instance-identifiers and maybe some other cases which I don't remember).

As for your support, it is very welcome because we simply do not have enough capacities to provide proper support and further development of all the software even though we are doing what we can. I think you can provide test cases using pull requests as there is likely no better way. It is true, the travis tests will initially fail, but I can then create another pull request fixing the problem and after it gets merged, we can restart the travis build for your pull request, it will pass and can be merged, too. What do you think? Also, it may be better to discuss this privately, it is not a sysrepo issue and it definitely does not concern this pull request anymore.

Regards,
Michal

frimpler:
ok and thanks!

PR #959, opened 2017-10-19
Do not consider operations in modules as data

michalvasko:
### Description
Changed the meaning of has_data flag, operations are no longer considered as such.


michalvasko:
Unfortunately, operations subscriptions are treated exactly as data subscriptions so changes would have to be more extensive than what is in this pull request. Since it fixes merely an inaccuracy or rather an inconsistency, leave it be for now.

PR #967, opened 2017-10-29
docker: add end to end testing

mislavn:
This is a demonstration on how to do end to end testing on a Ubuntu 16:04 and OpenWrt docker image.

The first run of the bash scripts will be slower because the base docker images are being build and for the OpenWrt image it will take some time, also the image is 10Gb big.

After that every rerun of the script will fetch the latest changes from the master branch, discard the previous docker cache and run the tests. This can be changed so the script run's only if there are new commits in the master branch.

For the NETCONF tests go-netconf is used and simple toml configuration file which specifies: url, username, password, XML request and XML response.


michalvasko:
Hi,
if someone could also test the OpenWRT image it would be great, but I guess it can be merged anyway.

Regards,
Michal

PR #987, opened 2017-11-15
sysrepoctl: Fix memory leak

jktjkt:
Commit d119b61a introduced a memory leak by forgetting to free the individual members of that array.

Discovered by ASAN; a simply `make install` was failing.

### DIFF
--- a/src/executables/sysrepoctl.c
+++ b/src/executables/sysrepoctl.c
@@ -1047,7 +1047,10 @@ cleanup:
     md_destroy(md_ctx);
     ly_ctx_destroy(ly_ctx, NULL);
     if (local_search_dir) {
-        free((char*)search_dirs);
+        for (int i = 0; i < search_dir_count; ++i) {
+            free(search_dirs[i]);
+        }
+        free(search_dirs);
     }
     return rc;
 }


src/executables/sysrepoctl.c
```diff
@@ -1047,7 +1047,10 @@ srctl_install(const char *yang, const char *yin, const char *owner, const char *
     md_destroy(md_ctx);
     ly_ctx_destroy(ly_ctx, NULL);
     if (local_search_dir) {
-        free((char*)search_dirs);
+        for (int i = 0; i < search_dir_count; ++i) {
```
michalvasko: We know there is only one search directory when `local_search_dir` is true, so the loop can be replaced with `free(search_dirs[0])`.

michalvasko:
Hi,
you are right, forgotten to free the directory itself, thanks. Please, address my request and then it can be merged.

Regards,
Michal

src/executables/sysrepoctl.c
jktjkt: I'm not a C programmer, I usually work in C++, so please take this with a grain of salt. I do not know C idioms.

That said, I think that this function is "complex enough" already; it has 150+ LOC. The allocation/deallocation is only performed if the caller did not provide an appropriate value. That is something that I find strange. There's just one caller for this function, so my recommendation is to move this alloc/dealloc thing outside to keep this one simpler. As an anecdote -- as I started work on fixing this leak, I originally thought that it should always deallocate each item, so I begun changing all other places to use `strdup`, similarly to what `srctl_get_dir_path` is already doing.

You're right, the size is always `1`. However, given that this is established on line 914 and the deallocation happens on line 1050, I think that this might change in future and nobody will notice. It's definitely not a hot path, so why not keep the `for` cycle around?

michalvasko: Firstly, yes, I agree that such conditional allocation should not happen and is wrong in principle. However, there are countless such things in *sysrepo* code and I no longer bother, I simply cannot fix them all, the project is too big for that. Getting that out of the way, when I look at this code some time later I would definitely spent a minute or two on figuring out that the loop could have been omitted as the count is always `1`. In other words, the freeing is already conditional as is the allocation, so why make it generic when the allocation is specific? It increases the confusion for me, that is all.

But, I do not know C idioms either, I am simply advocating principles that have worked for me. I also understand your point but consider my argument stronger. That being said, we are being too abstract for such small change. You have created the pull request, the code works, so it can be merged.

PR #989, opened 2017-11-16
swig/cpp: removed "using namespace std" from header files

Mandar12:


mislavn:
I'll need to do some work in SWIG to fix this. I'll make a new pull request today or tomorrow.

PR #992, opened 2017-11-16
tests: Unify and relax Unix group regex

jktjkt:
My CI was failing on this one because it was running as a user/group named `turbo-hipster`. This patch unifies all places where the username and gropu name are validated (except that sometimes the username is hardcoded). I'm also adding support for dashes and underscores.


michalvasko:
Hi,
right, thanks.

Regards,
Michal

PR #994, opened 2017-11-21
Do not use parent pointer directly in schema nodes

### LINKED ISSUE #993: Question about the function rp_dt_has_parent_list
Hi,
there is a question about `lys_parent` usage in `rp_dt_has_parent_list`.
I think below code should use lys_parent to get n's parent, not directly use pointer.
`struct lys_node *n = node->parent;`

if the n's parent is an augement node.  the following code `n = lys_parent(n);` will get NULL. 
but actually we want get the augement's target node as its parent, not NULL.

in libyang, the API `lys_parent` was meant to be used when one does not want to work with augments at all. 

I also open an issue ( https://github.com/CESNET/libyang/pull/397 ) in libyang.

could you give your comments?  modify sysrepo or libyang?

michalvasko:
### Description
Instead using parent pointer in schema nodes, always use lys_parent().

### Closure
Fixes #993

### DIFF
--- a/src/rp_dt_get.c
+++ b/src/rp_dt_get.c
@@ -876,7 +876,7 @@ static bool
 rp_dt_has_parent_list(struct lys_node *node, struct lys_node **found_list, size_t *depth)
 {
     if (NULL != node) {
-        struct lys_node *n = node->parent;
+        struct lys_node *n = lys_parent(node);
         size_t dep = 0;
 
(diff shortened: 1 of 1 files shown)


ustcpxy:
the rule should be applied in `rp_dt_no_parent_list_until` as well.

michalvasko:
Fair enough.

PR #1001, opened 2017-11-30
sr_realloc added and corresponding values and trees functions

### LINKED ISSUE #965: new function sr_realloc_values
One of the problems I encountered with the callback's of type:
`typedef int (*sr_dp_get_items_cb)(const char *xpath, sr_val_t **values, size_t *values_cnt, void *private_ctx);`

Is when I need to return a list and I don't know how big will it be, so I have to create a custom linked list, populate it with data and pass that data to `sr_val_t **` at the end of the callback.

Is it possible to implement a new function which can change the size of the array, example:

```
// start callback
size_t value_cnt = 2;
sr_val_t *values = NULL;
sr_new_values(value_cnt, &values);

/* working */

value_cnt = 4;
sr_realloc_values(value_cnt, &values);

/* working */

// end callback
```

This might be trick because of the custom sysrepo memory management, I'm not sure.

michalvasko:
### Description
sr_realloc function added to enable enlarging arrays of sysrepo trees and values.

### Test case
sr_realloc (also sr_realloc_trees, sr_realloc_values) tests added.

### Closure
Fixes #965

### DIFF
--- a/inc/sysrepo/trees.h
+++ b/inc/sysrepo/trees.h
@@ -66,6 +66,15 @@ int sr_new_tree(const char *root_name, const char *root_module_name, sr_node_t *
 int sr_new_trees(size_t tree_cnt, sr_node_t **trees);
 
 /**
+ * @brief Reallocate an array of sysrepo trees (uninitialized tree roots).
+ *
+ * @param [in] old_tree_cnt Current length of the tree array.
+ * @param [in] new_tree_cnt Desired length of the tree array.
+ * @param [in,out] trees Returned newly allocated/enlarged array of trees.
+ */
+int sr_realloc_trees(size_t old_tree_cnt, size_t new_tree_cnt, sr_node_t **trees);
+
+/**
  * @brief Set/change name of a Sysrepo node.
  *
  * @param [in] node Sysrepo node to change the name of.
--- a/inc/sysrepo/values.h
+++ b/inc/sysrepo/values.h
@@ -66,6 +66,15 @@ int sr_new_val(const char *xpath, sr_val_t **value);
 int sr_new_values(size_t value_cnt, sr_val_t **values);
 
 /**
+ * @brief Reallocate an array of sysrepo values.
+ *
+ * @param [in] old_value_cnt Current length of the value array.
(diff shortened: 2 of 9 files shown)


mislavn:
Thanks for the feature.

mislavn:
I got a segmentation fault, am I using `sr_realloc_values` wrong?

```
cat example.c 
#include "sysrepo.h"
#include "sysrepo/values.h"
#include "sysrepo/xpath.h"

int
main(int argc, char **argv)
{
    sr_val_t *v = NULL;
    int rc = SR_ERR_OK;

    rc = sr_new_values(1, &v);
    if (SR_ERR_OK != rc) {
        return rc;
    }

    sr_val_set_xpath(&v[0], "/ietf-interfaces:interfaces-state/interface[name='eth0']/type");
    sr_val_set_str_data(&v[0], SR_IDENTITYREF_T, "ethernetCsmacd");

    printf("call sr_realloc_values\n");
    rc = sr_realloc_values(1, 2, &v);
    if (SR_ERR_OK != rc) {
        return rc;
    }
    printf("called sr_realloc_values\n");

    sr_val_set_xpath(&v[1], "/ietf-interfaces:interfaces-state/interface[name='eth0']/oper-status");
    sr_val_set_str_data(&v[1], SR_ENUM_T, "down");

    printf("success\n");

    return 0;
}
```
```
gcc example.c -o example -lsysrepo
./example 
call sr_realloc_values
Segmentation fault (core dumped)

```

michalvasko:
Hi,
there were some minor issues, they should be fixed now and it should not crash any more (like I said, I am going to write some tests for values, there was the problem). However, like I said in the issue before, in your use-case the `realloc` performs exactly the same as `malloc` would because you interleave allocations (first `sr_new_values`, then `sr_val_set_xpath`, `sr_val_set_str_data`, and then `sr_realloc_values`). The xpath and string value get allocated right after the values so when you realloc, a new memory block must be taken leaving the previous one the way it was (sysrepo does not support freeing). So, if you first get to the final number of values and then start filling them, no memory should be wasted (provided that some conditions are met). I understand that may not be the most natural way of working with values, but sysrepo memory management is limited in this way.

Regards,
Michal

PR #1002, opened 2017-11-30
BUGFIX remove branch devel-cli from dockerfiles

mislavn:
### Description
The dockerfiles for devel branch do not build any more because there is no longer a `devel-cli` branch in Netopeer2.

### Test case
`docker build -t sysrepo/sysrepo-netopeer2:devel -f deploy/docker/sysrepo-netopeer2/Dockerfile.devel .`

The command will fail without the patch.


michalvasko:
Hi,
OSX tests are currently failing because of a permission problem probably introduced by some travis update. I would try waiting some time (we will probably not release sysrepo today, so till Monday at least) and if it is still not fixed, we can modify the travis file.

Regards,
Michal

PR #1035, opened 2018-01-18
docker: update openwrt build

mislavn:
After the LEDE/OpenWrt merge the script `build_and_run_on_openwrt.sh` in end-to-end tests fails, this patch fixes it and changes the configuration so it holds only the diff.


mislavn:
yes it works :)

Issue #1032, opened 2018-01-16
Infinite loop in datamanager due to inter dependencies


frimpler:
Hi,

please have a look at this commit https://github.com/frimpler/sysrepo/commit/c276c4e76420cfd36c6a32ea9c270dcb5f334fe9
It contains a test case + two yang modules which trigger an (infinite) recursion in datamanager during dependency resolution. 

Description:
---------------
The two yang modules define each leaf-refs pointing to the "opposite" module, which produces an inter dependency (->sysrepo-module-dependencies.xml). The test case just attempts a sr_get_item(), which now triggers the datamanager to resolve the dependency (get_item is in the test case not really of interest, it's just a means to start the resolution process). During this process the datamanager implementation does not recognize, that the dependend module has already been validated and starts again and again... 

Btw. when I added the module in validation prior validation to dm_session_ctx->session_modules (datamanager.c:2677), the recursion could be stopped/broken, but this solution caused problems in other test cases and therefore, I'd prefer that you do the correction :-). 

Rather a side problem, but important for the test case is, that we didn't find a means to uninstall those two modules, because they reference each other and refuse to get uninstalled. Would be nice, if you could point us here to a solution. (The uninstallation is out-commented in the test case)

Best regards,
Frank

michalvasko:
Hi Frank,
it took me some time to find a proper (hopefully) and simple fix, but it should work now with the pull request. Regarding uninstallation, that is a known issue not considered serious enough to fix for now. If I have time I will take a look at how to solve it and will reference this issue. Manually, you can uninstall any modules by simply deleting their subtrees in the sysrepo internal dependency XML tree `repository/data/internal/sysrepo-module-dependencies.xml` and their references in other modules in the file, in case you really needed it.

Regards,
Michal

frimpler:
Great! I will wait until your pull request has been merged and then I will create one for the test case.

Regards,
Frank

PR #1039, opened 2018-01-24
test case to verify solution of sysrepo issue #1032

frimpler:
Hi,

as discussed in the mentioned issue, here's the test case to verify the solution. For the description of the test, please refer to https://github.com/sysrepo/sysrepo/issues/1032

Btw.: 
In order to get rid of the installed modules afterwards, I had to add an ugly workaround. I hope, you can live with it, as long as no solution for the uninstallation of interlocked modules is in sight.

Kind Regards,
Frank Rimpler


michalvasko:
Hi Frank,
actually, it is already possible to uninstall several modules at once using `sysrepocfg`. I have referenced the commit in #1032 but not mentioned it explicitly, sorry about that.

Regards,
Michal

frimpler:
Hi Michal,

yes, I have already noticed it and tried to apply it, before I created the pull request, but it didn't work (in either un-installation order). Anyway, because my expectation was, that interlocked modules cannot be easily un-installed, I didn't undertake more investigations. Do you suggest any change in the test?

Regards,
Frank

michalvasko:
Hi Frank,
just uninstall the modules with this command
```C
exec_shell_command("../src/sysrepoctl --uninstall --module mutual-leafref-base,mutual-leafref-augment", ".*", true, 0);
```
It worked fine for me. However, I have tried uninstalling first `mutual-leafref-augment` and that was not handled properly so I have created a pull request which should make the order arbitrary.

Regards,
Michal

frimpler:
Hi Michal,

I'm lucky that un-installation works now. So please pull it in :-)

Regards,
Frank

PR #1043, opened 2018-01-31
test case + possible solution for feature enabling issue

frimpler:
Hi,

the pull request addresses an issue with feature enabling. Here's the description of the test case, which reflects the problem:

> In the YANG modules "feature-dependencies1/2/3" several features are defined.
   The feature "abc" defined in "feature-dependencies2" depends on feature "xyz"
   defined in "feature-dependency3". Both features can be enabled without problems.
   But if afterwards one tries to enable other features without dependencies
   e.g. "issue1" in "feature-dependencies1" or "issue2" in "feature-dependencies2"
   an error occurs: Feature "abc" is disabled by its 1. if-feature condition.
   The reason is the order in which the persistent data are loaded. The persistent
   data of imported modules should always be loaded before the data of the importer.

We have additionally to the test case made an attempt to fix it in datamanager.c. Please have a look at it.

Btw.: as it seems, there's a little inconsistency with libyang and sysrepo's devel branch. I had to adapt lys_print_mem()'s parameters just to get it compiled. Please check this also for correctness.

Regards,
Frank

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -951,13 +951,7 @@ dm_load_module(dm_ctx_t *dm_ctx, const char *module_name, const char *revision, 
     rc = dm_init_missing_node_priv_data(si);
     CHECK_RC_LOG_GOTO(rc, cleanup, "Failed to initialize private data for module %s", module->name);
 
-    /* apply persist data enable features, running datastore */
-    if (module->has_persist) {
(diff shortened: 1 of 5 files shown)


michalvasko:
Hi Frank,
thanks, seems good! Sorry about the missing parameters, I have forgotten to update sysrepo, so thanks again.

Regards,
Michal

PR #1044, opened 2018-02-06
bugfix: support "config false" node into choice node

ustcpxy:
the pull request can slove two issues:
case 1: if there is a choice node with config false leaf node, the leaf node can not be installed into op-data-subtrees in sysrepo-module-dependencies.xml .
case 2: if the state data is nested in choice/case node, the depth should not be increased.


michalvasko:
Hi,
seems fine, these nodes were not handled. Thanks.

Regards,
Michal

PR #1048, opened 2018-02-09
replace cp by install

frimpler:
Hi,

I made a little change in sysrepoctl: 
When installing modules with sysrepoctl the schema file will be copied to /etc/sysrepo/yang/ by "cp". "cp" copies also the permissions and if the source file has write protection, then a "reinstallation" will fail. Such a "reinstallation" happens for example, if one first installs ietf-interfaces and afterwards a module my-interfaces which imports ietf-interfaces.

I replaced "cp" by "install" with "user read+write permission" and all other with "read permission".

I hope it's ok in that way.

Regards,
Frank

Btw.: 
Do you plan to merge devel into master branch for sysrepo soon? It's already 2 month ahead  and I'd really like to see our recent changes at the master branch :-).

### DIFF
--- a/src/executables/sysrepoctl.c
+++ b/src/executables/sysrepoctl.c
@@ -758,7 +758,7 @@ srctl_schema_install(const struct lys_module *module, const char *yang_src, cons
                 printf("Schema of the module %s is already installed, skipping...\n", module->name);
             } else {
                 printf("Installing the YANG file to '%s'...\n", yang_dst);
-                snprintf(cmd, PATH_MAX, "cp %s %s", yang_src, yang_dst);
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi Frank,
makes sense and will be fine.

Regarding the merge, we are planning to talk about it on the next meeting (Wednesday) and my stance actually was that there are no critical changes and so the merge can wait. But if you want it merged I have no problem with it and will mention it.

Regards,
Michal

frimpler:
Yes, I'd really appreciate the merge as soon as possible and next week would be good.

Btw.: From my point of view there are corrections for critical issues, e.g.:  the infinite recursion mentioned in issue 1032 causes really trouble.

PR #1058, opened 2018-02-21
Cpp event notif

### LINKED ISSUE #1057: Support for sr_event_notif_send for SWIG python
I see that the interface file does not have sr_event_notif_send support. How to send notifications using SWIG/Python2?

mislavn:
### Description
Implemented the function sr_event_notif_send and sr_event_notif_send_tree in the language bindings. Also mada small bug fixes.

### Closure
closes  #1057

### DIFF
--- a/swig/cpp/src/Session.cpp
+++ b/swig/cpp/src/Session.cpp
@@ -37,7 +37,7 @@ extern "C" {
 }
 
 Session::Session(S_Connection conn, sr_datastore_t datastore, const sr_sess_options_t opts, \
-		 const char *user_name)
+                 const char *user_name)
 {
     int ret;
     _opts = opts;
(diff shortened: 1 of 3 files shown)


mislavn:
I'm not sure what should I do if the sr_event_notif_send function returns "SR_ERR_NOT_FOUND", at the moment an exception is thrown.

michalvasko:
Why not? That is how errors are communicated in C++ and I do not think it is a valid situation, it is an error.

Regards,
Michal

mislavn:
ok.

jktjkt:
FYI, this looks like duplicating 9e31589366041ec188186762c08bda0ad880f09c. I'm not really attached to the API names I used, so feel free to clean this up by removing any of these :).

mislavn:
Sorry, I haven't noticed them before.
I'll make a new pull request and remove send_event API to make it more consistent, thanks for the reminder.

PR #1060, opened 2018-02-23
fix bug: support list key define in use statement for action/rpc

qmm161:
### Description
sr_is_key_node does not work well if list key defined in use statement. 

yang module for demo:
---
module Demo {
    namespace "http://example.net/demo";
    prefix DM;
    yang-version "1.1";
    
    grouping KeyGroup {
    	leaf id{
    		type uint32;
    	}
        
    	leaf f1{
    		type uint32;
    	}        
    }

    container Root {
        action setdata {
            input {
                list data {
                    key id;
		    uses KeyGroup;
                }
            }
        }
    }
}


michalvasko:
Hi,
the fix is correct, merged.

Regards,
Michal

PR #1083, opened 2018-03-17
added libSysrepo-cpp.pc.in for pkg-config

andrei-pavel:
### Description
There is no `.pc` file to help include header files and link with the `libSysrepo.so` library.
I suggest `git diff --color-words` for an easy review.

### Test case
-

### Closure
-


michalvasko:
Hi,
seems fine, why not.

Regards,
Michal

Issue #1085, opened 2018-03-20
 Enabling of dependent features does not work (3)


frimpler:
Hi,

with the testcase cl_feature_dependencies_2_test() in this commit https://github.com/frimpler/sysrepo/commit/58579a3086ba5f9ff1cf3e60e401630ecf622aa0 an issue concerning the enabling of dependent features is triggered.

### Description
When a temporary libyang context is created in data_manger.c for a module which has already been parsed  and verified before and the features shall be set according to the original module using the function dm_enable_features_in_tmp_module() the error "Feature XXX is disabled by its 1. if-feature condition" occurs for dependent features because the imported modules are not considered in temporary contexts.

With this new testcase the REQUEST_TIMEOUT has to be increased to 4 seconds. Otherwise the old testcases cl_locking_test() and cl_ds_locking_test() get a timeout during the valgrind run when calling sr_lock_datastore().

### Proposed Solution
The proposed solution is to clone the features of the original module to the temporary one (instead of enabling / disabling them) in data_manager.c (https://github.com/frimpler/sysrepo/commit/d5e9ff7fb3a65ca159e2b89dfa575d9baaee34dc) by calling the new libyang function lys_features_clone() (https://github.com/frimpler/libyang/commit/b609a00cc5208ccd92cd7eea0cb7eb2a8f6d518b). The new function avoids the feature dependency checks which have been done already for the original module.

Regards,
Peter + Frank

michalvasko:
Hi Peter & Frank,
the solution seems reasonable and effective, so once you create a pull request, I think it can be merged. However, please put `lys_features_clone` into `src/common/sr_utils.c` and rename it to `sr_features_clone` or something similar. Generally speaking, it is not a good idea to clone features like this, only if you know what you are doing (this case). Thanks.

Regards,
Michal

PR #1092, opened 2018-03-28
Feature dependencies (3) solution

frimpler:
Hi,

this pull request is related to issue https://github.com/sysrepo/sysrepo/issues/1085 and contains the TC and the proposed solution.

Regards,
Peter


michalvasko:
Hi Peter,
you changed it like I asked you to, so it can be merged now. However, I do not particularly like the `usleep()` in the test. Still, similar practices are already in the tests so it's okay, I guess.

Regards,
Michal

PR #1099, opened 2018-04-17
make sysrepo connection timeout switchable

frimpler:
Hi,

please consider this little modification which makes it possible to disable the connection timeout in the sysrepo client library by compiler switch.

Background: 
In low performance environments, sysrepo clients might be too busy to answer in time, if the connection then times out it starts that things get worse. For this reason we want to be able to switch it off (and deal with the consequences, if necessary)

Regards,
Frank


rkrejci:
I'm fine with the idea, but could you integrate setting the variable into cmake processing? Ideally, I would connect it it somehow with `REQUEST_TIMEOUT` using some special value (zero or negative value).

frimpler:
Please check the update.

Thanks,
Frank

PR #1101, opened 2018-04-24
Removed separate long request timeout

michalvasko:
### Description
Little use to have 2 separate timeouts so one is used now for all requests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,27 +176,8 @@ set(STORE_CONFIG_CHANGE_NOTIF 1 CACHE BOOL
     "Save config-change notifications (RFC 6470) in the notification store (slows down the commit process).")
 
 # timeouts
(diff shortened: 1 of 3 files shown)


michalvasko:
@frimpler Are you fine with this change?

frimpler:
yeah, I definetely prefer simplifications. I assume, you know/remember the reasons for the two timeout values :-). In fact, we have in this way always 15 seconds instead of 4 as default.

michalvasko:
Was there any reason 2 timeouts were used? The worst thing that can now happen is that some operations could timeout after 15 seconds instead of 4 (by default) and they should timeout only if there is something seriously wrong.

frimpler:
We didn't introduce the timeouts (only kept them as they were). Or request was just to introduce the means to disable them (by compiler switch).

michalvasko:
By mistake this pull request now also includes fix for issue #1107, namespace duplication check was added to `sysrepoctl` install command.

Issue #1111, opened 2018-05-04
Cannot configure identityrefs with derived identity in separate model under certain circumstances


alangefe:
### Description
The following model graph is used to exhibit the problem:

```
      +-------------+
      |             |
      | id-ref-base |          +------------------+
      |             |          |                  |
      +----+-----+--+          | id-def-extension |
           ^     ^             |                  |
           .     | Augment     +------+--------+--+
           .     |                    |        .
           .     |                    |        .
    Extend .     |                    |        .
           .     |            Import  |        . Extend
           .     |                    |        .
           .     |                    v        v
       +---+-----+--+   Import      +-+--------+--+
       |            +-------------->+             |
       | id-ref-aug |               | id-def-base |
       |            +<..............+             |
       +------------+    Extend     +-------------+
```

If the model installation order is:

 1. `id-ref-base`
 2. `id-ref-aug`
 3. `id-def-base`
 4. `id-def-extension`

then a dependency will be left out of the graph and the user will be unable to configure the identity defined in `id-def-extension`.

The source of the problem is that a consumer of an identity (i.e., a module with an identityref) only adds the source of that identity module as an extension to itself when that source is marked as 'implemented'. However, this causes that dependency to be lost, as it won't be regenerated when the source of the identity is later installed with sysrepoctl.

We've written a unit test that demonstrates the problem. See `set_item_id_ref()` in https://github.com/alangefe/sysrepo/commit/5577960d27dfb18d19c28c20fb64711bb76cb57c.

### Proposed solution

We've implemented a proposed solution in the same commit referenced above. Our solution is to allow identityrefs to be added as extend dependencies even if the source module hasn't yet been implemented. The identity can then be used once the source of the identity is installed.

An alternative implementation to this could be to regenerate the identityref module when the identity source module is installed. However this seems like it would introduce quite a bit more complexity into the installation process.

michalvasko:
Hi,
thanks for the tests but I do not think we can merge your implemented solution. If the dependency is added, the model with the identity is loaded into *libyang* context of the other models as implemented meaning the identity can be used in data. That is contrary to YANG specification.

So, we will have to go with your other proposed solution even though it adds some complexity as the whole schema tree must be traversed again. Will you please create a pull request only with the new tests? Meanwhile, I will fix the problem. Thanks.

Regards,
Michal

michalvasko:
[...]

PR #1112, opened 2018-05-04
Add idrefs as ext deps even when the src mod is not implemented yet

alangefe:
### Description
See https://github.com/sysrepo/sysrepo/issues/1111 for a detailed description of the problem and the proposed solution.

### Test case
New unit test `set_item_id_ref()` in `rp_dt_edit_test` demonstrates the problem and tests the solution.

A minor change to `md_test` was also made to reflect the new expected behavior.


michalvasko:
Solution is unsuitable as explained in #1111.

Issue #1113, opened 2018-05-04
Error when change subscriber requests list of changes after receiving commit notification


alangefe:
### Description
When a subscriber who has been sent a commit notification requests the list of changes, sometimes the subscriber will get an error indicating that the commit context does not exist.

The problem is that sysrepo sends out the commit notifications prior to saving the commit context, meaning there is a window of time during which the subscriber can ask sysrepo for the changes before they have been saved.

The diagram below depicts the erroneous behavior:

```
    Errorneous case:

+----------+                  +-----------------+
| sysrepod |                  |change subscriber|
+-----+----+                  +--------+--------+
      |                                |
      |       Notify about changes     |
      +-------------------------------->
      |                                |
      |   Request list of changes      |
      <--------------------------------+
      |                                |
      |    Error: No commit context    |
      +-------------------------------->
      |                                |
      |                                |
      +--+                             |
      |  | Save commit context         |
      |  |                             |
      <--+                             |
      |                                |
      v                                v
```

### Proposed solution
We've implemented a proposed solution in https://github.com/alangefe/sysrepo/commit/06cd6c0d4f66c40a1a354278cb872bbea35684aa. This commit postpones sending the commit notifications until after the commit context has been saved. In other words:

```
+----------+                  +-----------------+
| sysrepod |                  |change subscriber|
+----+-----+                  +-------+---------+
     |                                |
     +--+                             |
     |  | Save commit context         |
     |  |                             |
     <--+                             |
     |                                |
     |                                |
     |       Notify about changes     |
     +-------------------------------->
     |                                |
     |   Request list of changes      |
     <--------------------------------+
     |                                |
     |   Return changes from context  |
     +-------------------------------->
     |                                |
     v                                v
```

michalvasko:
Proposed solution merged, should not happen any more.

Regards,
Michal

alangefe:
Closed with #1114, included in 0.7.4.

PR #1114, opened 2018-05-04
Save the commit context before sending notifications

alangefe:
This pull request corresponds to #1113 , which describes the problem and proposed solution. In summary, this request moves saving the commit context to before sending out commit notifications. This eliminates the window of time during which notified subscribers can query for the changes corresponding to the notification and get an error because the changes have not yet been saved.

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -4733,7 +4733,38 @@ dm_commit_notify(dm_ctx_t *dm_ctx, dm_session_t *session, sr_notif_event_t ev, d
                 d_cnt++;
             }
         }
+    }
 
+    if (SR_EV_VERIFY == ev) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi Andrew,
it seems fine to me, thanks.

Regards,
Michal

rosy8:
Hi,

In one of the scenario I got error like 
No NP commit context for commit ID ..

Is it related to this scenario ??

Regard's
Rosy

michalvasko:
Hi Rosy,
I think this scenario could have happened to you (the error message is the same) but it also could have been caused by some other bug. If it continues to happen to you, please start another issue.

Regards,
Michal

rosy8:
Hi Michal,

Ok. 
I will start a new thread for this issue.

Regard's
Rosy

PR #1120, opened 2018-05-11
Fix segfault due to NULL child if module contains augment whose children are contingent on a feature

### LINKED ISSUE #1105: sysrepoctl segmentation fault while installing YANG module with augments
### First items attempted
* [x] Did you test it on the latest sysrepo devel branch and the latest libyang devel branch?

### Description
Installing a YANG module with sysrepoctl results in a segmentation fault if the module contains an augment whose children are contingent upon a feature. The same segmentation fault occurs if the module contains an augment with a uses clause where the corresponding grouping’s children are contingent on a feature. The following YANG modules exhibit the issue.

```YANG
module test-ds0 {
  namespace "urn:ietf:params:xml:ns:yang:ietf-interfaces:test-ds0";
  prefix ds0;

  import ietf-interfaces {
    prefix if;
  }

  feature t1 {
  }

  augment "/if:interfaces-state/if:interface" {
    container foo {
      if-feature t1;
      leaf ds0ChannelNumber {
        type int32 {
          range "1..2147483647";
        }
      }
    }
  }
}
```

```YANG
module test-ds1 {
  namespace "urn:ietf:params:xml:ns:yang:ietf-interfaces:test-ds1";
  prefix ds1;

  import ietf-interfaces {
    prefix if;
  }

  feature t1 {
  }

  grouping group {
    leaf item {
      if-feature t1;
      type string;
    }
  }

  augment "/if:interfaces-state/if:interface" {
    uses group;
  }
}
```

The install command and output can be seen below.
```console
root@host:/# sysrepoctl --install --yang=/path/to/yang/models/ds0.yang --owner=root:root --permissions=666
Installing a new module from file '/path/to/yang/models/ds0.yang'...
Installing the YANG file to '/path/to/yang/repo/repository/yang/test-ds0.yang'...
Resolving dependency: 'test-ds0' imports 'ietf-interfaces'...
Installing the YANG file to '/path/to/yang/repo/repository/yang/ietf-interfaces@2014-05-08.yang'...

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

**Expected behavior:**
The module should install successfully.

**Actual behavior:**
Installing the module caused a segmentation fault in sysrepoctl dereferencing a NULL pointer.

### Versions
Versions used to reproduce the bug:
Sysrepo version: sysrepod - sysrepo daemon, version 0.7.3
Sysrepoctl version: sysrepoctl - sysrepo control tool, version 0.7.3
libyang version: yanglint 0.15.92

### Additional information
This was introduced in the recent fix for issue [#1011](https://github.com/sysrepo/sysrepo/issues/1011 "find wrong op-data-subtree for augment module if the top augment node is config false"). The segmentation fault occurs in [module_dependencies.c#1747](https://github.com/sysrepo/sysrepo/blob/c4fc434f0640905a3b4e3378151d2c399fba945a/src/module_dependencies.c#L1747). The child pointer is NULL presumably because the feature has not yet been enabled. This only occurs if the augment target is a non-configuration node. If the augment target contains configuration data, [module_dependencies.c#1712](https://github.com/sysrepo/sysrepo/blob/c4fc434f0640905a3b4e3378151d2c399fba945a/src/module_dependencies.c#L1712) verifies the child pointer is not NULL and essentially skips the other checks and asserts.

alangefe:
### Description
Installing a YANG module with `sysrepoctl` results in a segmentation fault if the module contains an augment whose children are contingent upon a feature. The same segmentation fault occurs if the module contains an augment with a uses clause where the corresponding grouping’s children are contingent on a feature. Issue #1105 documents this issue. The included test case demonstrates the failure. To address this, `md_traverse_schema_tree()` has been modified to tolerate the return of a null child from the call to the function `lys_getnext()`.

### Test case
The included test case attempts to install a valid model that would previously fail without this fix. The required yang model file is included.

### Closure
closes #1105

### DIFF
--- a/src/module_dependencies.c
+++ b/src/module_dependencies.c
@@ -1744,12 +1744,16 @@ md_traverse_schema_tree(md_ctx_t *md_ctx, md_module_t *module, md_module_t *main
                         if (augment) {
                             if (node->nodetype == LYS_AUGMENT) {
                                 child = (struct lys_node *)lys_getnext(NULL, node, NULL, 0);
-                                assert(!(child->nodetype & (LYS_CONTAINER | LYS_LIST)) || (intptr_t)child->priv & PRIV_OP_SUBTREE);
-                                assert(child->nodetype & (LYS_CONTAINER | LYS_LIST) || child->flags & LYS_CONFIG_R);
+                                assert(child == NULL || !(child->nodetype & (LYS_CONTAINER | LYS_LIST)) || (intptr_t)child->priv & PRIV_OP_SUBTREE);
+                                assert(child == NULL || child->nodetype & (LYS_CONTAINER | LYS_LIST) || child->flags & LYS_CONFIG_R);
                             } else {
                                 child = node;
                             }
-                            rc = md_check_op_data_subtree(dest_module, child);
+                            if (child != NULL) {
+                                rc = md_check_op_data_subtree(dest_module, child);
+                            } else {
+                                rc = SR_ERR_OK;
+                            }
                         } else {
                             child = node;
                         }
--- a/tests/md_test.c
+++ b/tests/md_test.c
@@ -1672,6 +1672,36 @@ md_test_insert_module_2(void **state)
     md_destroy(md_ctx);
 }
 
+/*
+ * @brief Test md_insert_module_5().
+ */
+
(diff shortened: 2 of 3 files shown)


michalvasko:
Hi,
the fix seems fine, sorry for not replying sooner, I was away. Once the tests pass, I will merge it, thanks.

Regards,
Michal

PR #1123, opened 2018-05-15
Unit test to verify issue #1111

alangefe:
Hello Michal,

Per your comment in issue #1111, here is a pull request which contains only the unit test changes necessary to validate the issue.

The tests in this pull request will fail, of course, until a suitable solution is found for issue #1111.

(This does not include our #1112 modifications to `md_test`, which assumed the #1112 changes to `md_collect_identity_dependencies()`'s behavior.)

Regards,
Andrew


michalvasko:
Hi Andrew,
thanks, tests pass now.

Regards,
Michal

Issue #1116, opened 2018-05-08
sysrepoctl error while installing module with a submodule containing a leafref


aweast:
### Things you may try first
(put "x" in "[]" if you already tried following)
* [x] Did you check if this is a duplicate issue?
* [x] Did you test it on the latest sysrepo devel branch?

### Description
Installing a yang model with submodules can fail if one of the submodules contains a LeafRef referring to an external module that is not yet installed. The ordering of include statements can cause the installation to fail or succeed.

In the attached modules, module `mws-main-module` includes two submodules `mws-main-submodule1` and `mws-main-submodule2`. Both `mws-main-submodule1` and `mws-main-submodule2` augment `/mws-main-module:main`. `mws-main-submodule2` imports `mws-other-module` and uses a leafref type that references `/mws-other-module:foo/mws-other-module:bar`. If `mws-main-module` includes `mws-main-submodule1` before `mws-main-submodule2`, installation will fail. On the other hand, if `mws-main-module` includes `mws-main-submodule2` before `mws-main-submodule1`, or model `mws-other-module` is loaded beforehand, installation will succeed.

```
      +--------------+    Include    +--------------+
      |              |-------------->|              |
      |  mws-main-   |               |  mws-main-   |
      |   module     |<--------------|  submodule1  |
      +--------------+    Augment    +--------------+
          |      ^
 Include  |      |  Augment
          v      |
      +--------------+    Import     +--------------+
      |              |-------------->|              |
      |  mws-main-   |               |  mws-other-  |
      |  submodule2  |-------------->|    module    |
      +--------------+    Leafref    +--------------+
                           XPATH
```

The install command and output can be seen below:

```console
# sysrepoctl --install --yang mws-main-module.yang 
Installing a new module from file 'mws-main-module.yang'...
Installing the YANG file to '/path/to/yang/repo/repository/yang/mws-main-module.yang'...
Resolving dependency: 'mws-main-module' includes 'mws-main-submodule1'...
Installing the YANG file to '/path/to/yang/repo/repository/yang/mws-main-submodule1.yang'...
Resolving dependency: 'mws-main-module' includes 'mws-main-submodule2'...
Installing the YANG file to '/path/to/yang/repo/repository/yang/mws-main-submodule2.yang'...
Resolving dependency: 'mws-main-submodule2' imports 'mws-other-module'...
Installing the YANG file to '/path/to/yang/repo/repository/yang/mws-other-module.yang'...
[...]

PR #1126, opened 2018-05-16
Tests and models to verify successful installation when submodule contains a leafref

aweast:
### Description
This is a follow-on pull request as requested by @michalvasko that contains only the test case and models which should now pass due to #1125 .  This PR is associated with issue #1116 

### Test case
The included test case attempts to install a valid model that would previously fail without @michalvasko 
fix in #1125 . This yang model is the same one attached to issue #1116

### Closure
Tests support PR #1125
Closes issue #1116

### DIFF
--- a/tests/md_test.c
+++ b/tests/md_test.c
@@ -1673,6 +1673,36 @@ md_test_insert_module_2(void **state)
 }
 
 /*
+ * @brief Test md_insert_module().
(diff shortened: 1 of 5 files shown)


michalvasko:
Hi,
thanks, it works now.

Regards,
Michal

Issue #1106, opened 2018-04-27
sysrepoctl assertion failure while installing module with augmented empty container 


aweast:
### First items attempted
* [x] Did you check if this is a duplicate issue?
* [x] Did you test it on the latest sysrepo devel branch?
* [x] Did you test it on the latest sysrepo devel branch and the latest libyang devel branch?

### Description
Installing a YANG module with sysrepoctl results in a failure due to an assertion if an otherwise empty container has children added  by augmentation.  The following YANG module exhibits the issue.

```YANG
module test-ds0 {
  namespace "urn:ietf:params:xml:ns:yang:ietf-interfaces:test-ds0";
  prefix ds0;

  import ietf-interfaces {
    prefix if;
  }

  augment "/if:interfaces-state/if:interface" {
      container foo {
          presence "This is a presence node";
          description
            "Container-specific state data.";
      }
  }

  augment "/if:interfaces-state/if:interface/foo" {
      leaf permit-foo-management {
          config false;
          type boolean;
          default "true";
          description
            "If true, permit foo management.";
      }
  }
}
```

The install command and output can be seen below.
```console
root@host:/# sysrepoctl --install --yang=/path/to/yang/models/ds0.yang --owner=root:root --permissions=666
Installing a new module from file 'models/test-ds0.yang'...
Installing the YANG file to '/opt/dev/sysrepo/cmake-build-debug/repository/yang/test-ds0.yang'...
Resolving dependency: 'test-ds0' imports 'ietf-interfaces'...
Installing the YANG file to '/opt/dev/sysrepo/cmake-build-debug/repository/yang/ietf-interfaces@2014-05-08.yang'...
sysrepoctl: /opt/dev/sysrepo/src/module_dependencies.c:1747: md_traverse_schema_tree: Assertion `!(child->nodetype & (LYS_CONTAINER | LYS_LIST)) || (intptr_t)child->priv & PRIV_OP_SUBTREE' failed.
Aborted (core dumped)
```

**Expected behavior:**
The module should install successfully.

**Actual behavior:**
Installing the module results in an assertion failure in  [module_dependencies.c#1747](https://github.com/sysrepo/sysrepo/blob/c4fc434f0640905a3b4e3378151d2c399fba945a/src/module_dependencies.c#L1747)

### Versions
Versions used to reproduce the bug:
Sysrepo version: sysrepod - sysrepo daemon, version 0.7.3
[...]

PR #1129, opened 2018-05-17
Resolve module installation assertion when empty container in augment and double augments

### LINKED ISSUE #1128: Assertion when installing a module with an augment containing an empty container
### First items attempted
* [x] Did you test it on the latest sysrepo devel branch and the latest libyang devel branch?

### Description
An assert is triggered when installing a module with an augment containing an empty container. In the included yang models below, we would expect to be able to install `main-aug`. This also leads to further use cases, such as `main-aug2`, which also currently fails installation. 

When looking at the assert while installing `main-aug`, we believe it is an issue with the child->priv not being set correctly when the container is initially empty. This causes the assert in module_dependencies.c#1745 in the devel branch when `main-aug`'s `foo` container is the child variable.

Similarly, the assert while installing `main-aug2` has the same problem. In this case, when we hit the assert on line 1745, the child->priv value is still not correctly set despite the child variable having children node itself. It appears that all of the child variable's descendent nodes do not have a priv value either.

```YANG
module main {
    namespace "test:main";
    prefix m;

    container main {
        config false;
        leaf baz {
            type string;
        }
    }
}
```
```YANG
module main-aug {
    namespace "test:main-aug";
    prefix ma;

    import main {
        prefix m;
    }

    augment "/m:main" {
        container foo {
            
        }
    }
}
```
```YANG
module main-aug2 {
    namespace "test:main-aug2";
    prefix ma2;

    import main {
        prefix m;
    }
    import main-aug {
        prefix ma;
    }

    augment "/m:main/ma:foo" {
        container bar {
            leaf foo { type string;}
        }
    }
}
```

The install command and output for `main-aug` can be seen below.
```console
#sysrepoctl --install --yang=path/to/yang/models/main-aug.yang 
Installing a new module from file 'path/to/yang/models/main-aug.yang'...
Installing the YANG file to '/path/to/repository/yang/main-aug.yang'...
Resolving dependency: 'main-aug' imports 'main'...
Installing the YANG file to '/path/to/repository/yang/main.yang'...
sysrepoctl: /path/to/src/sysrepo/src/module_dependencies.c:1745: md_traverse_schema_tree: Assertion `child == NULL || !(child->nodetype & (LYS_CONTAINER | LYS_LIST)) || (intptr_t)child->priv & PRIV_OP_SUBTREE' failed.
Aborted (core dumped)
```

The install output for `main-aug2` shows the same assert message.

**Expected behavior:**
The modules should install successfully.

**Actual behavior:**
Installing the module caused an assertion error, possibly because of the way the empty container in an augment is handled.

### Versions
Sysrepo version: sysrepod - sysrepo daemon, version 0.7.3
Sysrepoctl version: sysrepoctl - sysrepo control tool, version 0.7.3
libyang version: yanglint 0.15.96

**Additional Info**
This issue looks very similar to Issue #1106, but with the second augmentation layer in a different model.

aweast:
### Description
Installing a YANG module with sysrepoctl results in a failure due to an assertion when the module contains an augment containing an empty container.  This was very similar to #1106, but with the second augmentation layer in a different model.

### Test case
The test case, `md_test_insert_module_double_aug`, found in `md_test.c` and accompanying yang model, `mwa-aug2`, demonstrate the failure.

### Closure
closes #1128

### DIFF
--- a/src/module_dependencies.c
+++ b/src/module_dependencies.c
@@ -1742,8 +1742,16 @@ md_traverse_schema_tree(md_ctx_t *md_ctx, md_module_t *module, struct lys_node *
                         if (augment) {
                             if (node->nodetype == LYS_AUGMENT) {
                                 child = (struct lys_node *)lys_getnext(NULL, node, NULL, 0);
-                                assert(child == NULL || !(child->nodetype & (LYS_CONTAINER | LYS_LIST)) || (intptr_t)child->priv & PRIV_OP_SUBTREE);
-                                assert(child == NULL || child->nodetype & (LYS_CONTAINER | LYS_LIST) || child->flags & LYS_CONFIG_R);
+                                if (child != NULL) {
+                                    if (child->nodetype & (LYS_CONTAINER | LYS_LIST)) {
+                                        /* All op data means containers/lists containing children will have 
+                                           PRIV_OP_SUBTREE set. Empty containers/lists will not have this set. 
+                                           Neither will containers that only have children from a different schema. */
+                                        assert(((intptr_t)child->priv & PRIV_OP_SUBTREE) || child->child == NULL || main_module_schema != lys_node_module(child->child));
+                                    } else {
+                                        assert(child->flags & LYS_CONFIG_R);
+                                    }
+                                }
                             } else {
                                 child = node;
                             }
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi,
seems fine, thanks.

Regards,
Michal

PR #1131, opened 2018-05-22
Fix for the data feature interdependency scenario in sysrepo#1130

### LINKED ISSUE #1130: Edit configs fail for a certain data feature interdependency scenario
* [X] Did you check if this is a duplicate issue?
* [X] Did you test it on the latest sysrepo devel branch?

### Description
When models are loaded with `dm_module_clb` (typically when copying data from one module into another with a dependency on that data), the models are not loaded quite the same way as they are via the normal model-loading functions in `data_manager.c`. This leads to several problems when there are complex dependency graphs with data references and augments and feature toggles.

Currently if a module has a data dependency on another module and that module is itself augmented by a third module, and that third module's augmented data depends on a feature defined in yet a fourth module, that data won't be properly copied into the first module's context properly. This is because the third module is loaded with `dm_module_clb` and the feature from the fourth module isn't marked as enabled.

Additionally, if a module is loaded into the context due to being imported by an extension, but then is also required to copy a data-dependency tree into that context due to e.g. another augment, it won't be copied properly because the augment won't be resolved due to the module not being marked as "implemented".

### Proposed solution
A proposed solution and related tests are implemented in alangefe/sysrepo@ab596d5.

The proposed solution does several things:
- Whenever a module is loaded by `dm_module_clb`:
   - All imports of the loaded module (and the module itself) have any features enabled (according to the module_deps info)
   - All imports of the loaded module are marked as implemented if they are marked as implemented in the module_deps info (as there might not be a chance to do so later)
- Whenever a module is loaded as an extension or data dep of another module during normal module loading (as well as the module being loaded itself):
   - All imports of the loaded module are marked as implemented if they are marked as implemented in the module_deps info (as there might not be a chance to do so later)
- Validation issues were sometimes being lost in `dm_validate_session_data_trees` because the error code didn't match up. This has been changed to record all error types, not just `SR_ERR_VALIDATION_FAILED`.

The included unit tests have two sets of models.

For testing feature-enabling:
```
+------------------+   Imports    +------------------+
|data feat enable A+-------------->data feat enable C|
+---------^--------+              +------------------+
          |
          |Extension
          |
          |
+---------+--------+  Data refer. +------------------+
|data feat enable D+-------------->data feat enable B|
+------------------+              +------------------+
```

For testing marking-as-implemented:
```
+--------------+  Data ref.  +--------------+
|data imp dep A+------------->data imp dep C|
+------+-------+             +-------+------+
       |                             |
       | Imports            Extension|
       |                             |
       |      +--------------+       |
       +------>data imp dep B<-------+
              +--------------+
```

alangefe:
### Description
This pull request contains a proposed solution to issue #1130 and test cases to demonstrate the issue. The issue involves a data feature interdependency scenario. A more detailed description can be found in the issue notes.

A short summary of the changes:
- Enable features of imported modules from the module callback.
- Mark imported modules as implemented if they're considered implemented in the dependency graph.
- Fix reporting of other error codes when validating changes
- Test cases to cover the changes.

### Closure
closes #1130

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -400,6 +400,85 @@ cleanup:
     return rc;
 }
 
+static int
+dm_enable_features_with_imports(dm_ctx_t *dm_ctx, md_module_t *module, const struct lys_module *ly_module, struct ly_ctx *target_ctx)
+{
+    const struct lys_module *dest_module = NULL, *src_module = NULL;
+    dm_schema_info_t *si = NULL;
+    int ret = 0, rc = 0;
+
+    rc = dm_get_module_and_lock(dm_ctx, module->name, &si);
+    CHECK_RC_LOG_RETURN(rc, "Schema '%s' not found", module->name);
+    ret = sr_features_clone(si->module, ly_module);
+    pthread_rwlock_unlock(&si->model_lock);
+    CHECK_ZERO_LOG_RETURN(ret, SR_ERR_INTERNAL, "Failed to clone feature into module '%s'", module->name);
+
+    for (sr_llist_node_t *node = module->deps->first; node; node = node->next) {
+        md_dep_t *dep = (md_dep_t*)node->data;
+
+        if (dep->type != MD_DEP_IMPORT) {
+            continue;
+        }
+
+        dest_module = ly_ctx_get_module(target_ctx, dep->dest->name, NULL, 0);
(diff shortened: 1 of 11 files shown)


michalvasko:
Hi Andrew,
thanks for a nice contribution, seems fine to me.

Regards,
Michal

PR #1136, opened 2018-05-23
Reduce scope of test module installs from #1131

alangefe:
This is a small adjustment to the tests we submitted in #1131. I moved the module installation and feature enabling calls out of tests/CMakeLists.txt, where they would apply to all test cases, and instead added them to a setup function so that they only happen before the two `rp_dt_edit_test` test cases that actually need those modules. (This is similar to how the `cl_test` tests work.)


michalvasko:
Hi Andrew,
it is much cleaner now, thanks.

Regards,
Michal

Issue #1166, opened 2018-06-11
Failure when using identity if the identity's base is in an unimplemented model


aweast:
### First items attempted
* [x] Did you check if this is a duplicate issue?
  * This is a similar issue to #1111, but not identical
* [x] Did you test it on the latest sysrepo devel branch?
* [x] Did you test it on the latest sysrepo devel branch and the latest libyang devel branch?

### Description
There appears to be an issue with the use of identity-refs when sending an edit-config that uses an identity whose base is in an unimplemented model. When building sysrepo with tests, the id-ref and id-def modules almost demonstrate the issue. To fully demonstrate the issue, you can modify tests/CMakeList.txt to only install `id-ref-aug` and `id-def-extended` and not `id-ref-base` or `id-def-base`. After building sysrepo with tests, we see the following in `sysrepoctl --list`, which we believe is the expected output:

```
Module Name                | Revision   | Conformance | Data Owner          | Permissions | Submodules                    | Enabled Features
-----------------------------------------------------------------------------------------------------------------------------------------------
id-ref-aug                 |            | Installed   |                     |             |                               |
id-ref-base                |            | Implemented | developer:developer | 666         |                               |
id-def-base                |            | Imported    |                     |             |                               |
id-def-extended            |            | Installed   |                     |             |                               |
```

Now, when we run the rp_dt_edit_test, the set_item_id_ref test fails with the following output:
```
25: [ RUN      ] set_item_id_ref
25: [DBG] (rp_session_start:4087) RP session start, session id=123456.
25: [DBG] (sr_locking_set_lock_fd:789) File /opt/dev/sysrepo/cmake-build-debug/repository/data/id-ref-base.persist has been locked
25: [DBG] (pm_load_data_tree:267) Persist data successfully loaded from file '/opt/dev/sysrepo/cmake-build-debug/repository/data/id-ref-base.persist'.
25: [DBG] (sr_locking_set_unlock_close_fd:858) File /opt/dev/sysrepo/cmake-build-debug/repository/data/id-ref-base.persist (fd = 3) has been unlocked
25: [DBG] (sr_locking_set_lock_fd:789) File /opt/dev/sysrepo/cmake-build-debug/repository/data/id-ref-base.persist has been locked
25: [DBG] (pm_load_data_tree:267) Persist data successfully loaded from file '/opt/dev/sysrepo/cmake-build-debug/repository/data/id-ref-base.persist'.
[...]

PR #1167, opened 2018-06-11
Test modification to demonstrate problem when an identities base is in an unimplemented model

aweast:
### Description

To demonstrate the issue in #1166 , we removed id-def-base and id-ref-base from the list of automatically installed yang models. There is only one test that currently uses these models, so this change should be valid.

We also extended the test to attempt to get the value. With an unmodified CMakeList.txt file, we were seeing the rp_dt_get_value_wrapper fail with rc = SR_ERR_NOT_FOUND. Any fix that fixes the set-item operation should also ensure the get-item operation succeeds.

### Test case
Modified `set_item_id_ref` to be `set_and_get_item_id_ref` and removed installation of models from CMakeList.txt.

### Closure
Demonstrates issue filed in #1166


michalvasko:
Hi,
if you properly clean-up after the test, it should pass.

Regards,
Michal

michalvasko:
Okay, thanks.

Regards,
Michal

PR #1174, opened 2018-06-13
Handle circular references in module imports

aweast:
### Description
During some feature enables, we found some infinite recursion loops.  The pull request at https://github.com/sysrepo/sysrepo/pull/1171 fixed part of our issues.  This pull request fixes the remaining recursion loops we are seeing.

Let us know if we need to provide specific examples of the yang models we were seeing this on.


michalvasko:
Of course, thanks.

Regards,
Michal

PR #1177, opened 2018-06-14
Default sysrepo repository location

michalvasko:
### Description
Use default repository location so that all tests are executed.

Also, always use prefixes for idetityref values.


michalvasko:
@jktjkt In your previous pull request you have defined a custom repository location in one travis build and then it differed from the test repository so some tests were not executed.  I have removed all of it and now it works correctly. Can I merge this or there was some purpose to your changes?

Regards,
Michal

jktjkt:
@michalvasko 
I believe that the CI should *also* test a build which is close to what the end users are going to deploy -- and that means an out-of-tree repository location. That's why we're actually building sysrepo twice in our internal CI; once with all tests and an in-tree repo, and once again with a persistent repository location.

However, we were seeing hard-to-trace breakages when we were updating sysrepo because Travis was not testing that configuration. That's why I wanted to add a new configuration to the Travis build matrix to ensure that these failures won't hit me when I update. At the same time I wanted to keep the number of tested configurations low enough, so I only repurposed one of the existing jobs.

Any of these is fine with me:

- add one more job which explicitly tests an out-of-tree repo location
- revert this commit, i.e. have one of the existing jobs repurposed to *also* test an out-of-tree repo location in addition to no custom memory management

PR #1201, opened 2018-07-02
Improve search times for interconnected yang models.

bhart3:
### Description
When running the sysrepo with a large number of interconnected yang models, we have recently noticed a substantial slowdown during the `dm_load_module` process. We would like to propose the following changes:

#### Change the use of sr_list_t to sr_btree_t
With a large number of installed models and dependencies, the `loaded_deps` list must be traversed a large number of times. We replaced the use of `sr_list_t` with `sr_btree_t` for improved search times.

For the btree comparison function, we compared the pointer values instead of the module names. Are there any forseeable issues with this approach?

#### Move the module-in-context check
Previously, `dm_apply_persist_data_for_model` would check to see if a model was in the target context before attempting to load the persist data. However, before `dm_apply_persist_data_for_model` is called, `dm_apply_persist_data_for_model_imports` already tried to process all of the model's imports. We moved the check into `dm_apply_persist_data_for_model_imports` so the imported models are only processed if the current model is in the target context.

#### Add a completed_deps list
The `dm_apply_persist_data_for_model_imports` function has one recursive call that is not blocked by a check on the `loaded_deps` list. We believe that this is the correct behavior. However, this introduces a way for a single model to be processed by this function many times. To fix this, we added a `completed_deps` list. When the persist data of a model and all of its imported models are completely loaded, the model is added to the `completed_deps` list. If `dm_apply_persist_data_for_model_imports` is called on a model that is already in `completed_deps`, we can safely assume that all of the required persist data has already been loaded and exit the function immediately.

### Test case
All current tests pass. No new functionality added.

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -871,15 +871,25 @@ dm_enable_module_running_internal(dm_ctx_t *ctx, dm_session_t *session, dm_schem
 }
 
 static bool
-dm_load_module_deps_in_list(md_module_t *dep_module, sr_list_t *loaded_deps)
+dm_search_module_deps_in_tree(md_module_t *dep_module, sr_btree_t *loaded_deps)
 {
-    for (size_t i = 0; loaded_deps && (i < loaded_deps->count); ++i) {
-        if (0 == strcmp(dep_module->name, ((md_module_t *)loaded_deps->data[i])->name)) {
-            return true;
-        }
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi,
it seems you are right and the patch is clean enough, so no problem, merging. Thanks.

Regards,
Michal

Issue #1208, opened 2018-07-04
Cannot subscribe to RPCs


rjarry:
Hi,

I get an error when subscribing to rpc invocations (RPC_SUBS).

```
root@localhost:~# sysrepoctl -i --yang=example.yang
```

```c
res = sr_rpc_subscribe(session, "/example:do-it", cb, priv, opts, &subs);
```

```
Jul 04 13:39:15 ubuntu1604 sysrepod[8295]: [ERR] Cannot copy module 'example', it is not enabled.
Jul 04 13:39:15 ubuntu1604 sysrepod[8295]: [ERR] Unable to enable the module in the running datastore.
```

Here is the YANG module:

```
module example {
  rpc do-it {
    input {
      leaf args {
        type string;
      }
    }
    output {
      leaf result {
        type int8;
      }
    }
  }
}
```

After looking at the code, it seems that sysrepo is considering that a yang module that has no data nodes, cannot be "enabled".

If I add a dummy data node in the module, it goes a bit further but I get another error:

```
Jul 04 15:18:59 ubuntu1604 sysrepod[8288]: [ERR] libyang error: When condition "../type = 'module-change' or ../type = 'subtree-change' or ../type = 'dp-get-items'" not satisfied.
Jul 04 15:18:59 ubuntu1604 sysrepod[8288]: [ERR] Parsing persist data from file '/etc/sysrepo/data/example.persist' failed: When condition "../type = 'module-change' or ../type = 'subtree-change' or ../type = 'dp-get-items'" not satisfied.
Jul 04 15:18:59 ubuntu1604 sysrepod[8288]: [ERR] Unable to load persist data tree for module 'example'.
Jul 04 15:18:59 ubuntu1604 sysrepod[8288]: [ERR] Unable to save the subscription into persistent data file.
```

I looked in the file, I have the subscription registered:

```xml
<module xmlns="urn:ietf:params:xml:ns:yang:sysrepo-persistent-data">
  <name>example</name>
  <subscriptions>
    <subscription>
      <type>rpc</type>
      <destination-address>/var/run/sysrepo-subscriptions/example/8306.example.18580.sock</destination-address>
      <destination-id>18580</destination-id>
      <enable-running/>
      <xpath>/example:do-it</xpath>
      <api-variant>values</api-variant>
    </subscription>
  </subscriptions>
</module>
```

I don't know if the `sysrepo-persistent-data.yang` model is incorrect or if `<enable-running/>` should not be here.

[...]

PR #1209, opened 2018-07-05
persistence_manager: only write <enable-running /> for supported subscriptions

rjarry:
When accessing the sysrepo protobuf API without using the client library, it is possible to send a `RPC_SUBS` message with `->enable_running = true`. Sysrepo accepts the subscription and adds it in the persistant-data with a `<enable-running />` tag.

This leads to an error when sysrepo tries to read the file again and parse it with libyang:

```
[ERR] libyang error: When condition "../type = 'module-change' or ../type = 'subtree-change' or ../type = 'dp-get-items'" not satisfied.
[ERR] Parsing persist data from file '/etc/sysrepo/data/example.persist' failed: When condition "../type = 'module-change' or ../type = 'subtree-change' or ../type = 'dp-get-items'" not satisfied.
```

Make sure to respect the condition in the yang module.

This fixes the second part of #1208 but the problem with RPCs in yang modules that have no data nodes still remains. I did not find a way to fix it.

### DIFF
--- a/src/persistence_manager.c
+++ b/src/persistence_manager.c
@@ -1138,7 +1138,10 @@ pm_add_subscription(pm_ctx_t *pm_ctx, const ac_ucred_t *user_cred, const char *m
     CHECK_RC_MSG_GOTO(rc, cleanup, "Unable to add new subscription into the data tree.");
 
     /* set subscription details */
-    if (subscription->enable_running) {
+    if (subscription->enable_running && (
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you are not supposed to access sysrepo protobuf API directly but always use the client library. Anyway, the patch seems fine, so I can merge it.

Regards,
Michal

rjarry:
Thanks !

PR #1210, opened 2018-07-05
Tests to demonstrate failure when identityref references identity defined in imported model

aweast:
### Description
Similar to https://github.com/sysrepo/sysrepo/pull/1167.  I believe this is an edge case that we didn't catch in the previous tests.

### Test case
Test cases `ident_ref_in_installed_module` and `extended_ident_ref_in_installed_module` demonstrate the problem using models, `id-ref-main.yang`, `id-ref-extended-main.yang`, `id-ref-imported.yang`, `id-ref-installed.yang`.
### Closure
Demonstrates an edge case for #1166 that was not caught in #1167


michalvasko:
Hi,
it is passing now, thanks for the test.

Regards,
Michal

PR #1211, opened 2018-07-05
Fixing segfault when adding and deleting list row to running config with NACM enabled

aweast:
### Description
When NACM is enabled, creating a row, committing, and deleting a row from a list is causing a segfault.

### Test case
`add_delete_list_row_with_nacm_test` demonstrates the issue which is fixed by the changes in `data_manager.c`

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -4766,7 +4766,8 @@ dm_perform_netconf_access_control(nacm_ctx_t *nacm_ctx, dm_session_t *session, d
     }
 
     /* start NACM data access validation for this module */
-    rc = nacm_data_validation_start(nacm_ctx, session->user_credentials, new_info->node->schema,
+    struct lys_node *schema = new_info->node == NULL ? prev_info->node->schema : new_info->node->schema;
+    rc = nacm_data_validation_start(nacm_ctx, session->user_credentials, schema,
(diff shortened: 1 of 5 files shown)


michalvasko:
Hi,
right, so did your minor change fix the problem or is the crash still occurring?

Regards,
Michal

aweast:
@michalvasko, sorry for not getting back on this as I have been away for a few weeks.  I believe the minor change fixes the problem and it also passes on my machine.  I did see this in the travis CI log:
``` 
[DBG] [67340672] (nacm_report_edit_access_denied:1945) User 'travis' was blocked from creating the node '/commit-nacm-segfault:test-list[test-key='test-key']' by NACM.
``` 
It looks like the `travis` user does not have the correct permissions for the test ?

aweast:
@michalvasko, with the latest commit, it is much closer.  4/5 Travis CI jobs pass; however, one still fails with a unique error which I have not seen before:
``` 
[ERR] [139826197333824] (srctl_ly_log_cb:506) libyang: Unable to use search directory "/tmp/sysrepo-repository/yang/" (No such file or directory)
```

aweast:
@michalvasko, thank you for the suggestion.  I believe that did it.

PR #1213, opened 2018-07-06
C++: bugfix in exceptions

### LINKED ISSUE #1206: c++: module_change sub of augment-only module fails silently
* [x] Did you test it on the latest sysrepo devel branch? (logs generated with 724a62f but verified still exists in 5c05f3e)

### Description
Creating a module_change subscription on a module which contains only an augment and no local config fails silently with no exception thrown.

### Steps to Reproduce
1. Install module with no local config, only augmented
2. Create a module_change subscription to the module

I used the following yang:

```
module my-augment
{
  yang-version 1;

  namespace "http://example.net/my-augment";
  prefix "my-augment";

  import turing-machine { prefix "tm"; }

  description "Test augment model";

  revision 2018-07-03
  {
    description "Initial revision.";
    reference "TBD";
  }

  augment "/tm:turing-machine"
  {
    leaf my-augment
    {
      type string;
    }
  }

  /* Uncomment to enable local config */
//  leaf native-leaf { type string; }

  rpc my-rpc
  {
    description "rpc test";
  } // rpc my-rpc

} // module my-augment
```

**Expected behavior:**
The failure in module_change_subscribe() should trigger an exception -- or this call should succeed with the subscription applying to the augmented nodes.

**Actual behavior:**
The code finished without exception, even though the subscription was invalid -- evidenced by inspecting the value of *subscription->swig_sub, which is 0 and nonzero on a valid subscription. (Checking this makes for a useful workaround, commented in the attached source file.)

### Versions
Please provide versions for software needed to reproduce the bug:
Sysrepo version: 0.7.4 (devel branch, see above)
libyang version: 0.15.120

### Attachments
- [sysrepo_aug_sub_fail.zip](https://github.com/sysrepo/sysrepo/files/2160564/sysrepo_aug_sub_fail.zip) Test program, yang module and debug logs from the test program and sysrepod. I ran tests both with local config enabled and disabled.

mislavn:
### Description
Bugfix for not thrown C++ exception.

### Closure
Fixes #1206

### DIFF
--- a/swig/cpp/src/Sysrepo.cpp
+++ b/swig/cpp/src/Sysrepo.cpp
@@ -43,42 +43,7 @@ sr_error_t sysrepo_exception::error_code() const
 }
 
 void throw_exception(int error) {
-    switch(error) {
-    case(SR_ERR_INVAL_ARG):
-        throw sysrepo_exception(SR_ERR_INVAL_ARG);
-    case(SR_ERR_NOMEM):
-        throw sysrepo_exception(SR_ERR_NOMEM);
-    case(SR_ERR_NOT_FOUND):
-        throw sysrepo_exception(SR_ERR_NOT_FOUND);
-    case(SR_ERR_INTERNAL):
-        throw sysrepo_exception(SR_ERR_INTERNAL);
-    case(SR_ERR_INIT_FAILED):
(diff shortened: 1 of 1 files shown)


mislavn:
OK,

I'll keep that on mind from now on.

Issue #1218, opened 2018-07-09
C++: Add set_error()


kpbarrett:
* [x] Did you check if this is a duplicate issue?
* [x] Did you test it on the latest sysrepo devel branch?

### Description
Enhancement: Please add set_error() to Session class

PR #1219, opened 2018-07-09
Add Session::set_error()

kpbarrett:
### Description
Solves issue #1218


swig/cpp/src/Session.h
```diff
@@ -79,6 +79,7 @@ class Session
     void set_options(const sr_sess_options_t opts);
     S_Iter_Change get_changes_iter(const char *xpath);
     S_Change get_change_next(S_Iter_Change iter);
+    int set_error(const char *message, const char *xpath);
```
jktjkt: This should not return anything, i.e.:
```c++
void set_error(const char *message, const char *xpath);
```
This method already throws an exception when the error message cannot be set. That's how the C++ wrapper works. There's no need to always return a dummy value indicating "success". If it ever returns, then it succeeded.

PR #1226, opened 2018-07-11
Swig refactor

mislavn:
### Description
Refactored the Python bindings and simplified them, now there is no need for two separate sets of examples and unit tests for Python2 and Python3.

The `libsysrepoPython2` and `libsysrepoPython3` python modules are renamed to `sysrepo`.

### Test case
Updated the Python unit tests.


mislavn:
@michalvasko I renamed the Python bindings to `sysrepo` but if you want it can be easily changed to anything at this stage, now is the best time to do that.

michalvasko:
Hi Mislav,
sure, it is fine.

Regards,
Michal

PR #1227, opened 2018-07-12
Add request id in data provide items callback

samuel-gauthier:
### Description
 When a data request is made, sysrepo will ask to a subscriber for each level under the subscription, calling the subscriber's callback once per layer.

There are cases when this is a problem, because we can not be sure if the data is consistent, as it was asked for in several separate requests.

This patch series adds the request_id from the originating get_item request to the sr_dp_get_items_cb prototype. Then, the subscriber can know if his callback is called for one state data request or several by checking the given request id. The subscriber can take the necessary mesure to ensure the data is consistent.
It also changes the way data request message id are calculated. It is now increasing with the global number of messages sent, and is not per session.

Note that those patches breaks the API, as the sr_dp_get_items_cb prototype has a new request id argument.

### Test case
A new test has been added, named cl_request_id, part of cl_state_data_test. This test checks that several callbacks called for one request have the same request id, and that another request has another request id.
The sysrepo unit tests are 100% after this patch series, including the new test. I have not made other tests.

### DIFF
--- a/doc/start.dox
+++ b/doc/start.dox
@@ -204,7 +204,7 @@ the specific changed values.
 @subsection state State Data
 ~~~{.c}
 static int
-oven_state_cb(const char *xpath, sr_val_t **values, size_t *values_cnt, void *private_ctx)
+oven_state_cb(const char *xpath, sr_val_t **values, size_t *values_cnt, uint64_t request_id, void *private_ctx)
 {
     sr_val_t *vals;
     int rc;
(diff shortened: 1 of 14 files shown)


michalvasko:
Hi,
seems fine, but you forgot to update `tests/measure_performance.c" state data callback, that is why the tests failed to compile. Please, fix that and they will hopefully compile and pass, thanks.

Regards,
Michal

michalvasko:
Hi Samuel,
thanks a lot, seems fine now.

Regards,
Michal

PR #1236, opened 2018-07-20
rpc: forward error details to requestor if any

rjarry:
If there is an error in the RPC response from the provider, register it in the data manager context so that it is written in the response sent to the requestor.


michalvasko:
Hi,
seems reasonable, thanks.

Regards,
Michal

rjarry:
Thanks @michalvasko !

PR #1266, opened 2018-08-16
Validate inverse data dependencies on commit

### LINKED ISSUE #1242: sysrepo doesn't reject the deletion of referred-to data

### Description
Currently sysrepo does not seem to properly reject the removal of data that is depended upon by a leaf-ref. This causes problems later when doing a `<get-config>` as the fully-merged running datastore is no longer valid.

I couldn't figure out how to cleanly express this in a unit test, so I have captured it in an integration test:

https://github.com/ADTRAN/netopeer2-integration-tests/blob/master/tests/test_data_refs.py#L38

Normally I'd try to come with the fix as well, but I feel like this might require a significant change.

### Steps to Reproduce
1. Create a list value
2. Create another value that is a leaf-ref for that list (in a different module)
3. Delete the original list value

or alternatively:

Run the integration test `test_create_resolved_data_and_removing_referenced_data_is_error` from https://github.com/ADTRAN/netopeer2-integration-tests

**Expected behavior:**
An error at step 3

**Actual behavior:**
No error is emitted at step 3

### Versions
Latest of each dev branch

alangefe:
### Description
Whenever a module is changed, any other modules that have data dependencies on that module should also be checked to ensure their data references are still valid.

There is a bit of a performance penalty for doing this, and ideally one would only check the inverse deps if the specific nodes being referenced were changed. That is an optimization that could be done in the future.

### Test case
These changes will cause the failing test case `test_create_resolved_data_and_removing_referenced_data_is_error` in https://github.com/ADTRAN/netopeer2-integration-tests to pass.

### Closure
Fixes #1242

### DIFF
--- a/src/data_manager.c
+++ b/src/data_manager.c
@@ -3427,6 +3427,92 @@ cleanup:
     return rc;
 }
 
+static void
+dm_record_errors(int rc, sr_error_info_t **errors, size_t *err_cnt, dm_data_info_t *info)
+{
+    if (SR_ERR_VALIDATION_FAILED == rc) {
+        if (SR_ERR_OK != sr_add_error(errors, err_cnt, ly_errpath(info->schema->module->ctx), "%s",
+                                        ly_errmsg(info->schema->module->ctx))) {
+            SR_LOG_WRN_MSG("Failed to record validation error");
+        }
+    } else {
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi Andrew,
that is how I started my implementation, using the inverted dependencies but quickly run into several problems so I decided to leave it for now. Thank you very much, you have saved me a considerable amount of time!

Regards,
Michal

Issue #1274, opened 2018-08-21
LYB: cross-compiles need a way to change file paths within data/internal/sysrepo-module-dependencies.lyb


jktjkt:
When cross-compiling sysrepo, it's common to build native binaries first, and then build a second version for the target. These two builds typically share the same sysrepo repository because the host tools are used for installing modules and feeding these with initial data.

With the XML format, it was easy to simply `sed` the resulting file prior to preparing the image for the target device, like this:
```
sed -i 's|$(TARGET_DIR)/etc/sysrepo|/etc/sysrepo|g' \
  $(TARGET_DIR)/etc/sysrepo/data/internal/sysrepo-module-dependencies.xml
```
This no longer works with the new default LYB format for sysrepo's data store. After `sed`ing, `sysrepod` won't start on the target device anymore:

```
sysrepod: /home/jkt/work/prog/_build/br-cfb/build/libyang-custom/src/parser_lyb.c:870:
lyb_parse_schema_hash: Assertion `hash[j - 1] & (LYB_HASH_COLLISION_ID >> (j - 1))' failed.
Aborted
```
On the other hand, not changing the path leads to various runtime errors on the target device because it attempts to access files which are not available, e.g.:
```
# sysrepocfg --export -f json -d running ietf-netconf-server
[ERR] (dm_ly_log_cb:1617) libyang error: Opening file
"/home/jkt/work/prog/_build/br-cfb/target/etc/sysrepo/yang/ietf-netconf-server@2016-11-02.yang" 
failed (No such file or directory).
Segmentation fault
```
This can be fixed in several ways:
- change sysrepo to produce and accept relative paths in the `sysrepo-module-dependencies` module
- provide a tool to perform this relocation of the LYB file

PR #1276, opened 2018-08-21
Store relative paths in sysrepo-module-dependencies

jktjkt:
Cross-compiles typically build sysrepo twice -- once for the host system (the machien doing the build), and once for the target (the machine which will run sysrepod in the end). The host build often runs with the repository location set to something like `${TARGET}/etc/sysrepo` where `${TARGET}` is a directory which will become the target's root filesystem.

Sysrepo creates one private data file which contains absolute paths to YANG schemas of all known modules. If we therefore use host's binaries to manage the target's upcoming repository, sysrepo stores host's paths in there. This makes sysrepod explode on target later on.

Previously, with the XML backend, I used to simply `sed` away the host's paths at image creation time. That way I could always use host's tools prior to finishing the image, and then execute a simple fixup just prior to deploying the image on the target. This no longer works when using the LYB serialization which is a new default.

This patch ensures that the paths stored in the repository are relative to the schema search dir. They are still being converted to an absolute path right after reading so that the rest of the code doesn't have to change.

The actual code cannot use `asprintf` because the existing code doesn't define neither `_BSD_SOURCE` nor `_GNU_SOURCE` and I'm too afraid to add that.

Fixes: #1274

### DIFF
--- a/src/common/sr_utils.h
+++ b/src/common/sr_utils.h
@@ -26,9 +26,6 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdbool.h>
-//! @cond doxygen_suppress
-#define __USE_XOPEN
-//! @endcond
 #include <time.h>
 
 #include <libyang/libyang.h>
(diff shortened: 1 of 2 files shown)


jktjkt:
I would appreciate if you take a look at it. Re `asprintf`, I know that the build complained about an implicit declaration of this function even though I included a correct header. I think that there are some implications and maybe even behavior changes when one adds a `_GNU_SOURCE`, so that's something that I do not want to do. I am not sure whether it affects portability to other C standard libraries.

michalvasko:
Hi,
seems to work now, merging.

Regards,
Michal

PR #1277, opened 2018-08-21
Include deviations as extension dependencies

### LINKED ISSUE #1275: Deviations are not included as extension dependencies in the dep graph

### Description
Deviations are not included as extension deps like augments are. This means that syrepo doesn't load the deviation, but netopeer2-server does and that causes some inconsistencies in the behavior.

### Steps to Reproduce
1. Add a deviation with a 'not-supported' node in it
2. `<get-config>` requests to netopeer2-server will start to fail.

**Expected behavior:**
The deviation should not break things.

**Actual behavior:**
The deviation breaks things.

### Versions
sysrepo 9281f63

alangefe:
Fixes #1275

A deviation can be thought of essentially as a fancy augment for most purposes. So this patch makes the behavior of deviations consistent with augments for the purposes of dependency resolution.

(Patch and test by @jwwilcox.)

### DIFF
--- a/src/module_dependencies.c
+++ b/src/module_dependencies.c
@@ -1830,6 +1830,7 @@ md_insert_lys_module(md_ctx_t *md_ctx, const struct lys_module *module_schema, c
     struct lys_include *inc = NULL;
     struct lys_ident *ident = NULL;
     struct lys_node_augment *augment = NULL;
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
seems fine to me, thanks for the patch!

Regards,
Michal

PR #1279, opened 2018-08-23
Modify rp_dt_reload_nacm() to wait for request to be processed

### LINKED ISSUE #1278: Race condition in application of new NACM configuration

### Description
We've noticed that there is a small window of time between when the server sends an OK in response to a change in NACM configuration and the time when the new NACM configuration actually goes into effect. Because of this, it is possible to send a NETCONF request after receiving an OK for the NACM configuration change and still see the new request processed according to the old NACM rules.

This seems to be due to the fact that `rp_dt_reload_nacm()` is an asynchronous operation that does not wait until the new NACM configuration has taken effect.

### Steps to Reproduce
1. Alter the NACM configuration and receive an OK response.
2. Send a NETCONF request that should be processed differently according to the new configuration.

**Expected behavior:**
The new NETCONF request should be processed according to the new configuration.

**Actual behavior:**
The new NETCONF request is sometimes processed according to the old configuration.

### Versions
sysrepo: 0.7.4 (latest devel commit)
libyang: 0.16.17 (latest devel commit)
netopeer2-server: 0.6.10 (latest devel-server commit)

alangefe:
Fixes #1278

This commit modifies `rp_dt_reload_nacm()` to wait until the NACM request has been fully processed before returning.

(Patch by @bhart3.)

### DIFF
--- a/src/common/sr_data_structs.c
+++ b/src/common/sr_data_structs.c
@@ -534,6 +534,23 @@ sr_cbuff_dequeue(sr_cbuff_t *buffer, void *item)
     return true;
 }
 
+bool
+sr_cbuff_search(sr_cbuff_t *buffer, void *item)
+{
+    if (NULL == buffer || 0 == buffer->count) {
+        return false;
+    }
+
+    for (size_t i = 0; i < buffer->count; ++i) {
+        if (item == ((uint8_t*)buffer->data + (buffer->head * buffer->elem_size * i))) {
+            SR_LOG_DBG("item: %p found in buffer: %p", item, (void*)buffer);
+            return true;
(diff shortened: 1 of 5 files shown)


michalvasko:
Hi,
it seems `usleep()` cannot be used, so can you please replace it  with `nanosleep()`? Other than that it seems fine and can be merged, thanks!

Regards,
Michal

alangefe:
Thanks, Michal. I've replaced `usleep()` with `nanosleep()`.

michalvasko:
Alright, seems fine to me now and can be merged.

Regards,
Michal

PR #1285, opened 2018-08-31
Moved Sysrepo.h to Sysrepo.hpp to avoid conflict with sysrepo.h

### LINKED ISSUE #1284: C++ bindings cannot be built on macOS
macOS has case insensitive file system by default and sysrepo uses both Sysrepo.h and sysrepo.h header file names. I changed the name of Sysrepo.h into Sysrepo.hpp in PR I'll provide.

### Things you may try first

### Description
macOS has case insensitive file system by default and sysrepo uses both Sysrepo.h and sysrepo.h header file names.

### Steps to Reproduce
1. on sysrepo (master branch, devel can't find LYD_ANYDATA_LYBD and co) directory: mkdir build
2. cd build
3. cmake -DCMAKE_BUILD_TYPE=Debug ..
4. make

**Expected behavior:**
Build works

**Actual behavior:**
Build fails with ```.../inc/sysrepo/trees.h:62:70: error: unknown type name 'sr_node_t' ``` after ```/swig/cpp/src/Internal.h:28:10: warning: non-portable path to file '"Sysrepo.h"'; specified path differs in case from file name on disk [-Wnonportable-include-path]```

### Versions
sysrepod - sysrepo daemon, version 0.7.4
yanglint 0.15.124

### Attachments
From make >& log

[log.txt](https://github.com/sysrepo/sysrepo/files/2340574/log.txt)

fxdupont:
### Description
C++ bindings uses a header which does not differ from another (C) file on macOS case insensitive file systems. Cf Issue #1284

### Test case
In fact I suggest to use another naming scheme for C++ headers, i.e. to move to the upper case first character to .hpp. Now I did it on the only conflict to keep the change minimal.

### Closure
Closes #1284


mislavn:
I intended to rename the headers like they are in [libyang](https://github.com/CESNET/libyang/tree/master/swig/cpp/src), I can make a pull request next week which will rename all C++ headers.

Also you need to use the devel branch as base.

fxdupont:
> I intended to rename the headers like they are in libyang, I can make a pull request next week which will rename all C++ headers.

=> perfect. Please stall my PR.

> Also you need to use the devel branch as base.

=> I tried but it didn't reach the swig stuff. I am attaching cmake and make logs in the case someone knows an easy fix like upgrading my libyang (something I'll try anyway).
[logc.txt](https://github.com/sysrepo/sysrepo/files/2341244/logc.txt)
[log.txt](https://github.com/sysrepo/sysrepo/files/2341249/log.txt)

fxdupont:
With devel libyang I managed to build devel sysrepo at the exception of lua bindings so if needed I can provide new PRs based on the devel branch (for #1285 and #1287).

fxdupont:
Addressed in devel branch. Closing.

PR #1290, opened 2018-09-03
add java package support

9606:
Because most Java applications have packages, it is necessary to add packages to the swig extension of Java.
The example and test cases are running correctly.


9606:
I did not change the C and c++ code, and I saw that the wrong test cases were the original test cases of C language, and I had no problem on my computer.

PR #1293, opened 2018-09-06
Java package

9606:
### Description
Because most Java applications have packages, it is necessary to add packages to the swig extension of Java.

### Test case
Continue to use the original test cases.

### DIFF
--- a/swig/java/CMakeLists.txt
+++ b/swig/java/CMakeLists.txt
@@ -12,9 +12,9 @@ set(CMAKE_SWIG_FLAGS "-c++")
 set(CMAKE_SWIG_FLAGS "-I${PROJECT_SOURCE_DIR}")
 set(CMAKE_SWIG_OUTDIR "${CMAKE_CURRENT_BINARY_DIR}")
 
-set_source_files_properties(${JAVA_SWIG_BINDING}.i PROPERTIES CPLUSPLUS ON)
+set_source_files_properties(${JAVA_SWIG_BINDING}.i PROPERTIES CPLUSPLUS ON SWIG_FLAGS "-package;com.sysrepo")
 
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/classes")
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/com/sysrepo")
 
 swig_add_module(${JAVA_SWIG_BINDING} java ${JAVA_SWIG_BINDING}.i)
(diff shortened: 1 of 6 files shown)


michalvasko:
Hi,
seems fine to me.

@mislavn Do you have any problems with this change?

Regards,
Michal

9606:
@michalvasko Hello, I have a question why there are so many methods to be ignored in the java_base.i file. What if I just want to use the methods I ignore?
In fact, I need to use this method in the process of using sysrepo:
Val::set(char const *, uint32_t, sr_type_t);
Then I commented out all the %ignores in the java_base.i file, then recompiled and used them, and there was no problem during use.

mislavn:
So the issue is that Java does not support some types like `uint16_t` and the function constructors get overridden so in Java only `Val::set(const char *xpath, int64_t int64_val, sr_type_t type)` is available. We still haven't solved the issue of `uint64_t` in Java, probably will need to use BigInteger or something like that.

```
void Val::set(const char *xpath, int64_t int64_val, sr_type_t type) {
    if (_val == nullptr) throw_exception(SR_ERR_OPERATION_FAILED);

    int ret = sr_val_set_xpath(_val, xpath);
    if (ret != SR_ERR_OK) throw_exception(ret);

    if (type == SR_UINT64_T) {
        _val->data.uint64_val = (uint64_t) int64_val;
    } else if (type == SR_UINT32_T) {
        _val->data.uint32_val = (uint32_t) int64_val;
    } else if (type == SR_UINT16_T) {
        _val->data.uint16_val = (uint16_t) int64_val;
    } else if (type == SR_UINT8_T) {
        _val->data.uint8_val = (uint8_t) int64_val;
    } else if (type == SR_INT64_T) {
        _val->data.int64_val = (int64_t) int64_val;
    } else if (type == SR_INT32_T) {
        _val->data.int32_val = (int32_t) int64_val;
    } else if (type == SR_INT16_T) {
        _val->data.int16_val = (int16_t) int64_val;
    } else if (type == SR_INT8_T) {
        _val->data.int8_val = (int8_t) int64_val;
    } else {
        throw_exception(SR_ERR_INVAL_ARG);
    }

    _val->type = type;
}
```

9606:
@michalvasko uint64_t will be converted to BigInteger  in swig 3.0([http://www.swig.org/Doc3.0/SWIGDocumentation.html](http://www.swig.org/Doc3.0/SWIGDocumentation.html)):

unsigned long long const unsigned long long & | java.math.BigInteger | jobject
-- | -- | --

I will test to see if there is a problem recently.If there is no problem, I will commit again.
The previous two pull requests are a bit confusing, I have closed, the new pull request is at https://github.com/sysrepo/sysrepo/pull/1294

PR #1295, opened 2018-09-07
Namespace refactor

mislavn:
### Description
Add sysrepo namespace to C++, this includes some code in Python and Java bindings,
also the headers are renamed from ".h" to ".hpp".


mislavn:
This PR covers https://github.com/sysrepo/sysrepo/pull/1285.

michalvasko:
Hi Mislav,
I do not think that is a good solution. It would be better if you put the DEB packages directly into *sysrepo* repository so that it is not dependent on *libyang*.

Regards,
Michal

mislavn:
@michalvasko I moved the DEB packages into sysrepo.

mislavn:
You are right, that's strange. I'll take a closer look in the next day or two.

mislavn:
I add a new example with C++ callbacks based on [rpc_example.c](https://github.com/sysrepo/sysrepo/blob/master/examples/rpc_example.c) and all callbacks work there.

Also when I edited [oper_data_example.c](https://github.com/sysrepo/sysrepo/blob/master/examples/oper_data_example.c) to use "dummy-amp:amplifier" the callback's didn't get called.

mislavn:
@jktjkt I just tested "dummy-amp:amplifier" on the latest devel branch with the oper_data_example.c example and the callback is called there, so the issue is not connected to C++.

mislavn:
@jktjkt  I fixed the issue with https://github.com/sysrepo/sysrepo/pull/1295/commits/ce141bb185dd0dcb96dd47031433b0ba6a15d057 , please can you verify it.

The issue was that default values were not properly handled with the namespace change, also they made them obsolete, so I removed them.

mislavn:
Python had issues so I still could not use Sysrepo enums but custom ones with https://github.com/sysrepo/sysrepo/pull/1295/commits/177ce27b33efad9e192cc68985a4971864fcdeb6.

mislavn:
@jktjkt please can you verify the PR?

PR #1299, opened 2018-09-10
DO NOT MERGE: Travis: see if my patch to cmocka fixes the CI failures

jktjkt:
A real fix would be to depends on a cmocka release of course, but this is handy to have stuff CI-checked. Sorry for noise.


michalvasko:
Hi,
the tests passed and thank you for looking into it. The patch seems OK to me, but are also the changes in travis install scripts necessary?

Regards,
Michal

jktjkt:
This was meant just as a test for upstream that this patch fixes the root caus in CMocka. It's been merged into their master.

We might want to switch to a release branch to prevent these annoyances in future.

I'll kick the other pull requests now.

PR #1320, opened 2018-10-11
Fixed moved operation display

fxdupont:
### Description
The cpp_application_changes_example.cpp crashes on a moved operation to the first place of the list by using the old value which is null in this case.

Fix is trivial. I took the occasion to reformat the moved case code in the style of other cases. Note if the term "first" is not good enough please replace it by a better one.

### Test case
Create a list with int id and string name leaves. Use id as the key and set the by user (vs system) ordering. Create 2 entries: id=1 name="foo" and id=2 name="bar". Launch cpp_application_changes_example and with sysrepocfg changes the id of the first entry from 1 to 10.

The system will produce some changes: delete the first entry, create a new one with id=10 name="foo" and what will cause the crash move the new entry to the first place in the list.

### Closure
None.

### DIFF
--- a/swig/cpp/examples/cpp_application_changes_example.cpp
+++ b/swig/cpp/examples/cpp_application_changes_example.cpp
@@ -61,8 +61,15 @@ print_change(sysrepo::S_Change change) {
         }
     break;
     case SR_OP_MOVED:
-        if (nullptr != change->new_val()) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
it is just an example, so your fix is fine, thanks.

Regards,
Michal

PR #1354, opened 2018-11-14
add originating xpath to sr_dp_get_items_cb

samuel-gauthier:
When a data request is made, sysrepo will ask to a subscriber for each level under the subscription, calling the subscriber's callback once per layer. When a lot of state nodes are present, it can trigger performance problems in the subscriber.

This patch adds the xpath from the originating get_item request to the sr_dp_get_items_cb prototype. Then, the subscriber can know what it has to answer, and can filter beforehand what it should return to sysrepo, lowering the load needed to get the informations, and the number of requests between sysrepo and the subscriber.

Note that those patches breaks the API, as the sr_dp_get_items_cb prototype has a new original_xpath argument.

### DIFF
--- a/doc/start.dox
+++ b/doc/start.dox
@@ -204,7 +204,7 @@ the specific changed values.
 @subsection state State Data
 ~~~{.c}
 static int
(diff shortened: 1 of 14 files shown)


michalvasko:
Hi,
seems fine, thanks!

Regards,
Michal

PR #1376, opened 2018-12-10
added compiler attribute __format__ to chekc variadic arguments of pr…

palohan:
…intf-like functions

Signed-off-by: Pavol Hanzel <pavol.hanzel@pantheon.tech>

### Description
Compiler will now check the type of variadic arguments to public printf-like functions.
Making the change as agreed with Michal Vasko.

### Test case
No new functionality; This just needs to build properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,19 @@ set(INDIRECT_YANG_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/
 set(CALL_SYSREPOCTL_BIN "${CMAKE_BINARY_DIR}/src/sysrepoctl" CACHE PATH "Path to the sysrepoctl binary to be used at install time")
 set(CALL_SYSREPOCFG_BIN "${CMAKE_BINARY_DIR}/src/sysrepocfg" CACHE PATH "Path to the sysrepocfg binary to be used at install time")
 
+if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
+    set(COMPILER_FORMAT_ATTR "__attribute__ ((__format__(archetype, string_index, first_to_check)))")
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -45,6 +45,19 @@ set(INDIRECT_YANG_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/
...
+    # require at least clang 3.0
+    if (CMAKE_C_COMPILER_VERSION VERSION_LESS 3.0)
+        message(FATAL_ERROR "Clang version must be at least 3.0!")
+    endif()
+    set(COMPILER_FORMAT_ATTR "__attribute__ ((__format__(archetype, string_index, first_to_check)))")
+else()
+    message(WARNING "You are using an unknown compiler, it must support C11 standard \"_Generic\" statement or the \"__builtin_types_compatible_p\" built-in function.")
+    set(COMPILER_FORMAT_ATTR "")
+endif()
+
```
michalvasko: Even though I said it should be something like in *libyang*, I did not mean exactly the same. I wanted you to check what compilers support this attribute and define it appropriately. But, what I imagined this will contain was simply a check whether you are using `gcc` or `clang`, in which cases the attribute is defined and it is not in any other case. There is no need for `clang` version check or any warnings.

michalvasko:
Hi,
seems fine, just please address the comment and it should be ready for merging, thanks.

Regards,
Michal

CMakeLists.txt
palohan: thanks for the suggestion, done

michalvasko: Um, if you performed the change I do not see here any new commits, did you do everything correctly?

palohan: Yes, I forgot to add the file, many apologies...it should be there now

michalvasko:
Alright, now seems OK.

Regards,
Michal

PR #1382, opened 2018-12-13
install common.h i.o. to make them visible for libsysrepo users

frimpler:
Most probably since commit 00344253 (introduction of "format checking") it is not possible to build netopeer2-server.

Reason:
The newly introduced and generated file common.h has not been installed and libsysrepo users, like netopeer2-server, fail to include (the correct) common.h. In fact, a "common.h" was apparently found, because gcc didn't complain about it, but it didn't contain the definition for "FORMAT".
Sysrepo compilation itself is not affected, because common.h has been generated into sysrepo's build directory, too.

Consideration:
I have cared for, that it would be installed in the same manner as other header files for sysrepo. And netopeer2-server compilation is working, again, but the (default) install location "/usr/include" and the file name "common.h" might quickly become ambiguous. Maybe @palohan, the author, has an opinion about it.


michalvasko:
Hi,
it seems I have not thought through having a public `common.h`. I would fix it by moving the `@` variable directly to `sysrepo.h` and generate it from `sysrepo.h.in` so that no additional file is needed. I will implement it tomorrow if none will have anything against.

Regards,
Michal

frimpler:
I don't really care, as long as we can compile netopeer2-server w/o issues :-)

michalvasko:
Hi,
it should work fine now.

Regards,
Michal

PR #1387, opened 2018-12-17
Copy all ns fix

kpbarrett:
### Description
Fix an error in sr_copy_all_ns(), which fails when a list is augmented in another module, and the list key contains a slash character. This causes Netopeer2 to fail to deliver the netconf-config-change notification to clients.

### Test case
The PR contains a commit with the failing case followed by the proposed fix. See new test case "sr_copy_all_ns_test" in tests/common_test.c.


michalvasko:
Hi,
seems good, thanks.

Regards,
Michal

PR #1393, opened 2019-01-02
Application receive information from which netconf client (session-id) specific rpc has been sent.

wmakuch:
### Description
Extended _rpc_cb_ with parameter _netconf_id_. This value is set in Netopeer2-server. In the case of unknown netconf id, this parameter should be 0.

### Test case
Manual tests.

### Closure
Netopeer2 #363

### DIFF
--- a/src/clientlib/cl_subscription_manager.c
+++ b/src/clientlib/cl_subscription_manager.c
@@ -866,7 +866,7 @@ cl_sm_rpc_process(cl_sm_ctx_t *sm_ctx, cl_sm_conn_ctx_t *conn, Sr__Msg *msg)
     sr_node_t *input_tree = NULL, *output_tree = NULL;
     sr_mem_ctx_t *sr_mem_resp = NULL;
     size_t input_cnt = 0, output_cnt = 0;
-    const char *op_name = NULL;
+    const char *op_name = NULL, *netconf_id = NULL;
     bool action = false;
     sr_rpc_cb cb = NULL;
     sr_rpc_tree_cb cb_tree = NULL;
@@ -875,6 +875,8 @@ cl_sm_rpc_process(cl_sm_ctx_t *sm_ctx, cl_sm_conn_ctx_t *conn, Sr__Msg *msg)
     CHECK_NULL_ARG4(sm_ctx, msg, msg->request, msg->request->rpc_req);
 
     action = msg->request->rpc_req->action;
+    netconf_id = msg->request->rpc_req->netconf_id;
+
     op_name = action ? "Action" : "RPC";
     SR_LOG_DBG("Received %s request (%s) for subscription id=%"PRIu32".",
             op_name, msg->request->rpc_req->xpath, msg->request->rpc_req->subscription_id);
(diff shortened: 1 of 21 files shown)


wmakuch:
Hi,
You right. Extending API is not necessary. I prepared new patch with suggested change. 
Soon I will open new pull request.

Best Regard,
Wojciech Makuch

wmakuch:
Hi,

I updated pull request.
After further testing I realised that API must be extended in order to make this feature useful.
Your suggestion was pass `sr_session_ctx_t*` by `private_ctx`, but this parameter is set during `sr_rpc_subscribe` (not during `sr_rpc_send`), so _Netopeer2-server_ has nothing to do with it. I decided to pass` sr_session_ctx_t*` directly.

Kindly please take a look at my changes and share your opinion, comments.

PR to _Netopper2_ repository: https://github.com/CESNET/Netopeer2/pull/376

Best Regards,
Wojciech Makuch

michalvasko:
Hi Wojciech,
you are right, this is basically what I had in mind, I just thought a `session` is passed to all the callbacks. So it generally seems fine except the tests failed for some reason. I am sorry, but it is mostly up to you to fix it, I have no time to spare working on the old *sysrepo* as I am developing the new one.

Regards,
Michal

src/clientlib/cl_subscription_manager.c
```diff
@@ -900,17 +904,25 @@ cl_sm_rpc_process(cl_sm_ctx_t *sm_ctx, cl_sm_conn_ctx_t *conn, Sr__Msg *msg)
         goto cleanup;
     }
 
+    if (NULL != subscription->private_ctx) {
+        memcpy(session, subscription->private_ctx, sizeof(sr_session_ctx_t));
+    }
```
michalvasko: I am totally confused about this `if`, what is going on here? Isn't it just some leftover from your testing?

wmakuch: Not exactly. Memcpy fails, when we pass null pointer. _Sysrepo_ tests crash/not pass, when there is no check if private_ctx is null. Maybe more elegant is use just
`if (subscription->private_ctx) {` ?

michalvasko: But I did not find in any other changes you made that you would set the `private_ctx` anywhere. Also, if you do, then you should not because it is meant to be used by the user, not internally.

wmakuch: the `netconf_id` value is passed by np2srv_sr_rpc_send according to https://github.com/CESNET/Netopeer2/pull/376. User must pass session as private_ctx to `sr_rpc_subscribe`.

michalvasko: In the *netopeer2-server* pull request it seems fine but I am no closer to understanding of these changes than I was before. Maybe it is my fault but my general idea would be that you set the NETCONF SID to a *sysrepo* session you are using to send the RPC (for example). It will go through the whole *sysrepo* and appear in the callback session *sysrepo* provides for the RPC subscriber. There it can be read from this session, that is what I think you want. All your changes fit into this idea, I think, except for this one `if` clause.

Why would then the `private_ctx` of `sr_rpc_subscribe` be needed to be set to anything by the user?

wmakuch:
Hi,
Thanks for good tips!
Review updated.

PR #1394, opened 2019-01-02
Use atomic variable for thread-shared counter

### LINKED ISSUE #1392: Call sr_get_items() in high frequency executes failed
### Things you may try first
(put "x" in "[]" if you already tried following)
* [] Did you test it on the latest sysrepo devel branch?

### Description
1. Use example turing-machine.yang and get `/turing-machine:turing-machine/state` state data
2. Write daemon process and subscribe `sr_dp_get_items_subscribe()` `/turing-machine:turing-machine/state`
3. Write client process, `sr_session_refresh()` and `sr_get_items()`
4. Repeat executes step 3 in with interval time 10ms

### Steps to Reproduce
1. `sudo sysrepod -d`
2. `./op_data_example -d`
3. `./op_data_example -c` (open 7 windows and run 7 clients to execute the client example at same time )

**Actual behavior:**
sysrepo run abnormal and the log is shown below:
```
subscription xpath: /turing-machine:turing-machine
subscription xpath: /turing-machine:turing-machine
subscription xpath: /turing-machine:turing-machine
subscription xpath: 8
[ERR] Namespace copy failed
[ERR] Node validation failed for xpath 8
[ERR] Subscriptions to schema nodes failed
[ERR] rp_dt_xpath_requests_state_data failed
[ERR] rp_dt_prepare_data failed
[ERR] Get items failed for '/turing-machine:turing-machine/state', session id=1303975091.
subscription xpath: 8
[ERR] Namespace copy failed
[ERR] Node validation failed for xpath 8
[ERR] Subscriptions to schema nodes failed
[ERR] rp_dt_xpath_requests_state_data failed
[ERR] rp_dt_prepare_data failed
[ERR] Get items failed for '/turing-machine:turing-machine/state', session id=523203734.
```
It seems that the cache data in `md->cached_data->data` has been broken and the xpath is abnormal.

Most important:
when I remove `sr_session_refresh()` before `sr_get_items()`,  the test examples run normal.

### Versions
```
$ git branch
* master
$ sysrepod -v
sysrepod - sysrepo daemon, version 0.7.6 
```

### Attachments
[op_data_example.zip](https://github.com/sysrepo/sysrepo/files/2719890/op_data_example.zip)

Thanks!

### Solutions
1. Add check function in `pm_get_cached_subscriptions()` as shown below. When cache xpath check error, it will read from file.
```
 /**
  * @brief Return cached subscriptions of requested module and type, if present in the cache.
  */
@@ -695,6 +719,10 @@ pm_get_cached_subscriptions(pm_ctx_t *pm_ctx, const char *module_name, Sr__Subsc
 
         for (size_t i = 0; i < cd->subscriptions->count; i++) {
             subscription = cd->subscriptions->data[i];
+            /* chuny add start for pm-cache */
+            rc = __check_xpath(subscription);
+            CHECK_RC_MSG_GOTO(rc, cleanup, "Cache xpath check error.");
+            /* chuny add end for pm-cache   */
 
             rc = sr_list_add(res_subscriptions, subscription);
             CHECK_RC_MSG_GOTO(rc, cleanup, "Unable to add a subscription into the subscription list.");
```

```
+/* when muti clients(6+) use get_items to get provider data, the cache data will be abnormal. */
+/* The subscription xpath is modified to special charsomehow, we did not find the root cause. */
+/* So we add the checking mechanism, when cache data is abnormal, we read it from data file.  */
+/* chuny add start for pm-cache */
+static int
+__check_xpath(np_subscription_t *pstSubscription)
+{
+    char *pcColon = NULL;
+
+    if (NULL == pstSubscription) {
+        return SR_ERR_INVAL_ARG;
+    }
+
+    if (NULL != pstSubscription->xpath) {
+        pcColon = strchr(pstSubscription->xpath, ':');
+        if ((pstSubscription->xpath[0] != '/') || (NULL == pcColon)) {
+            return SR_ERR_INTERNAL;
+        }
+    }
+
+    return SR_ERR_OK;
+}
+/* chuny add end for pm-cache */
+
```
2. `persistence_manager.c` file

[persistence_manager.zip](https://github.com/sysrepo/sysrepo/files/2719897/persistence_manager.zip)

michalvasko:
### Description
To prevent read-increase-write integer data races use an atomic variable.

### Closure
Fixes #1392


michalvasko:
Hi,
in case the header is not available it is not reverted back to the old behavior but compiler macros are used instead of C11-standardized atomic variables. They should provide the same functionality but are probably a bit slower and perhaps not available on all compilers? I don't know, if there is someone with such a compiler, we will fix it. But in that case they will certainly get a compile-time error.

Regards,
Michal

PR #1400, opened 2019-01-11
Add support for Vals/Vals_Holder reallocation using sr_realloc_values

kpbarrett:
### Description
Enhancement to C++ API. Original author: gunter@vcinity.io

### Test case
Please see example: cpp_turing_rpc_example.cpp


mislavn:
It looks correct, I'll test it next Monday.

Issue #1396, opened 2019-01-07
sysrepod crash with SIGBUS inside dm_load_data_tree


przemekr:
Hi,

From time to time (i.e 1/100 run of a test case) I am getting a SIGBUS crash inside sysrepod. 

sysrepod - sysrepo daemon, version 0.7.7
yanglint 0.16.99

(gdb) bt
#0  __memcpy_avx_unaligned () at ../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:163
#1  0x00007fb66a5a9e94 in memcpy (__len=1, __src=0x7fb66bc7f000, __dest=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/string3.h:53
#2  lyb_read (data=data@entry=0x7fb66bc7f000 <error: Cannot access memory at address 0x7fb66bc7f000>, buf=<optimized out>, buf@entry=0x7fb668dd67c7 "", count=count@entry=1, lybs=lybs@entry=0x7fb668dd67d0)
    at libyang/libyang-v0.16.99/src/parser_lyb.c:69
#3  0x00007fb66a5ac0e3 in lyb_parse_magic_number (lybs=0x7fb668dd67d0, data=0x7fb66bc7f000 <error: Cannot access memory at address 0x7fb66bc7f000>) at libyang/libyang-v0.16.99/src/parser_lyb.c:1153
#4  lyd_parse_lyb (ctx=ctx@entry=0x7fb660000e30, data=data@entry=0x7fb66bc7f000 <error: Cannot access memory at address 0x7fb66bc7f000>, options=options@entry=8705, data_tree=data_tree@entry=0x0, yang_data_name=yang_data_name@entry=0x0, parsed=parsed@entry=0x0)
    at libyang/libyang-v0.16.99/src/parser_lyb.c:1216
#5  0x00007fb66a5d8a3f in lyd_parse_ (yang_data_name=0x0, data_tree=0x0, options=8705, format=LYD_LYB, data=0x7fb66bc7f000 <error: Cannot access memory at address 0x7fb66bc7f000>, rpc_act=0x0, ctx=0x7fb660000e30)
    at libyang/libyang-v0.16.99/src/tree_data.c:961
[...]

PR #1403, opened 2019-01-16
Use LINUX_OFD_LOCK file locks when available.

przemekr:
fix issue #1396.

F_OFD_SETLK works as F_SETLK, only it is associated with file descriptor
rather then the process id, so concurrent access from two different threads
can be prevented (see fcntl man pages).

### Closure
#1396

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,6 +140,7 @@ set(CMAKE_REQUIRED_LIBRARIES pthread)
 include(CheckFunctionExists)
 include(CheckIncludeFiles)
 include(CheckStructHasMember)
+include(CheckSymbolExists)
 CHECK_FUNCTION_EXISTS(pthread_rwlockattr_setkind_np HAVE_PTHREAD_RWLOCKATTR_SETKIND_NP)
 CHECK_FUNCTION_EXISTS(getpeereid HAVE_GETPEEREID)
 CHECK_FUNCTION_EXISTS(getpeerucred HAVE_GETPEERUCRED)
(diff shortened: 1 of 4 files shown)


src/common/sr_utils.c
```diff
@@ -20,6 +20,15 @@
  * limitations under the License.
  */
 
+#include "sr_constants.h"
+#ifdef HAVE_LINUX_OFD_LOCK
+#define _GNU_SOURCE
+#else
+//! @cond doxygen_suppress
+#define __USE_XOPEN
```
michalvasko: I suppose `_GNU_SOURCE` is required by the OFD locks but `__USE_XOPEN` is always required by `strptime()`, why do you not define it if OFD locks are used? That is why the tests failed. If it causes some problems, you can try defining `_XOPEN_SOURCE` instead, that is what `man` pages say one should do.

CMakeLists.txt
```diff
@@ -152,6 +153,7 @@ if(HAVE_MKSTEMPS)
     set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -DHAVE_MKSTEMPS")
 endif(HAVE_MKSTEMPS)
 CHECK_STRUCT_HAS_MEMBER("struct stat" st_mtim "sys/stat.h" HAVE_STAT_ST_MTIM)
+CHECK_C_SOURCE_COMPILES("#define _GNU_SOURCE\n#include <fcntl.h>\nint main(int argc, char* argv[]) {return F_OFD_SETLKW;}" HAVE_LINUX_OFD_LOCK)
```
michalvasko: I thought you would use simple `CHECK_SYMBOL_EXISTS` on `F_OFD_SETLKW`. This is a legacy way of doing things from `autotools`, inefficient.

src/common/sr_utils.c
przemekr: yes, agree, when I view the headers it seems to me that _GNU_SOURCE will imply __USE_XOPEN, but that of course might be only true for my version of the compiler.  I will update to #define _XOPEN_SOURCE

CMakeLists.txt
przemekr: I thought CHECK_SYMBOL_EXISTS will not detect macro definition, but now I see from the cmake doc page:
"""
If the header files define the symbol as a macro it is considered available and assumed to work
"""

so yes will fix it! :)

PR #1405, opened 2019-01-17
Use LINUX_OFD_LOCK file locks when available.

przemekr:
### Description
F_OFD_SETLK works as F_SETLK, only it is associated with file descriptor
rather then the process id, so concurrent access from two different threads
can be prevented.

### Closure
fix issue #1396.


michalvasko:
Alright, now everything should be fine, thanks.

Regards,
Michal

PR #1406, opened 2019-01-18
Rpc nullptr crash fix

gunterw:
### Description
Data Manager nullptr crash

### Test case
Performing an rpc with '/ietf-system:set-current-datetime/current-datetime = 2019-01-11T11:43:00Z'
would cause a nullptr crash in the data manager.


mislavn:
Hi @gunterw ,

please use the devel branch for the pull request.

PR #1410, opened 2019-01-21
Cpp reallocation2

gunterw:
### Description
Improvements for cpp allocation/reallocation

### Test case
1. Calling Vals_Holder::reallocate before Vals_Holder::allocate causes an exception to be thrown.

2. Calling Vals_Holder::allocate(0) followed by Vals_Holder::allocate(n) where n>0 causes an exception to  be thrown.

This change performs an allocation instead of throwing an exception for both of these cases.

### DIFF
--- a/swig/cpp/src/Internal.cpp
+++ b/swig/cpp/src/Internal.cpp
@@ -70,6 +70,8 @@ Deleter::Deleter(sr_session_ctx_t *sess) {
 }
 Deleter::~Deleter() {
     switch(_t) {
+    case Free_Type::INVALID:
+    break;
     case Free_Type::VAL:
         if (v._val) sr_free_val(v._val);
     v._val = nullptr;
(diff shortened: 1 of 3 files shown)


gunterw:
Sorry again... not sure what is going on.. 
I just started using SmartGit to push to github and it is obviously not doing what I expect.
I will switch to good old command line...

gunterw:
FYI - I'm seeing some asserts using the new cpp reallocation method in my local testing

sr_mem_mgmt.c:337
` 
/* it must have been found, otherwise the input was invalid */
assert(node_ll && i < MAX_BLOCKS_AVAIL_FOR_ALLOC-1);
`
When/If I get this figured out I will create a new pull request.

gunterw:
No longer seeing the assert problem with https://github.com/sysrepo/sysrepo/issues/1413 fix so I re-opened this pull request. I also added a new commit to handle Deleter objects properly

PR #1411, opened 2019-01-21
Fix null ptr crash that occurred when validating

gunterw:
### Description
Fix null ptr crash that occurred when validating 
rpc '/ietf-system:set-current-datetime/current-datetime = 2019-01-11T11:43:00Z'


michalvasko:
Hi,
I was just about to verify this but @jktjkt was faster so fine, thanks for the fix.

Regards,
Michal

PR #1412, opened 2019-01-23
Added a test case for the Vals class which shows

gunterw:
### Description
Added a test case for the Vals class which shows an error in sr_realloc

### Test case
The test case improperly fails with this assert 

> sr_realloc: Assertion `node_ll && i < MAX_BLOCKS_AVAIL_FOR_ALLOC-1' failed.


michalvasko:
Hi,
issue is fixed and so I can merge this, thanks.

Regards,
Michal

PR #1424, opened 2019-02-07
Cpp reallocation

gunterw:
### Description
Need to add a null ptr check to recent reallocate changes, Vals objects can
be created with a nullptr deleter object (see Session.cpp:473 rpc_cb)

Also change the order of member initialization so Vals_Holder object is unmodified on reallocate failure.

### DIFF
--- a/swig/cpp/src/Struct.cpp
+++ b/swig/cpp/src/Struct.cpp
@@ -436,7 +436,8 @@ sr_val_t* Vals::reallocate(size_t n) {
     if (ret != SR_ERR_OK)
         throw_exception(ret);
     _cnt = n;
-    _deleter->update_vals_with_count(_vals, _cnt);
+    if (_deleter)
+        _deleter->update_vals_with_count(_vals, _cnt);
     return _vals;
 }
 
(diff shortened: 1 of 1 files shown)


gunterw:
The Vals::Vals constructor defaults the delete pointer to null, there are many places where the Vals deleter will be null (maybe you are confusing the Vals delete ptr with the Val delete ptr??)

Here's an example of creating a Vals object with a nullptr, you can also look in Session.cpp for all the callbacks and a nullptr is passed for the deleter object.

```S_Vals Vals_Holder::allocate(size_t n) {
    if (_allocate == false)
        throw_exception(SR_ERR_DATA_EXISTS);

    if (n == 0)
        return nullptr;

    int ret = sr_new_values(n, p_vals);
    if (ret != SR_ERR_OK)
        throw_exception(ret);

    *p_cnt = n;
    _allocate = false;

    p_Vals = S_Vals(new Vals(p_vals, p_cnt, nullptr));
    return p_Vals;
}
```

jktjkt:
Ah, right, so this is for a use case where the user constructs ``Vals`` on their own. I've only tested with values obtained from within sysrepo in various callbacks where that `nullptr` default never happened in my testing. Thanks!

PR #1443, opened 2019-03-12
Move test timeouts to CMake's native features

jktjkt:
##  tests: Use CMake's native timeout 
The test suite used two non-standard mechanisms for setting a timeout for each test:

- if the `timeout` binary was available, all tests were wrapped within that and with a timeout of six minutes,
- a custom watchdog thread was running, killing the process after a configurable timeout which was five minutes for all tests except nacm_cl_test where it was a bit more than eight hours

Given that CMake has a native feature for this, let's use that one and remove both custom methods.

The timeout for `nacm_cl_test` was set to 30'000 seconds (that's about a working day). That looks like a nonsensical value, so let's set it to 10 minutes and hope for the best.

## tests: Allow perf_test a bit longer 
When running within -fsanitize=thread, it typically takes 288 secs
within my CI environment *when it succeeds*. It also occasionally fails,
and when it fails, it is due to that five-minute timeout.


jktjkt:
The test failure within Travis is due to https://github.com/CESNET/libyang/issues/730

michalvasko:
Yes, the changes themselves look fairly simple and make the tests simpler as well, thanks!

Regards,
Michal

PR #1446, opened 2019-03-18
Install the pkg-config file unconditionally

jktjkt:
The C++ bindings install their own `.pc` without checking whether the pkg-config helper is available on the build system. That's not necessarily a wrong thing to do; the `libsysrepo.pc` is a tiny file whose presence definitely doesn't hurt.

I came here because a colleague ended up with an installation that contained just `libSysrepo-cpp.pc` which referred to `libsysrepo.pc` that was not present.

Cc: @obonovai

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,12 +328,9 @@ if(ENABLE_NACM)
     INSTALL_YANG("ietf-netconf-acm" "@2018-02-14" "644")
 endif(ENABLE_NACM)
 
-find_package(PkgConfig QUIET)
-if(PKG_CONFIG_FOUND)
-    # generate and install pkg-config file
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
why not do it the other way around and condition installing C++ `pkg-config` file?

Regards,
Michal

jktjkt:
Because I think that there's no value in skipping that installation if `pkg-config` is not available. If it is not available and if the user tries to use some SW which requires `pkg-config` later on, it's easy to install it. The current form requires reconfiguring sysrepo (and libyang). I just walked a student through this process, it only wasted five minutes of our time, but it's a bit annoying.

Anyway, I don't really care :). Please feel free to reject this patch if you think that not installing that tiny file is a better solution for some reason.

michalvasko:
Hi,
fair enough, merging.

Regards,
Michal

Issue #1450, opened 2019-03-25
TSAN: data race in notifications with custom event loop


jktjkt:
We have a custom event loop implementation that is based on libev. A test case which sends notifications from a background thread ends up with the following report from TSAN:
```
2019-03-25 09:36:24.392001 | w | WARNING: ThreadSanitizer: data race (pid=917)
2019-03-25 09:36:24.392098 | w |   Write of size 4 at 0x7b1800003470 by thread T12:
2019-03-25 09:36:24.392340 | w |     #0 sr_msg_free /home/ci/src/cesnet-gerrit-czechlight/CzechLight/dependencies/sysrepo/src/common/sr_mem_mgmt.c:595 (libsysrepo.so.0.7+0x998c0)
2019-03-25 09:36:24.392573 | w |     #1 cm_out_msg_process /home/ci/src/cesnet-gerrit-czechlight/CzechLight/dependencies/sysrepo/src/connection_manager.c:1756 (libsysrepo.so.0.7+0xcb519)
2019-03-25 09:36:24.392830 | w |     #2 cm_msg_enqueue_cb /home/ci/src/cesnet-gerrit-czechlight/CzechLight/dependencies/sysrepo/src/connection_manager.c:1819 (libsysrepo.so.0.7+0xb34c6)
2019-03-25 09:36:24.392929 | w |     #3 ev_invoke_pending ??:? (libev.so.4+0x4f72)
2019-03-25 09:36:24.393172 | w |     #4 cm_event_loop_threaded /home/ci/src/cesnet-gerrit-czechlight/CzechLight/dependencies/sysrepo/src/connection_manager.c:1891 (libsysrepo.so.0.7+0xb41cb)
2019-03-25 09:36:24.393208 | w |
2019-03-25 09:36:24.393333 | w |   Previous write of size 4 at 0x7b1800003470 by thread T8:
2019-03-25 09:36:24.393555 | w |     #0 sr_msg_free /home/ci/src/cesnet-gerrit-czechlight/CzechLight/dependencies/sysrepo/src/common/sr_mem_mgmt.c:595 (libsysrepo.so.0.7+0x998c0)
2019-03-25 09:36:24.393805 | w |     #1 rp_msg_dispatch /home/ci/src/cesnet-gerrit-czechlight/CzechLight/dependencies/sysrepo/src/request_processor.c:3687 (libsysrepo.so.0.7+0xdec92)
[...]

PR #1451, opened 2019-03-26
Use atomic for message ref counter

michalvasko:
### Description
Atomic variable should be used for message reference counter to avoid data races.

### Closure
Refs #1450


jktjkt:
I'll give this a try, but it might not be today. Thanks for looking into this!

PR #1504, opened 2019-05-27
Ensure that feature test macros work correctly

jktjkt:
These feature test macros are [required to be defined prior to including any standard headers](http://man7.org/linux/man-pages/man7/feature_test_macros.7.html). This is a transitive requirement because headers are, in general, free to include other headers. The sysrepo's code violated this because it included `<stdatomic.h>`.

This fixes a build failure when using clang on Fedora 29. Apparently, the order of `#include` is different from using GCC. By the time `sr_utils.c` defined `_GNU_SOURCE`, the `<features.h>` was already processed, and therefore `_GNU_SOURCE` was not turned into `__USE_GNU` (which is what `<bits/fcntl-linux.h>` uses as a guard for the `F_OFD_GETLK` etc.


jktjkt:
Can I get this merged, please? It's currently blocking our upgrades past 89cfe683 internally.

Issue #1042, opened 2018-01-30
Updating running DB 


rosy8:
Hi Michal,

We can set/update the running db directly  by using the sr_set_item  right ?
I donot think  we need to  execute sr_commit() is  explicitly post sr_set_item(..) call for the data to be persistent on DB.

Please correct me if I am wrong.

Regards,
Rosy

michalvasko:
Hi Rosy,
you need to execute `sr_commit()` for the data to be persistent and visible for other sessions.

Regards,
Michal

rosy8:
Hi Michal,

Thanks for your response.
 Is there any other API which I can use for this purpose ?
 Which will  copy already validated/verified data in to sysrepo running DB and it will be persistent .
 If I do sr_commit, I will get SR_EV_VERIFY/SR_EV_APPLY  which I want to skip .

Regards,
Rosy

michalvasko:
Hi Rosy,
you should be able to do this if you perform those additional changes in a separate thread.

Regards,
Michal

rosy8:
Hi Michal,

edit-config - commit
SR_EV_VERIFY
-- Everything OK
SR_EV_APPLY
====>> Here I am adding data manipulation of the Sysrepo running DB by using sr_set_item  and have skipped sr_commit .
====>>
return Ok (Response  for SR_EV_APPLY event processing)

And for this I could see data is persistent between  different sessions .
I am talking about this specific scenario..

Please let me know your feedback on this.

Regards,
Rosy

michalvasko:
Hi Rosy,
I cannot tell you for sure whether it is supported and will not cause some problems but I would try what I have suggested, when processing `SR_EV_APPLY` I would create another thread and do `sr_set_item` and `sr_commit`. Not the most elegant way, but may work.

Regards,
Michal

rosy8:
Hi Michal,

Thanks for responding back.
 I create a new thread and perform the sr_set_items/sr_commit on the new thread.
Then after sr_commit, it will generate SR_EV_VERIFY/SR_EV_APPLY events right ?
So how sysrepo is going to handle that ?? 
As per my knowledge  sysrepo data manipulation are blocking call.
So sysrepo is not going to process the new SR_EV_VERIFY unless it gets the response back for the earlier
event (SR_EV_APPLY from the main thread ). Also here only one connection is there.

But i donot want that to happen..I want the data manipulation to be done when the main thread returns the response for SR_EV_APPLY.

[...]

PR #1517, opened 2019-06-17
Don't send notifications back to the session that changes database

mbennett2212:
If a single threaded process is subscribed to validate the database and
then commits to the database it will deadlock/timeout as the validate
callback can never be run. In this case it could be assumed that the
process has already validated this data before committing it to the
database so it would be useful if it was possible for this functionality
to work without deadlocking. To achieve this we now store the session id
that subscribed to validate the database and if this session id matches
the id of the session that has committed to the database then we skip
the subscription notification - avoiding the deadlock.

Relates to #1042


mbennett2212:
This is useful for my use case and I notice it may also be related to #1042 . Please let me know if this is useful for upstream and if so how best to get it accepted (perhaps this should be configurable as part of the subscribe, or a new function for committing is required).

michalvasko:
Hi,
what exactly are you doing in the validation callback? Generally, it should not block unless you are calling some *sysrepo* functions.

So, I do not think this is, in general, such a good idea, it seems too specific. Why should the value not be validated just because it was written using the same session as was used for the subscription? 

Regards,
Michal

mbennett2212:
Hi Michal,

The process writing to the paths in the database is also subscribed to validate those paths. The process receives the sysrepo callbacks in the same thread as the thread that writes to the database (I followed the example in application_fd_watcher_example.c). The write to the database blocks until it can be validated but since the validation can never run (since it has to run in the same thread) the request times out.

In my system this process is subscribed to validate these paths for requests that come through netconf, gnmi or other processes writing to sysrepo. However it may receive changes 'out of band' (that do not come through sysrepo) so it needs to update sysrepo with these changes as they occur. It is not possible for us to change these 'out of band' changes to instead go through sysrepo, and writing these changes to sysrepo from a separate thread in our process introduces race/timing problems when other changes are also going to sysrepo from netconf/gnmi/other-processes.

I don't think this should be the normal sysrepo behaviour (don't validate if the subscriber is doing the write), but we definately has a use case for this behaviour so it seems like it would be nice for it to be optional behaviour that can be used.

Thanks,
Matt

michalvasko:
Hi Matt,
okay, I understand, but this pull request does not make this behavior optional and the condition of the same session sending the changes and subscribing to them seems a bit arbitrary to me.

The whole problem is having the configuration in 2 places, it is not supported by *sysrepo*. However, we do have some (vague) plans to add this support with a custom datastore feature although not in the near future.

Best that you could do is to somehow use only the *sysrepo* datastore but that would involve a lot of work, I imagine. Other than that, you can create the separate thread and solve the data races that could occur.

Regards,
Michal

PR #1584, opened 2019-08-27
rp_dt_get_subtrees_wrapper return value

### LINKED ISSUE #1581: Func "rp_dt_get_subtrees_wrapper" Error Return Value
#659 # Description
[Description of the bug or feature]
when i used "get_subtrees" to get list,if the list is null,the sysrepo will report a err msg,but when i used "get_items",there is no err msg.

in the func "rp_dt_get_subtrees_wrapper", when func "rp_dt_get_subtrees" return "SR_ERR_NOT_FOUND", which is not used outside:
```C
    rc = rp_dt_get_subtrees(rp_ctx->dm_ctx, rp_session, data_tree, sr_mem, xpath,
                            dm_is_running_ds_session(rp_session->dm_session), subtrees, count);
    if (**SR_ERR_UNAUTHORIZED** == rc) {
        rc = SR_ERR_NOT_FOUND;
    } else if **(SR_ERR_OK != rc**) {
        SR_LOG_ERR("Get subtrees failed for xpath '%s'", xpath);
    } 
```

Thanks!

shubengang:
Use default coding style defined at CODINGSTYLE.md

### Description
when i used func "get_subtrees" to get list,if the list is null,log will printf a err msg.

### Test case
when the func  ‘rp_dt_get_subtrees’ return a value(rc == SR_ERR_NOT_FOUND),we should use it, and not go into the last ‘else if’,so than will not pintf err log.

### Closure
Put closes #1581 in your comment to auto-close the issue that pull request fixes.

### DIFF
--- a/src/rp_dt_get.c
+++ b/src/rp_dt_get.c
@@ -1492,7 +1492,7 @@ rp_dt_get_subtrees_wrapper(rp_ctx_t *rp_ctx, rp_session_t *rp_session, sr_mem_ct
 
     rc = rp_dt_get_subtrees(rp_ctx->dm_ctx, rp_session, data_tree, sr_mem, xpath,
                             dm_is_running_ds_session(rp_session->dm_session), subtrees, count);
-    if (SR_ERR_UNAUTHORIZED == rc) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
okay, no problem, merging.

Regards,
Michal

PR #1629, opened 2019-10-04
cl_subscription_manager: Remove usleep

neheb:
usleep was removed from POSIX 2008 and is optionally unavailable with
uClibc-ng.


michalvasko:
Hi,
okay, no problem, thanks.

Regards,
Michal

PR #1648, opened 2019-10-22
Split the sr_error_info_t struct

pieterlexis:
### Description
While working on a Golang wrapper for sysrepo, I noticed that the `github.com/xlab/c-for-go` tool does not handle struct definitions inside struct definitions. Although a deficiency in that tool, this change does not affect sysrepo usage.


michalvasko:
Hi,
no problem, merging. And thanks for working on the wrapper!

Regards,
Michal

PR #1664, opened 2019-11-05
Fix Memory leak in python SWIG callback (dp_get_items etc) (#1657)

### LINKED ISSUE #1657: Memory leak in python SWIG callback (dp_get_items etc)
### Things you may try first
(put "x" in "[]" if you already tried following)
* [] Did you test it on the latest sysrepo devel branch?

### Description
Python process that subscribes dp_get_items will slowly increase memory use as callbacks are triggered

### Steps to Reproduce
1. Subscribe to dp_get_items in one python process
2. Call get_item in another process
3. Monitor memory use in first process

**Expected behavior:**
 Memory use remains constants

**Actual behavior:**
 Memory use slowly increases

### Versions
Sysrepo version: 0.7.7

### Analysis
In swig interface file sysrepo.i method
```
int dp_get_items(const char *xpath, sr_val_t **values, size_t *values_cnt, uint64_t request_id, const char *original_xpath, PyObject *private_ctx)
```
calls SWIG_NewPointerObj with last parameter SWIG_POINTER_DISOWN
but SWIG_POINTER_DISOWN is a constant associated with pointer conversions not with pointer create
```
/* Flags for pointer conversions */
#define SWIG_POINTER_DISOWN        0x1
#define SWIG_CAST_NEW_MEMORY       0x2

/* Flags for new pointer objects */
#define SWIG_POINTER_OWN           0x1
```
since the value of the last parameter is 0x1, this means that SWIG_NewPointerObj is actually being called with SWIG_POINTER_OWN?

This means it is necessary to decrement the object reference after the python callback has returned?
```
Py_DECREF(arglist);
Py_DECREF(out); //object no longer needed?
```
Similarly for the other callback methods

### Attachments (suggested patch)

[sysrepo-python-cbmemleak.txt](https://github.com/sysrepo/sysrepo/files/3797839/sysrepo-python-cbmemleak.txt)

maggie450:
### Description
Fix Memory leak in python SWIG callback (dp_get_items etc) (#1657)
Confusion arose from last argument in SWIG_NewPointerObj which referenced a pointer-conversion #define
(#define SWIG_POINTER_DISOWN 0x1)
but was being interpreted as a new pointer object #define
(#define SWIG_POINTER_OWN 0x1)
This means it is necessary to decrement the reference after finish using the new object.
Added the DECREF and tidied the logic a little

### Test case
Memory use of python process that subscribed callback does not increase as get_item requests are made.

### Closure
Closes #1657

### DIFF
--- a/swig/python/sysrepo.i
+++ b/swig/python/sysrepo.i
@@ -74,16 +74,16 @@ public:
 
         sysrepo::Session *sess = (sysrepo::Session *)new sysrepo::Session(session);
         std::shared_ptr<sysrepo::Session> *shared_sess = sess ? new std::shared_ptr<sysrepo::Session>(sess) : 0;
-        PyObject *s = SWIG_NewPointerObj(SWIG_as_voidptr(shared_sess), SWIGTYPE_p_std__shared_ptrT_sysrepo__Session_t, SWIG_POINTER_DISOWN);
+        PyObject *s = SWIG_NewPointerObj(SWIG_as_voidptr(shared_sess), SWIGTYPE_p_std__shared_ptrT_sysrepo__Session_t, SWIG_POINTER_OWN);
 
         arglist = Py_BuildValue("(OsiO)", s, module_name, event, private_ctx);
         PyObject *result = PyEval_CallObject(_callback, arglist);
         Py_DECREF(arglist);
+        sess->~Session();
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
thanks, merging.

Regards,
Michal

PR #1672, opened 2019-11-11
sub shm CHANGE check errno after select only when ret = -1

barbarost:


michalvasko:
Hi,
you are right, merging.

Regards,
Michal

PR #1687, opened 2019-11-27
Sysrepo SWIG Python 3 bindings

lpaulic:
### Description
Add SWIG Python 3 binding including updated tests.


CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 if(NOT UNIX)
     message(FATAL_ERROR "Only Unix-like systems are supported.")
 endif()
+#Language bindings
+set(GEN_LANGUAGE_BINDINGS "Enable python bindings." ON) 
+set(GEN_PYTHON_BINDINGS "Enable python bindings." ON) 
+set(GEN_PYTHON_VERSION "3" CACHE STRING "Python version")
```
michalvasko: I believe Python 2 will no longer be supported, right @rkrejci ? So this option and any references should be removed.

CMakeLists.txt
```diff
@@ -239,6 +242,9 @@ if(GEN_CPP_BINDINGS)
         add_subdirectory(swig)
     endif()
 endif()
+if(GEN_LANGUAGE_BINDINGS)
+    add_subdirectory(swig)
```
michalvasko: This option should be global and duplicating code avoided. But I can fix this myself.

swig/python/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+if(${CMAKE_VERSION} VERSION_LESS "3.8.0")
+    swig_add_module(${PYTHON_SWIG_BINDING} python ${PYTHON_SWIG_BINDING}.i)
+else()
+    swig_add_library(${PYTHON_SWIG_BINDING} LANGUAGE python SOURCES ${PYTHON_SWIG_BINDING}.i)
+endif()
+swig_link_libraries(${PYTHON_SWIG_BINDING} ${PYTHON_LIBRARIES} Sysrepo-cpp)
+
+file(COPY "examples" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+file(COPY "tests" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```
michalvasko: Is this necessary, can't examples and tests be directly compiled at the right place?

swig/python/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+execute_process(COMMAND
+    ${PYTHON_EXECUTABLE} -c
+    "from distutils.sysconfig import get_python_lib; print(get_python_lib())"
+OUTPUT_VARIABLE PYTHON_MODULE_PATH OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+install( FILES "${CMAKE_CURRENT_BINARY_DIR}/_${PYTHON_SWIG_BINDING}.so" DESTINATION ${PYTHON_MODULE_PATH} )
+install( FILES "${CMAKE_CURRENT_BINARY_DIR}/${PYTHON_SWIG_BINDING}.py" DESTINATION ${PYTHON_MODULE_PATH} )
+
+if("${TEST_REPOSITORY_LOC}" STREQUAL "${REPOSITORY_LOC}" AND ENABLE_TESTS)
```
michalvasko: Please avoid this magic with repository locations from the old *sysrepo*. If the tests are enabled they should be possible to run no matter where the repository is, they can properly set and clear their modules and data.

CMakeLists.txt
borna-blazevic: I'm not really sure what you mean by that if you can elaborate I can do it, but it's fine if you do it.

swig/python/CMakeLists.txt
borna-blazevic: Yes, sysrepo.py and _sysrepo.so files generated by swig are installed globally so they can be run from anywhere.

swig/python/CMakeLists.txt
borna-blazevic: I just removed these lines of code since the tests and examples can be run from anywhere.

CMakeLists.txt
michalvasko: So we can keep it for now, I guess. But the hand-written bindings will be only for Python 3.

PR #1700, opened 2019-12-09
Fix python bindings tests, make them install required modules

borna-blazevic:
Use default coding style defined at CODINGSTYLE.md

### Description
In short, why is this pull request created. What it solves and how.

### Test case
Give us reason why you think it is correct solution.
For new code, make test which tests the addition.
For modifications, reference existing test.

### Closure
Put closes #XXXX in your comment to auto-close the issue that pull request fixes.


michalvasko:
Hi,
could you please continue with the CMake structure that I have created? There may be some changes required but I still think the only `CMakeLists.txt` that needs to be modified is the Python-specific one in `sysrepo/bindings/python/`.

Regards,
Michal

lpaulic:
While working on the test we found out that the `Subscription::additional_cleanup` function does not work as intended. SWIG can not override functions that C++ uses internally. Because of this the `unsubscribe` was extended in SWIG in that way that it does the same thing as in C++ but the `additional_cleanup` functions is embedded in  the SWIG extended function.

borna-blazevic:
Hi,

Sorry, I did not see the comment, It's done now.

Regards,
Borna

lpaulic:
All test except NotificationTest.py and SubscriptionTest.py are working. These two fail because of the libyang error that occurs while installing ietf-interfaces, ietf-ip, iana-if-type.
These two test are commented for now.

lpaulic:
The issue with the NotificationTest.py and SubscriptionTest.py are fixed. All tests should be working now.

michalvasko:
Hi,
I have fixed the test macro (please merge current `devel`) and when I tried running all the tests with this pull request changes, 2 failed (log attached) but I think it should be easy to fix. Do you have different output, does it work for you?

Regards,
Michal

[tests.log](https://github.com/sysrepo/sysrepo/files/3954838/tests.log)

lpaulic:
Hi Michal,
The tests.log file is empty I don't see anything.

michalvasko:
Hi,
sorry about that, did not paste the content.

Regards,
Michal

lpaulic:
Hi Michal,

I have created a docker file, cloned sysrepo (devel fork) and all sysrepo dependencies. Then I ran the Python tests manually and all of them passed.

What tests have failed in your case?

Regards,
Luka

michalvasko:
Hi Luka,
the problem may be exactly that you run them manually and not using `ctest` so maybe the CMakeLists.txt (`ADD_PYTHON_TEST` macro) needs a fix. The failing tests can be seen in the log I posted previously.

Regards,
Michal

lpaulic:
Hi Michal,

I have tried running the test with ctest and all test pass as expected.

Here is the output of the ctests command.
[ctest.log](https://github.com/sysrepo/sysrepo/files/3955330/ctest.log)

Regards,
Luka

michalvasko:
Hi Luka,
okay, it seems we are finally getting somewhere. After a bit of testing I learned that the problem was that the tests always use the installed *sysrepo* python bindings. When I first installed the bindings from this pull request, all the tests passed.

So, please fix this so that the bindings form the repository are always used, it will probably be some path from [here] (that I have removed in a previous commit and I should not have), so just add is as the test property. You can easily test it by not having any *sysrepo* python bindings installed. I could probably do it myself, but I am still fixing some other bugs. I assume you have till tomorrow, I will not be finished sooner.

Regards,
Michal

lpaulic:
Hi Michal,

I have added the the path to sysrepo Python module as a test property and tested it. Everything should be working now.

Regards,
Luka

michalvasko:
Hi Luka,
sorry, I did not get to his until today. And after a minor fix that I committed into your branch it seems to work as it should, thanks.

Regards,
Michal

PR #1702, opened 2019-12-10
c++ fix wrong order of argument in Session::set_error()

ishidawataru:


michalvasko:
Hi,
right, no way of allowing to use the variable arguments in C++.

Regards,
Michal

ishidawataru:
thanks for merging

PR #1740, opened 2020-01-07
CMakelists.txt: link with atomic

ffontaine:
On some architectures, atomic binutils are provided by the libatomic
library from gcc. Linking with libatomic is therefore necessary,
otherwise the build fails with:
```
/home/buildroot/autobuild/instance-2/output-1/host/opt/ext-toolchain/bin/../lib/gcc/sparc-buildroot-linux-uclibc/7.4.0/../../../../sparc-buildroot-linux-uclibc/bin/ld: libsysrepo.so.0.7.8: undefined reference to `__atomic_fetch_add_4'
```

This is often for example the case on sparc v8 32 bits.

Fixes:
 - http://autobuild.buildroot.org/results/d807f3dc922a61e099a66de3bc340c018a86e150

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


michalvasko:
Hi,
okay, but I will merge it to `devel`, we are keeping `master` to reflect the latest release.

Regards,
Michal

PR #1793, opened 2020-01-29
python3-bindings: SubscriptionTest.py increase sleep time

lpaulic:
In order to fix the issue regarding Travis CI build test failure for the ARM 64 architecture, this quick fix is presented to solve the issue.


michalvasko:
Hi,
okay, it seems the builds are working again, there were not yesterday. Anyway, did you test it without the byte code? What was the timeout required then?

Regards,
Michal

lpaulic:
Hi Michal,

I have tested it without the bytecode and then with and without the sleep timeouts the test seem to get stuck and the ctest freezes. I have removed the option PYTHONDONTWRITEBYTECODE=1 and than all python tests, with the sleep duration increases, work as expeced.

Is this suitable for now or is do you know a better way to dispose the bytcode that gets generated during the execution of the python tests?

Regards,
Luka

michalvasko:
Hi Luka,
as the comment says, I have added the flag just so that no files are written into the repository, which the Python tests do otherwise but it is no big deal. What I would really like is for the notification test, for example, not to take almost 10s but I guess that could happen only once all the sleeps are removed. So, just please keep the byte code flag there, just comment it out with information that it causes the Python tests to not work, thanks.

Regards,
Michal

PR #1826, opened 2020-02-19
Two new APIs provided to get data from full data tree in sr_val_t.

sriramkandukuri:
It is suggested to use sr_get_data to get data tree in libyang node type for better performance.
In implementations based on old sysrepo sr_val_t is used extensively.
Getting sr_val_t values from existing sr_get_item(s) is costly than getting data from sr_get_data.

To achive performance from sr_get_data and also to provide data
in sr_val_t type added new APIs sr_get_config_item(s)

Any subscriber or plugin who uses sr_get_item(s) earlier, can use these APIs with
a pre build data tree for better performance. Getting whole data tree at start of any
transaction and reusing the same with these APIs improve performance multifold.

### DIFF
--- a/src/utils/values.c
+++ b/src/utils/values.c
@@ -840,3 +840,85 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
 
     return len;
 }
+
+API int
+sr_tree_to_val(const struct lyd_node *data, const char *path, sr_val_t **value)
+{
+    sr_error_info_t *err_info = NULL;
+    struct ly_set *set = NULL;
+
+    SR_CHECK_ARG_APIRET(!data || !path || !value, NULL, err_info)
+
+    *value = NULL;
+
+    set = lyd_find_path(data, path);
+
+    if (!set) {
+        sr_errinfo_new_ly(&err_info, lyd_node_module(data)->ctx);
+        goto cleanup;
+    } else if (!set->number) {
+        /* Not building err_info to avoid error logs when no item found */
+        ly_set_free(set);
+        return SR_ERR_NOT_FOUND;
+    } else if (set->number > 1) {
+        sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, NULL, "More subtrees match \"%s\".", path);
+        goto cleanup;
+    }
+
+    /* create return value */
+    *value = malloc(sizeof **value);
+    SR_CHECK_MEM_GOTO(!*value, err_info, cleanup);
+
+    if ((err_info = sr_val_ly2sr(set->set.d[0], *value))) {
+        goto cleanup;
+    }
+
+    /* success */
+cleanup:
+    ly_set_free(set);
+    return sr_api_ret(NULL, err_info);
+}
+
+API int
+sr_tree_to_values(const struct lyd_node *data, const char *xpath, sr_val_t **values, size_t *value_cnt)
+{
+    sr_error_info_t *err_info = NULL;
+    struct ly_set *set = NULL;
+    uint32_t i;
+
+    SR_CHECK_ARG_APIRET(!data || !xpath || !values || !value_cnt, NULL, err_info);
+
+    *values = NULL;
+    *value_cnt = 0;
+
+    set = lyd_find_path(data, xpath);
+
+    if (!set) {
+        sr_errinfo_new_ly(&err_info, lyd_node_module(data)->ctx);
+        goto cleanup;
+    } else if (!set->number) {
+        /* Not building err_info to avoid error logs when no item found */
+        ly_set_free(set);
+        return SR_ERR_NOT_FOUND;
+    } else {
+        *values = calloc(set->number, sizeof **values);
+        SR_CHECK_MEM_GOTO(!*values, err_info, cleanup);
+
+        for (i = 0; i < set->number; ++i) {
+            if ((err_info = sr_val_ly2sr(set->set.d[i], (*values) + i))) {
+                goto cleanup;
+            }
+            ++(*value_cnt);
+        }
+    }
+
+    /* success */
+cleanup:
+    ly_set_free(set);
+    if (err_info) {
+        sr_free_values(*values, *value_cnt);
+        *values = NULL;
+        *value_cnt = 0;
+    }
+    return sr_api_ret(NULL, err_info);
+}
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
this seems to me is a reasonable idea and I have nothing against adding it into *sysrepo* with some adjustments. Firstly, please move the functions to `utils/values.c` and `utils/values.h` because that is where they belong. Additionally, please rename them to `sr_tree_to_val` and `sr_tree_to_values` and remove the `session` parameter, it is redundant. Thanks.

Regards,
Michal

sriramkandukuri:
Hi Michal,

Thanks for considering the changes.

Session is needed for returning error info. Please correct me.

Thanks & regards,
Sriram.K

michalvasko:
Hi Sriram,
yes, but I think these functions will be just fine with returning only the error code, there are not that many error situations. The additional argument is not worth it.

Regards,
Michal

sriramkandukuri:
Hi Michal,

I have updated as per your review comments and resolved some unwanted commits.
Please review again.

Thanks & Regards,
Sriram.K

src/utils/values.c
```diff
@@ -840,3 +840,77 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
 }
+
+API int
+sr_tree_to_val(struct lyd_node *data, const char *path, sr_val_t **value)
+{
+    sr_error_info_t *err_info = NULL;
+    struct ly_set *set = NULL;
+
+    if(!data || !path || !value)
+        return SR_ERR_INVAL_ARG;
```
michalvasko: Use `SR_CHECK_ARG_APIRET` macro instead.

src/utils/values.c
```diff
@@ -840,3 +840,77 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
+
+    *value = NULL;
+
+    set = lyd_find_path(data, path);
+
+    if (!set || !set->number) {
```
michalvasko: `!set` should probably be handled specially, look [here](https://github.com/sysrepo/sysrepo/blob/master/src/common.c#L3135), for example.

src/utils/values.c
```diff
@@ -840,3 +840,77 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
 
     return len;
 }
+
+API int
+sr_tree_to_val(struct lyd_node *data, const char *path, sr_val_t **value)
```
michalvasko: `data` should be `const`.

src/utils/values.c
```diff
@@ -840,3 +840,77 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
+
+    if(!data || !xpath || !values || !value_cnt)
+        return SR_ERR_INVAL_ARG;
+
+    *values = NULL;
+    *value_cnt = 0;
+
+    set = lyd_find_path(data, xpath);
+
+    if(set) {
```
michalvasko: Formatting.

src/utils/values.c
```diff
@@ -840,3 +840,77 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
+        if (set->number) {
+            *values = calloc(set->number, sizeof **values);
+            SR_CHECK_MEM_GOTO(!*values, err_info, cleanup);
```
michalvasko: `set` leaks here.

src/utils/values.c
```diff
@@ -840,3 +840,77 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
+        }
+
+        for (i = 0; i < set->number; ++i) {
+            if ((err_info = sr_val_ly2sr(set->set.d[i], (*values) + i))) {
+                goto cleanup;
```
michalvasko: Again, `set` leak.

src/utils/values.h
```diff
@@ -190,6 +190,28 @@ char *sr_val_to_str(const sr_val_t *value);
  */
 int sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size);
 
+/**
+ * @brief Similar to ::sr_get_item but only gets configuration data from given data tree.
```
michalvasko: Comment is wrong and insufficient. It should first say what it does - basically transforms `struct lyd_node` to `sr_val_t`. Then the intentional usage should be described, the one you mentioned, that if you want to work with the structures but do not want to call `sr_get_item(s)` many times because of performance, you can call `sr_get_data()` once and then use this function.

michalvasko:
Hi Sriram,
these are fairly minor problems that I can fix myself, if you want, it is up to you.

Regards,
Michal

sriramkandukuri:
Hi Michal,

I will fix and update the pull request.

Thanks & Regards,
Sriram.K

src/utils/values.c
sriramkandukuri: Hi Michal,

Actually empty set can be possible if the requested node's module is not present in data tree.
I came across such scenarios in my testing. 

Any way node is not found, so returning NOT_FOUND error. Does it cause any unexpected? Please help me to understand.

Thanks & Regards,
Sriram.K

sriramkandukuri: Fixed as per your comment

src/utils/values.c
```diff
@@ -840,3 +840,80 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
+    uint32_t i;
+
+    SR_CHECK_ARG_APIRET(!data || !xpath || !values || !value_cnt, NULL, err_info);
+
+    *values = NULL;
+    *value_cnt = 0;
+
+    set = lyd_find_path(data, xpath);
+
+    if (set) {
```
michalvasko: Should probably be handled similarly as in the previous function, explicit check and error.

sriramkandukuri: Done please check

src/utils/values.c
```diff
@@ -840,3 +840,87 @@ sr_val_to_buff(const sr_val_t *value, char buffer[], size_t size)
...
+
+    if (!set) {
+        sr_errinfo_new_ly(&err_info, lyd_node_module(data)->ctx);
+        goto cleanup;
+    } else if (!set->number) {
+        /* Not building err_info to avoid error logs when no item found */
+        ly_set_free(set);
+        return SR_ERR_NOT_FOUND;
+    } else {
+        if (set->number) {
```
michalvasko: Useless check.

sriramkandukuri: Done please check

michalvasko:
Okay, I have no more objections, merging.

Regards,
Michal

PR #1848, opened 2020-03-06
python binding BUGFIX global_loop should run more than once

johnbeckettn2e:
Python environments such as jupyter notebooks allow modifying the
code on the fly, injecting a ctrl-c and re-running the same piece
of code.

The global_loop function should reset exit_application so that the
loop will not exit immediately in-case another call to global_loop
is made.


johnbeckettn2e:
this should have been raised against devel pull request 1850 rasied

Issue #1854, opened 2020-03-11
python bindings question - rpc_subscribe


johnbeckettn2e:
Hi,

I'm trying to use the python bindings to make a python example based on the contents of ` examples/rpc_subscribe_example.c`, which I can run and works.

I'm stuck with trying to get rpc_subscribe to work as I don't seem to be able to get the arguments correct for the callback function which I am trying to register.

I am using v1.4.2 of sysrepo and my environment has the ``examples`` yang module installed:
````
$  sysrepoctl -l                                                                                         
Sysrepo repository: /sysrepo/sysrepo/build/repository                                                                              

Module Name                | Revision   | Flags | Owner     | Permissions | Submodules | Features                                  
-------------------------------------------------------------------------------------------------                                  
examples                   |            | I     | root:root | 600         |            |                                           
ietf-datastores            | 2017-08-17 | i     |           |             |            |                                           
ietf-inet-types            | 2013-07-15 | i     |           |             |            |                                           
ietf-netconf               | 2011-06-01 | I     | root:root | 666         |            |                                           
ietf-netconf-notifications | 2012-02-06 | I     | root:root | 666         |            |                                           
ietf-netconf-with-defaults | 2011-06-01 | I     | root:root | 666         |            |                                           
ietf-origin                | 2018-02-14 | I     | root:root | 666         |            |                                           
ietf-yang-library          | 2019-01-04 | I     | root:root | 666         |            |                                           
ietf-yang-metadata         | 2016-08-05 | i     |           |             |            |                                           
ietf-yang-types            | 2013-07-15 | i     |           |             |            |                                           
yang                       | 2017-02-20 | I     | root:root | 666         |            |                                           
````
Here is the code that I am trying to use:
```python
import sysrepo as sr
module_name = "examples"
rpc_path = "/" + module_name + ":oper"

# This fails when registering - bad prototype??
def my_rpc_cb(cb_session, op_path, in_vals, evt, request_id, holder, private_data):
    print(in_vals)
    return sr.SR_ERR_OK

connection = sr.Connection()
session = sr.Session(connection, sr.SR_DS_RUNNING)
subscribe = sr.Subscribe(session)

# Running the following line causes runtime error:
subscribe.rpc_subscribe(rpc_path, my_rpc_cb)
```
This fails with the following:
```
RuntimeError                              Traceback (most recent call last)
<ipython-input-5-49855a15c4a7> in <module>()
      3 subscribe = sr.Subscribe(session)
      4 
[...]

PR #1857, opened 2020-03-11
python3-bindings: fix rpc subscribe

johnbeckettn2e:
Potential fix for issue #1854 

Looks like when renaming `private_ctx` to `private_data` part of the python binding was missed.

I've tested that using the c `rpc_send_example /examples:oper` that there is a callback with the following python:
```python
import sysrepo as sr
module_name = "examples"
rpc_path = "/" + module_name + ":oper"

def my_rpc_cb(cb_session, op_path, in_vals, evt, request_id, holder, private_data):
    print(in_vals)
    return sr.SR_ERR_OK

connection = sr.Connection()
session = sr.Session(connection, sr.SR_DS_RUNNING)
subscribe = sr.Subscribe(session)
subscribe.rpc_subscribe(rpc_path, my_rpc_cb)

sr.global_loop()
```

### DIFF
--- a/bindings/swig_base/cpp_classes.i
+++ b/bindings/swig_base/cpp_classes.i
@@ -45,7 +45,7 @@
 %newobject Session::action_send;
 
 %shared_ptr(sysrepo::Callback);
-%ignore Callback::private_ctx;
+%ignore Callback::private_data;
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you are right, thanks for looking into it.

Regards,
Michal

PR #1612, opened 2019-09-21
Add uninstall and uninstall_with_repo capability

brchiu:
Add `make uninstall` and `make uninstall_with_repo` capabilities as master branch does.

PR #1907, opened 2020-04-10
Add some testcases for lock

fredgan:


fredgan:
I don't know why no check triggered , I also find another one like this #1612

tests/test_lock.c
```diff
@@ -150,7 +172,35 @@ test_one_session(void **state)
...
+    
+    /* lock all modules */
+    ret = sr_lock(sess, NULL);
+    assert_int_equal(ret, SR_ERR_OK);
+    
+    /* create a candidate datastore file in /dev/shm */
+    ret = asprintf(&path, "%s/sr_%s.candidate", SR_SHM_DIR, "test");
+    assert_int_not_equal(ret, -1);   
+    ret = create_file(path, 0666);
+    assert_int_equal(ret, SR_ERR_OK);
```
michalvasko, line 189: Please, do not use similar hacks, there is no need for them nor for an additional `utils.c` file for that matter. You can simply set some changes in the `candidate` datastore and it will then be created.

michalvasko:
Never mind the travis problem, they are being run (and passed for your case) except they never finish because of a platform support in beta that seems to not work correctly, I will try to fix it.

tests/test_lock.c
```diff
@@ -35,6 +35,7 @@
 
 #include "tests/config.h"
 #include "sysrepo.h"
+#include "utils.h"
```
michalvasko, line 38: You forgot to remove this.

fredgan: sorry,it's my careless,I will modify it immediately.

michalvasko:
Okay, now it looks fine, thanks.

Regards,
Michal

PR #1921, opened 2020-04-16
Add editorconfig file

rjarry:
EditorConfig is a file format and collection of text editor plugins for maintaining consistent coding styles between different editors and IDEs.

Initialize the file following the current coding style in existing files.

In order for this file to be taken into account (unless they use an editor with built-in EditorConfig support), developers will have to install a plugin.

See the following links for more details.

https://editorconfig.org/
https://github.com/editorconfig/editorconfig-emacs
https://github.com/editorconfig/editorconfig-vim

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# ex: ft=dosini
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
okay I suppose, I have nothing against adding this small config file into the repo even though we are not using this tool.

Regards,
Michal

PR #1922, opened 2020-04-16
Disable cwd searchdir in libyang context

rjarry:
YANG models should only be searched in sysrepo YANG dir.


michalvasko:
Hi,
okay, that makes sense, thanks.

Regards,
Michal

Issue #1931, opened 2020-04-20
there are some problems  when I call connection diff callback 'sr_diff_check_cb' which fails in fuction 'sr_changes_notify_store'


fredgan:
Hi Michal,
  I implement a simple callback sr_diff_check_cb  `diff_check_cb` and simply set changes in a running datastore. It causes a coredump because of `session->err_info ` which is  NULLL when `sr_set_error` is not been used.

What's more, I use sr_set_error to set errors, however, `sr_set_error` fails error SR_ERR_INVAL_ARG because the `session->ev` value is SR_SUB_EV_NONE which is not been supported in sr_set_error.

The code is as follows:

```
 static sr_error_info_t *
sr_changes_notify_store(struct sr_mod_info_s *mod_info, sr_session_ctx_t *session, uint32_t timeout_ms, int wait,sr_error_info_t **cb_err_info)
{
         ......
    /* call connection diff callback */
    if (mod_info->diff && session->conn->diff_check_cb && (ret = session->conn->diff_check_cb(session, mod_info->diff))) {
        /* create cb_err_info */
        if (session->err_info && session->err_info->err_code == SR_ERR_OK) {
            err_msg = session->err_info->err[0].message;
        }
        if (!err_msg) {
            err_msg = sr_strerror(ret);
        }
       // whether to add  judgment session->err_info == NULL? 
        err_xpath = session->err_info->err[0].xpath; 
        sr_errinfo_new(cb_err_info, ret, err_xpath, err_msg);
        goto cleanup;
    }
       ......
}
API int
sr_set_error(sr_session_ctx_t *session, const char *path, const char *format, ...)
{
    sr_error_info_t *err_info = NULL;
    va_list vargs;

    SR_CHECK_ARG_APIRET(!session || ((session->ev != SR_SUB_EV_CHANGE) && (session->ev != SR_SUB_EV_UPDATE)
            && (session->ev != SR_SUB_EV_OPER) && (session->ev != SR_SUB_EV_RPC)) || !format, session, err_info);
      ......
} 
```

test code is as follows :
```
static int
diff_check_cb(sr_session_ctx *session, const struct lyd_node *diff)
{
    const char *user;
    struct passwd *pwd;

    user = sr_session_get_user(session);
    pwd  = getpwuid(getuid());
    if (strncmp(user, pwd->pw_name, strlen(pwd->pw_name))) {
       /*  set  error when failure , no error been set,session->err_info NULL causes coredump  */
[...]

PR #1935, opened 2020-04-22
add sr_set_diff_check_callback testcases for connection and session

fredgan:
This is only a part of testcase for  issue #1931 .I will add other testcases for connection and session later.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -22,7 +22,7 @@ configure_file("${PROJECT_SOURCE_DIR}/tests/config.h.in" "${PROJECT_BINARY_DIR}/
 
 # lists of all the tests
 set(tests test_modules test_validation test_edit test_candidate test_operational test_lock test_apply_changes
-    test_copy_config test_rpc_action test_notif test_get test_process)
+    test_copy_config test_rpc_action test_notif test_get test_process test_connection_and_session)
 
 foreach(test_name IN LISTS tests)
     add_executable(${test_name} ${test_sources} ${test_name}.c)
--- a/tests/test_connection_and_session.c
+++ b/tests/test_connection_and_session.c
@@ -1,0 +1,213 @@
+/**
+ * @file test_connection_and_session.c
+ * @author Fred Gan <ganshaolong@vip.qq.com>
+ * @brief test for sysrepo API, connection and session
+ *
+ * @copyright
+ * Copyright 2018 Deutsche Telekom AG.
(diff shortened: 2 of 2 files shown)


michalvasko:
Hi Fred,
I see what you tried to do but there are a few problems. Generally, I would like to avoid calling `system()` but that is not really the problem. To add users, you would need to run this test as `root` and that should be avoided if possible. However, you only need 2 users for the test. You can use the user of the process and then just take another arbitrary one from the system. That should not be a problem using the functions for traversing all the users on a system.

That way both requiring `root` access and modifying the users on the system would be avoided.

Regards,
Michal

michalvasko:
Okay, the test passed now but what about the name `connection_and_session`? It is quite vague and I am also not sure it covers the actual test. Do you plan to add any more, what exactly will they test?

Regards,
Michal

fredgan:
@michalvasko 

Hi Michal,
Do you have any suggestions or tips for the form of the testcases? Which part should be key module that we should focus on. In fact, now I am not quite sure whether I am in the right direction.

michalvasko:
Hi,
well, you see that even this test initially failed so I have no problem with it in general. However, what is mostly needed are use-cases of functions that allow large number of different uses, made even more if they are combined.

One such (family of) function is `sr_set_item_str()` because its behavior depend on any already existing changes in the session. But there are not that many cases and I think they should be fairly well covered by now. You can still take a look and I am certain you can come up with new use-cases and perhaps even some non-working ones.

The worst in this aspect are subscriptions. You can have as many of them as you want, they are combined to communicate using a single shared memory if they operate on the same module, and there are even priorities and callback failures (abort events) to worry about. This is a huge number of possibilities that can never be exhaustively tested. My approach was to test the cases I could think of and though potentially problematic while adding any new ones discovered in the issues. So, you can definitely add more of these and recently even a `test_process` was added that includes a basic infrastructure for testing running several *sysrepo* processes at once. That can be additionally used to test concurrency.

But I was only talking about change subscriptions. You can add operational ones that add another layer of complexity and are complicated even on their own because they allow nesting or overlapping.

Note that to come up with reasonable test cases for these mechanisms, you need an in-depth understanding of them. So starting with simpler tests and slowly making them more complex is probably the way to go. Also, any feedback regarding the documentation is as important as actual tests. Meaning please report any behavior you had to learn on your own and could be documented in a few sentences instead.

Regards,
Michal

fredgan:
Hi Michal, Thanks for your tips. I will take them consider.

PR #2000, opened 2020-05-27
Fix XPath trimming predidcates heap overflow

fredgan:
Hi Michal,
  I'm implementing the test cases of rpc and xpath part. I find a heap overflow issue for the invalid xpath  as following. I tried to fix it by this PR. Please review it.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -2557,14 +2557,14 @@ sr_get_trim_predicates(const char *expr, char **expr2)
         }
     }
 
-    /* copy last expr chunk */
-    strncat(str, start, ptr - start);
-
     if (quot || pred) {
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
your fix works but I do not think it really makes sense in the logic of the function. I think it is better to switch the order of [this](https://github.com/sysrepo/sysrepo/blob/devel/src/common.c#L2561) and [this](https://github.com/sysrepo/sysrepo/blob/devel/src/common.c#L2563) statement. Your test then passed for me. Also, if this was its final placement, please put the test separately as it tests only the input parameters. The rest of the tests execute some callbacks. Thanks.

Regards,
Michal

michalvasko:
Hi Fred,
seems good to me now, thanks.

Regards,
Michal

Issue #2027, opened 2020-06-16
sysrepoctl does not list imports of submodules


pfeige:
Hi colleagues,

for submodules the sysrepoctl command does not list the imported modules. This can be verified by using the yang files from this commit https://github.com/ADTRAN/sysrepo/commit/56d6bc8a36eb42f8b059204511ae705d4a6e0420. After importing the module srctl-lst.yang a sysrepoctl -l shows only the imported module srctl-lst-imp but not the module srctl-lst-inc-imp imported by the submodule srctl-lst-inc.

With the proposed solution in the commit https://github.com/ADTRAN/sysrepo/commit/b6ebb92a929bfa33ecc608ecf56439e7fba608e6 this issue is solved and srctl-lst-inc-imp is also listed.

Best Regards,
Peter

michalvasko:
Hi Peter,
okay, seems fine to me so you can create a pull request with the fix. Thanks.

Regards,
Michal

PR #2028, opened 2020-06-16
sysrepoctl does not list imports of submodules

pfeige:
See issue https://github.com/sysrepo/sysrepo/issues/2027.


michalvasko:
Seems good, thanks.

Regards,
Michal

Issue #2033, opened 2020-06-25
Shared mutex dead lock


rjarry:
Hi Michal,

We frequently encounter a dead lock with the main mutex. It seems there are cases where it is not properly released by the application that acquired it. Causing all further attempts to acquire it to fail since the application has exited and will never be able to release it. Since it is a shared pthread mutex, the OS (in our case, Linux) has no chance to cleanup and release it. Basically, all sysrepo applications are "frozen" and cannot perform anything. The only way to recover is to `SIGKILL` all these applications and purge all the shared memory files (not far from rebooting the system).

I have spent a lot of time trying to debug the issue but to no avail. Somehow, I managed to reproduce the bug (almost every time) without all our environment which allows me to submit this issue :)

Here are the versions I am using:

Libyang: 1.0.180
Sysrepo: 1.4.67

Build both projects with default options and apply the following patch to sysrepo to add code to reproduce the issue:

```bash
cd $path_to_sysrepo
curl https://gist.githubusercontent.com/rjarry/a19629ae41668a46d44566b4ea16e0d4/raw/47fc7e4e3b7e553f5d6ea223048a1b08ac47d4c7/mutex-bug.patch | git am
cd mutex-bug
make   # takes quite some time as there are a lot of YANG modules to install
bash -x test.sh
```

Wait a few seconds and then hit `ctrl-c`. You should get:

```
...
++ sysrepoctl -C
0
++ sysrepoctl -l
[ERR]: Locking a mutex failed (sr_connect: Connection timed out).
sysrepoctl error: Failed to connect (Timeout expired)
```

As you can see, no sysrepo applications are running but the mutex cannot be acquired anymore.

The problem does not appear every time. It seems to depend when `ctrl-c` is hit. Also, it is very hard to reproduce with only a few 
YANG modules installed. This is why we install 300 dummy modules in the `Makefile`. Only the first two are actually used by the applications.

Thanks a lot in advance for your help!

rjarry:
In my opinion, this is a critical issue.

Ideally, if one shared mutex has been acquired by an application which died without releasing it (it can happen for many reasons, segmentation fault, bug in sysrepo, `SIGKILL` while in sysrepo code, etc.), there should be an automatic way of recovering from this.

This could be an interesting lead: https://stackoverflow.com/questions/30770236/pthread-process-shared-mutex-deadlock However I'm not sure this is POSIX compliant.

rjarry:
[...]

PR #2041, opened 2020-06-30
shm CHANGE allow setting a custom prefix for SHM files

rjarry:
Allow exporting a `SYSREPO_SHM_PREFIX` environment variable. When that variable is set, all files created under `/dev/shm` will use that prefix. If unset, the files wille have their "default" names.

This can be useful for testing purposes, or to run multiple "groups" of sysrepo applications in concordance with the `SYSREPO_REPOSITORY_PATH` environment variable.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1418,16 +1418,78 @@ sr_create_module_imps_incs_r(const struct lys_module *ly_mod)
     return NULL;
 }
 
+static sr_error_info_t *
+sr_shm_prefix(const char **prefix)
+{
+    sr_error_info_t *err_info = NULL;
+
+    *prefix = getenv(SR_SHM_PREFIX_ENV);
+    if (*prefix == NULL) {
+        *prefix = SR_SHM_PREFIX_DEFAULT;
+    } else if (strchr(*prefix, '/') != NULL) {
+        *prefix = NULL;
+        sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, NULL, "%s cannot contain slashes.", SR_SHM_PREFIX_ENV);
+    }
+
+    return err_info;
+}
+
+sr_error_info_t *
+sr_path_main_shm(char **path)
+{
+    sr_error_info_t *err_info = NULL;
+    const char *prefix;
+
+    err_info = sr_shm_prefix(&prefix);
+    if (err_info) {
+        return err_info;
+    }
+
+    if (asprintf(path, "/%s_main", prefix) == -1) {
+        SR_ERRINFO_MEM(&err_info);
+        *path = NULL;
+    }
+
+    return err_info;
+}
+
+sr_error_info_t *
+sr_path_ext_shm(char **path)
+{
+    sr_error_info_t *err_info = NULL;
+    const char *prefix;
+
+    err_info = sr_shm_prefix(&prefix);
+    if (err_info) {
+        return err_info;
+    }
+
+    if (asprintf(path, "/%s_ext", prefix) == -1) {
+        SR_ERRINFO_MEM(&err_info);
+        *path = NULL;
+    }
+
+    return err_info;
+}
+
 sr_error_info_t *
 sr_path_sub_shm(const char *mod_name, const char *suffix1, int64_t suffix2, int abs_path, char **path)
 {
     sr_error_info_t *err_info = NULL;
+    const char *prefix;
     int ret;
 
+    err_info = sr_shm_prefix(&prefix);
+    if (err_info) {
+        return err_info;
+    }
+
     if (suffix2 > -1) {
-        ret = asprintf(path, "%s/srsub_%s.%s.%08x", abs_path ? SR_SHM_DIR : "", mod_name, suffix1, (uint32_t)suffix2);
+        ret = asprintf(path, "%s/%ssub_%s.%s.%08x", abs_path ? SR_SHM_DIR : "",
+                prefix, mod_name, suffix1, (uint32_t)suffix2);
     } else {
-        ret = asprintf(path, "%s/srsub_%s.%s", abs_path ? SR_SHM_DIR : "", mod_name, suffix1);
+        ret = asprintf(path, "%s/%ssub_%s.%s", abs_path ? SR_SHM_DIR : "",
+                prefix, mod_name, suffix1);
     }
 
     if (ret == -1) {
@@ -1440,11 +1502,18 @@ sr_error_info_t *
 sr_path_ds_shm(const char *mod_name, sr_datastore_t ds, int abs_path, char **path)
 {
     sr_error_info_t *err_info = NULL;
+    const char *prefix;
     int ret;
 
     assert((ds == SR_DS_RUNNING) || (ds == SR_DS_CANDIDATE) || (ds == SR_DS_OPERATIONAL));
 
-    ret = asprintf(path, "%s/sr_%s.%s", abs_path ? SR_SHM_DIR : "", mod_name, sr_ds2str(ds));
+    err_info = sr_shm_prefix(&prefix);
+    if (err_info) {
+        return err_info;
+    }
+
+    ret = asprintf(path, "%s/%s_%s.%s", abs_path ? SR_SHM_DIR : "",
+            prefix, mod_name, sr_ds2str(ds));
     if (ret == -1) {
         *path = NULL;
         SR_ERRINFO_MEM(&err_info);
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -352,8 +352,20 @@ cleanup:
         sr_conn_free(conn);
         if (created) {
             /* remove any created SHM so it is not considered properly created */
-            shm_unlink(SR_MAIN_SHM);
-            shm_unlink(SR_EXT_SHM);
+            sr_error_info_t *tmp_err = NULL;
+            char *shm_name = NULL;
+            if ((tmp_err = sr_path_main_shm(&shm_name))) {
+                sr_errinfo_merge(&err_info, tmp_err);
+            } else {
+                shm_unlink(shm_name);
+                free(shm_name);
+            }
+            if ((tmp_err = sr_path_ext_shm(&shm_name))) {
+                sr_errinfo_merge(&err_info, tmp_err);
+            } else {
+                shm_unlink(shm_name);
+                free(shm_name);
+            }
         }
     } else {
         *conn_p = conn;
--- a/src/common.h.in
+++ b/src/common.h.in
@@ -131,6 +131,12 @@ int pthread_mutex_timedlock(pthread_mutex_t *mutex, const struct timespec *absti
 /** where SHM files are stored */
 #define SR_SHM_DIR "/dev/shm"
 
+/** default prefix for SHM files in /dev/shm */
+#define SR_SHM_PREFIX_DEFAULT "sr"
+
+/** environment variable for setting a custom prefix for SHM files */
+#define SR_SHM_PREFIX_ENV "SYSREPO_SHM_PREFIX"
+
 /** all ext SHM item sizes will be aligned to this number (B) */
 #define SR_SHM_MEM_ALIGN (sizeof(void *))
 
@@ -689,6 +695,22 @@ sr_error_info_t *sr_create_startup_file(const struct lys_module *ly_mod);
  * @return err_info, NULL on success.
  */
 sr_error_info_t *sr_create_module_imps_incs_r(const struct lys_module *ly_mod);
+
+/**
+ * @brief Get the path the main SHM.
+ *
+ * @param[out] path Created path. Should be freed by the caller.
+ * @return err_info, NULL on success.
+ */
+sr_error_info_t *sr_path_main_shm(char **path);
+
+/**
+ * @brief Get the path the external SHM.
+ *
+ * @param[out] path Created path. Should be freed by the caller.
+ * @return err_info, NULL on success.
+ */
+sr_error_info_t *sr_path_ext_shm(char **path);
 
 /**
  * @brief Get the path to a subscription SHM.
--- a/src/shm.h
+++ b/src/shm.h
@@ -30,8 +30,6 @@
 
 #include "common.h"
 
-#define SR_MAIN_SHM "/sr_main"              /**< Main SHM name. */
-#define SR_EXT_SHM "/sr_ext"                /**< External SHM name. */
 #define SR_MAIN_SHM_LOCK "sr_main_lock"     /**< Main SHM file lock name. */
 #define SR_SHM_VER 1                        /**< Main and ext SHM version of their expected content structures. */
 
--- a/src/shm_main.c
+++ b/src/shm_main.c
@@ -1716,14 +1716,21 @@ sr_shmmain_main_open(sr_shm_t *shm, int *created)
 {
     sr_error_info_t *err_info = NULL;
     sr_main_shm_t *main_shm;
+    char *shm_name = NULL;
     int creat = 0;
     mode_t um;
 
+    err_info = sr_path_main_shm(&shm_name);
+    if (err_info) {
+        return err_info;
+    }
+
     /* try to open the shared memory */
-    shm->fd = shm_open(SR_MAIN_SHM, O_RDWR, SR_MAIN_SHM_PERM);
+    shm->fd = shm_open(shm_name, O_RDWR, SR_MAIN_SHM_PERM);
     if ((shm->fd == -1) && (errno == ENOENT)) {
         if (!created) {
             /* we do not want to create the memory now */
+            free(shm_name);
             return NULL;
         }
 
@@ -1731,10 +1738,11 @@ sr_shmmain_main_open(sr_shm_t *shm, int *created)
         um = umask(00000);
 
         /* create shared memory */
-        shm->fd = shm_open(SR_MAIN_SHM, O_RDWR | O_CREAT | O_EXCL, SR_MAIN_SHM_PERM);
+        shm->fd = shm_open(shm_name, O_RDWR | O_CREAT | O_EXCL, SR_MAIN_SHM_PERM);
         umask(um);
         creat = 1;
     }
+    free(shm_name);
     if (shm->fd == -1) {
         sr_errinfo_new(&err_info, SR_ERR_SYS, NULL, "Failed to open shared memory (%s).", strerror(errno));
         goto error;
@@ -1783,12 +1791,19 @@ sr_error_info_t *
 sr_shmmain_ext_open(sr_shm_t *shm, int zero)
 {
     sr_error_info_t *err_info = NULL;
+    char *shm_name = NULL;
     mode_t um;
+
+    err_info = sr_path_ext_shm(&shm_name);
+    if (err_info) {
+        return err_info;
+    }
 
     /* set umask so that the correct permissions are really set */
     um = umask(00000);
 
-    shm->fd = shm_open(SR_EXT_SHM, O_RDWR | O_CREAT, SR_MAIN_SHM_PERM);
+    shm->fd = shm_open(shm_name, O_RDWR | O_CREAT, SR_MAIN_SHM_PERM);
+    free(shm_name);
     umask(um);
     if (shm->fd == -1) {
         sr_errinfo_new(&err_info, SR_ERR_SYS, NULL, "Failed to open ext shared memory (%s).", strerror(errno));


michalvasko:
Hi Robin,
I get the idea and I think it should be possible to implement but in a different way. You see, SHM files are created with `shm_open(3)`, which will always ignore whatever prefix you want to set, it always uses the same one (and it should be `/dev/shm`). The only way to allow several *sysrepo* instances would be to add a dir after the base path so you get `/dev/shm/dir/sr_main`, for example. At least I think it should be possible to use a prefix in the name given to `shm_open()`.

Regards,
Michal

rjarry:
Hi Michal,

unfortunately, this is not possible. `shm_open(3)` (on linux) states:

> The operation of `shm_open()` is analogous to that of `open(2)`.  `name` specifies the shared memory object  to  be created or opened.  For portable use, a shared memory object should be identified by a name of the form `/somename`; that is, a null-terminated string of up to `NAME_MAX` (i.e., 255) characters consisting  of  an  initial slash, followed by one or more characters, __none of which are slashes__.

The POSIX man page is also conservative about that:

> It is unspecified  whether the  `name`  appears in the file system and is visible to other functions that take pathnames as arguments. The `name` argument conforms to the construction rules for a pathname, except that the  interpretation  of  <slash> characters  other  than  the leading <slash> character in name is implementation-defined, and that the length limits for the name argument are implementation-defined and need not be  the  same  as  the  pathname  limits {PATH_MAX} and {NAME_MAX}.  If name begins with the <slash> character, then processes calling `shm_open()` with the same value of name refer to the same shared memory object, as long as that name has not been removed.  If name does not begin with the <slash> character, the effect is implementation-defined.

The idea behind this patch is to use for example:

```bash
export SYSREPO_SHM_PREFIX="group1-"
```

And it will create `/dev/shm/group1-*` files.

michalvasko:
I see, okay, then what I suggested is not possible what in the pull request you have not changed names for `shm_open()` at all so it would not work the way you said it should. Also, all *sysrepo* SHM files use either `sr_` or `srsub_` prefix (unless I forgot something), which are currently hard-coded so the best solution would probably be to unify all these prefixes to be returned by a function that can get them from the environment.

Regards,
Michal

rjarry:
Unless I made a mistake, the `sr_` or `srsub_` prefixes are still here. `SYSREPO_SHM_PREFIX` (if defined) is only prepended before these.

Here is an example:

```bash
~# SYSREPO_SHM_PREFIX="group1-" ./application &
~# ls -l /dev/shm/group1-*
-rw-rw-rw- 1 root root 1.9K 07-01 13:42 /dev/shm/group1-sr_ext
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-datastores.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-datastores.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-netconf-notifications.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-netconf-notifications.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-netconf.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-netconf.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-netconf-with-defaults.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-netconf-with-defaults.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-origin.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-origin.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-yang-library.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_ietf-yang-library.running
-rw-rw-rw- 1 root root 6.5K 07-01 13:42 /dev/shm/group1-sr_main
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_sysrepo-monitoring.operational
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_sysrepo-monitoring.running
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_mymodule.operational
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_mymodule.running
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1-srsub_mymodule.oper.c2a945a3
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_yang.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1-sr_yang.running
```

Do you mean you would like to use the same `sr` prefix by default for all SHM files and change it if `SYSREPO_SHM_PREFIX` is set?

michalvasko:
Oh, sorry, my bad, the prefix is always used. Okay, so please put getting the prefix into separate function like I said before and use it on these 2 places. It should probably also be checked in the function that the prefix does not include any `/`.

Regards,
Michal

rjarry:
Ok I'll do that.

What do you think about completely replacing the `sr` prefix when `SYSREPO_SHM_PREFIX` is set? That would give something like this:

With `SYSREPO_SHM_PREFIX="group1"`:

```
-rw-rw-rw- 1 root root 1.9K 07-01 13:42 /dev/shm/group1_ext
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-datastores.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-datastores.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-netconf-notifications.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-netconf-notifications.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-netconf.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-netconf.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-netconf-with-defaults.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-netconf-with-defaults.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-origin.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-origin.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-yang-library.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_ietf-yang-library.running
-rw-rw-rw- 1 root root 6.5K 07-01 13:42 /dev/shm/group1_main
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1_sysrepo-monitoring.operational
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1_sysrepo-monitoring.running
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1_mymodule.operational
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1_mymodule.running
-rw------- 1 root root    7 07-01 13:42 /dev/shm/group1_sub_mymodule.oper.c2a945a3
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_yang.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/group1_yang.running
```

And if SYSREPO_SHM_PREFIX is not set:

```
-rw-rw-rw- 1 root root 1.9K 07-01 13:42 /dev/shm/sr_ext
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-datastores.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-datastores.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-netconf-notifications.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-netconf-notifications.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-netconf.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-netconf.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-netconf-with-defaults.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-netconf-with-defaults.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-origin.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-origin.running
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-yang-library.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_ietf-yang-library.running
-rw-rw-rw- 1 root root 6.5K 07-01 13:42 /dev/shm/sr_main
-rw------- 1 root root    7 07-01 13:42 /dev/shm/sr_sysrepo-monitoring.operational
-rw------- 1 root root    7 07-01 13:42 /dev/shm/sr_sysrepo-monitoring.running
-rw------- 1 root root    7 07-01 13:42 /dev/shm/sr_mymodule.operational
-rw------- 1 root root    7 07-01 13:42 /dev/shm/sr_mymodule.running
-rw------- 1 root root    7 07-01 13:42 /dev/shm/sr_sub_mymodule.oper.c2a945a3
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_yang.operational
-rw-rw-rw- 1 root root    7 07-01 13:42 /dev/shm/sr_yang.running
```

Note that this would change `srsub_` to `sr_sub_` for subscriptions.

rjarry:
Or I can skip the additional `_` to preserve existing file names.

michalvasko:
I guess it is fine to completely replace the prefix but I would prefer to keep it the way it is if the prefix is not set. You then always have `sr_main` and `sr_ext`, which are custom shared memory chunks, `sr_*`, which are LYB files, and `srsub_*`, which are subscription shared memory chunks. Changing it to `sr_sub_*` could cause collisions with modules starting with `sub_`.

Regards,
Michal

rjarry:
It took me more time than I expected. I did not understand what you were saying:

> which are currently hard-coded

It should be OK now.

michalvasko:
Hi Robin,
I am sorry but I do not like the current changes. I am not even sure what exactly your use-case is when you need to have several independent *sysrepo* instances but the added complexity is not worth it in my opinion. To learn from the code the final path for some SHM file was difficult enough, now it is almost impossible. So, unless you can somehow simplify it (even I can attempt it once I have some time), I would prefer to keep it the way it is now without this feature.

Regards,
Michal

rjarry:
Hi Michal,

The main use-case is parallel testing and automated builds. The big issue here is that you cannot (per POSIX standards) control where the SHM files are created by `shm_open()`. If you have a build system (for example, [buildroot](https://buildroot.org/)) that needs to call `sysrepoctl` and/or `sysrepocfg` commands to install YANG modules and prepare startup configs into a staging directory (before generation of a `bzImage`), without my change there is no way to have multiple builds in parallel. Not to mention the issue if the build host is already running sysrepo applications.

This change allows some kind of isolation by using unique `SYSREPO_SHM_PREFIX` values for each separate build or test run. For example I am planning to use this feature for unit tests of the sysrepo python bindings I am working on.

I tried hard to find another solution to trick `shm_open()` into using a specific path but that is not possible without **major** `LD_PRELOAD` hacks. If you give me guidance on how to implement this in a simpler way, I can adapt my code.

Thanks

michalvasko:
Okay, fair enough. So these are my ideas that may make the code better.

There would be 2 prefixes, one `sr_` by default, the other `srsub_` by default. There would be a function to retrieve each of them. In your case you could set both prefixes to be the same but the distinction should be kept.

As for `main` and `ext` SHM, there could be again 2 functions that would generate their path and they would use the `sr_` prefix. I think that may make it a bit clearer in the code what is actually happenning (`sr_path_global_shm` seems too vague to me).

Also, please follow the rest of the code conventions in always returning `sr_error_info_t *`, NULL on success and the check can be together with the function call for these short prototypes.

Regards,
Michal

rjarry:
Ok I can try that however, having 2 separate prefixes means we need to export two variables. It seems a bit tedious to me. Why not use a single variable and append `sub` for subscriptions?

michalvasko:
Okay, then the format can be `PREFIX(sub)_NAME` meaning with the default prefix `sr` you get `sr_NAME` and `srsub_NAME` (just like it is now). Meaning the underscore will be hard-coded and not part of the prefix.

rjarry:
Hi Michal, is that ok like this?

src/common.c
```diff
@@ -1418,16 +1418,81 @@ sr_create_module_imps_incs_r(const struct lys_module *ly_mod)
...
+    if (err_info) {
+        goto end;
+    }
+
+    if (asprintf(path, "/%s_main", prefix) == -1) {
+        SR_ERRINFO_MEM(&err_info);
+        *path = NULL;
+    }
+
+end:
```
michalvasko, line 1454: Please do not use redundant labels, this one is not needed at all.

src/common.c
```diff
@@ -1418,16 +1418,81 @@ sr_create_module_imps_incs_r(const struct lys_module *ly_mod)
...
+    if (err_info) {
+        goto end;
+    }
+
+    if (asprintf(path, "/%s_ext", prefix) == -1) {
+        SR_ERRINFO_MEM(&err_info);
+        *path = NULL;
+    }
+
+end:
```
michalvasko, line 1474: Another redundant label.

src/common.c
```diff
@@ -1418,16 +1418,81 @@ sr_create_module_imps_incs_r(const struct lys_module *ly_mod)
...
     int ret;
 
+    err_info = sr_shm_prefix(&prefix);
+    if (err_info) {
+        return err_info;
+    }
+
     if (suffix2 > -1) {
-        ret = asprintf(path, "%s/srsub_%s.%s.%08x", abs_path ? SR_SHM_DIR : "", mod_name, suffix1, (uint32_t)suffix2);
+        ret = asprintf(path, "%s/%ssub_%s.%s.%08x",
```
michalvasko, line 1491: The convention I used is that in case a line is too long (more than 120 chars), I continue on the next line starting with double indentation (8 spaces). There are several similar places where this is different.

src/sysrepo.c
```diff
@@ -352,8 +352,17 @@ sr_connect(const sr_conn_options_t opts, sr_conn_ctx_t **conn_p)
         sr_conn_free(conn);
         if (created) {
             /* remove any created SHM so it is not considered properly created */
-            shm_unlink(SR_MAIN_SHM);
-            shm_unlink(SR_EXT_SHM);
+            char *shm_name = NULL;
+            if (sr_path_main_shm(&shm_name) == NULL) {
```
michalvasko, line 356: Please use `!` instead of `== NULL` just like in the rest of the code.

rjarry: done

src/common.c
rjarry: done

src/common.c
rjarry: I removed it.

src/common.c
rjarry: I removed it.

rjarry:
@michalvasko I have no issue in enforcing coding style consistency, don't worry :)

I think I have addressed all your comments. Please tell me if I need to change something else.

Maybe we could set up some sort of automatic coding style checks (with [clang-format](https://clang.llvm.org/docs/ClangFormat.html) for example), what do you think?

src/sysrepo.c
```diff
@@ -352,8 +352,17 @@ sr_connect(const sr_conn_options_t opts, sr_conn_ctx_t **conn_p)
         sr_conn_free(conn);
         if (created) {
             /* remove any created SHM so it is not considered properly created */
-            shm_unlink(SR_MAIN_SHM);
-            shm_unlink(SR_EXT_SHM);
+            char *shm_name = NULL;
+            if (!sr_path_main_shm(&shm_name)) {
```
michalvasko, line 356: Actually, I did not realize that this is wrong because in the unlikely event that this fails, the `err_info` will leak. In cases when the error should not be propagated, you should just free it, it is printed when it is generated.

src/sysrepo.c
```diff
@@ -352,8 +352,17 @@ sr_connect(const sr_conn_options_t opts, sr_conn_ctx_t **conn_p)
...
+                shm_unlink(shm_name);
+                free(shm_name);
+                shm_name = NULL;
+            }
+            if (!sr_path_ext_shm(&shm_name)) {
```
michalvasko, line 361: Similar leak.

src/sysrepo.c
rjarry: right, I'll fix that

michalvasko:
Well, we have some automatic code formatting in our wishlist for quite some time now but not having a high priority, it is consistently postponed. But if you can help us with it, we would appreciate it.

Regards,
Michal

src/sysrepo.c
rjarry: Can I use sr_errinfo_merge ?

michalvasko: Yes, I only looked just now that it is executed only when some error occurred so I guess it makes more sense to also return it.

rjarry: I should be OK now.

rjarry:
> Well, we have some automatic code formatting in our wishlist for quite some time now but not having a high priority, it is consistently postponed. But if you can help us with it, we would appreciate it.

I'll see if I find some time.

src/common.c
```diff
@@ -1418,16 +1418,79 @@ sr_create_module_imps_incs_r(const struct lys_module *ly_mod)
...
+{
+    sr_error_info_t *err_info = NULL;
+
+    *prefix = getenv(SR_SHM_PREFIX_ENV);
+    if (*prefix == NULL) {
+        *prefix = SR_SHM_PREFIX_DEFAULT;
+    } else if (strchr(*prefix, '/') != NULL) {
+        *prefix = NULL;
+        sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, NULL,
+            SR_SHM_PREFIX_ENV" cannot contain slashes.");
```
michalvasko, line 1432: One last formatting problem.

src/sysrepo.c
```diff
@@ -352,8 +352,22 @@ sr_connect(const sr_conn_options_t opts, sr_conn_ctx_t **conn_p)
...
-            shm_unlink(SR_MAIN_SHM);
-            shm_unlink(SR_EXT_SHM);
+            sr_error_info_t *tmp_err = NULL;
+            char *shm_name = NULL;
+            if ((tmp_err = sr_path_main_shm(&shm_name))) {
+                sr_errinfo_merge(&err_info, tmp_err);
+            } else {
+                shm_unlink(shm_name);
+                free(shm_name);
+                shm_name = NULL;
```
michalvasko, line 362: Seems that this line is not needed.

src/sysrepo.c
```diff
@@ -352,8 +352,22 @@ sr_connect(const sr_conn_options_t opts, sr_conn_ctx_t **conn_p)
...
+            }
+            if ((tmp_err = sr_path_ext_shm(&shm_name))) {
+                sr_errinfo_merge(&err_info, tmp_err);
+            } else {
+                shm_unlink(shm_name);
+                free(shm_name);
+                shm_name = NULL;
```
michalvasko, line 369: Same as above.

src/sysrepo.c
rjarry: What line exactly? If `sr_path_main_shm` failed, `shm_name` is `NULL`.

src/common.c
rjarry: done

src/sysrepo.c
michalvasko: But it does not matter whether `shm_name` is NULL or not in the code. I would just like to keep this code short because it is only error handling.

rjarry: I don't understand what I need to change here.

If `shm_name` is `NULL`, `shm_unlink` should not be called. If it is not `NULL`, it must be freed after calling `shm_unlink`. Did you have something else in mind?

michalvasko: We seem not to understand each other, all I am saying is that the code `shm_name = NULL;` seems redundant to me and can be removed. If you want to keep it consistent then I would make sure the functions either allocate `shm_name` on success or set it to `NULL` on error. Maybe even `shm_name` initialization can be removed but since it is not taking up another line, I leave that up to you.

rjarry: We did not understand each other, indeed :) I'll remove that.

rjarry:
The checks on ppc64le are failing because of some lock timeout. I'm not sure it is related to my changes here. However could it be related to #2033?

michalvasko:
Well, the timeouts are because of the unhandled locks, yes. But the problem is caused by some other, not intentional, lock problem and it is not related to your changes. Not sure if I can fix it, though, because it seems to depend on some data race as well.

Regards,
Michal

rjarry:
Thanks Michal.

Issue #1207, opened 2018-07-04
subscription conn closed when cl_sm_conn_in_buff_process() excute failed


chunlovemail:
### Things you may try first
* [x] Did you check if this is a duplicate issue?
* [x] Did you test it on the latest sysrepo devel branch?

### Description

1. Process1 use sr_dp_get_items_subscribe() function to subscribe some state value
2. Process2 use sr_get_item() to  get the state data. Occasionally, the following problem arise in Process1.

```
sysrepo-CSManager[1661]: [ERR] (cl_sm_conn_msg_process:1061) Unable to unpack the message (conn=0x7fcd3c001370).
sysrepo-CSManager[1661]: [ERR] (cl_sm_conn_in_buff_process:1132) Error by processing of the message.
sysrepo-CSManager[1661]: [WRN] (cl_sm_fd_read_data:1216) Error by processing of the input buffer of fd=13, closing the connection.
```
3. In sysrepo/src/clientlib/cl_subscription_manager.c file, cl_sm_fd_read_data() function.
It seems that Process1 has close the subscription conn between Process1 and sysrepo. 
```
if (SR_ERR_OK == rc) {
        rc = cl_sm_conn_in_buff_process(sm_ctx, conn);
        if (SR_ERR_OK != rc) {
            SR_LOG_WRN("Error by processing of the input buffer of fd=%d, closing the connection.", conn->fd);
            conn->close_requested = true;
        }
    }

    /* close the connection if requested */
    if (conn->close_requested) {
        cl_sm_conn_close(sm_ctx, conn);
        rc = SR_ERR_DISCONNECT;
    }
```
4. As a result, the Process1 won't recv any message from sysrepo when Process2 get state data.
Finally, In Process2, sr_get_item() returns "Timeout has expired"

**Expected behavior:**
1. When this problem arise, how can Process1 to check the subscription is normal? For the  subscription conn  is a internal implementation mechanism, Process1 doesn't know the conn is closed.
2. When "Unable to unpack the message" occurs, can we did not call function cl_sm_conn_close(sm_ctx, conn)? 

### Versions
Sysrepo version: `sysrepo daemon, version 0.7.4`.
libyang version: `yanglint 0.15.131`

Thanks!

chunlovemail:
Hi Michal,
    The output is as shown below:
```
APPS[1857]: [OPT][command][cmd_control.c:389] Get pd power +++++
[...]

PR #2067, opened 2020-07-27
Fix issue #1207

gunterw:


michalvasko:
Thanks for debugging it, it may help several people still using this legacy version.

Regards,
Michal

PR #2072, opened 2020-07-29
Various improvements in the sysrepo examples

gkatsikas:
Signed-off-by: Georgios Katsikas <katsikas.gp@gmail.com>

### DIFF
--- a/examples/rpc_send_example.c
+++ b/examples/rpc_send_example.c
@@ -118,8 +118,7 @@ main(int argc, char **argv)
     sr_conn_ctx_t *connection = NULL;
     sr_session_ctx_t *session = NULL;
     int rc = SR_ERR_OK;
-    sr_val_t *output = NULL;
-    size_t i, output_count = 0;
+    size_t i = 0;
     const char *path;
 
     if (argc != 2) {
@@ -145,8 +144,16 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
+    /* RPC input */
+    sr_val_t *input = NULL;
+    const size_t input_cnt = 0;
+
+    /* RPC output */
+    sr_val_t *output = NULL;
+    size_t output_count = 0;
+
     /* send the RPC */
-    rc = sr_rpc_send(session, path, NULL, 0, 0, &output, &output_count);
+    rc = sr_rpc_send(session, path, &input[0], input_cnt, 0, &output, &output_count);
     if (rc != SR_ERR_OK) {
         goto cleanup;
     }
--- a/examples/rpc_send_example.c
+++ b/examples/rpc_send_example.c
@@ -144,16 +144,12 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
-    /* RPC input */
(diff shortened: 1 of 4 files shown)


examples/rpc_send_example.c
```diff
@@ -145,8 +144,16 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
+    /* RPC input */
+    sr_val_t *input = NULL;
+    const size_t input_cnt = 0;
```
michalvasko, line 149: I do not think using such empty structures is an improvement because it may easily be misinterpreted in the way that it is required even though it is not. I would prefer you keep it the way it was.

examples/sr_get_items_example.c
```diff
@@ -131,13 +132,18 @@ main(int argc, char **argv)
...
+        } else if (!strcmp(argv[2], "startup")) {
+            ds = SR_DS_STARTUP;
+        } else if (!strcmp(argv[2], "candidate")) {
+            ds = SR_DS_CANDIDATE;
         } else {
             printf("Invalid datastore %s\n", argv[2]);
             return EXIT_FAILURE;
         }
     }
+    op_str = argv[2] ? argv[2] : "running";
```
michalvasko, line 144: I do not think this is safe, better use `argc > 2 ? argv[2] : "running"`.

examples/rpc_send_example.c
gkatsikas: Done

examples/sr_get_items_example.c
gkatsikas: Done

examples/sr_get_items_example.c
```diff
@@ -131,13 +132,18 @@ main(int argc, char **argv)
             ds = SR_DS_RUNNING;
         } else if (!strcmp(argv[2], "operational")) {
             ds = SR_DS_OPERATIONAL;
+        } else if (!strcmp(argv[2], "startup")) {
+            ds = SR_DS_STARTUP;
+        } else if (!strcmp(argv[2], "candidate")) {
+            ds = SR_DS_CANDIDATE;
```
michalvasko, line 138: Have not noticed this, please add the new datastores to the brief usage on line 126.

gkatsikas: Right, I missed that, thanks!

examples/rpc_send_example.c
```diff
@@ -145,6 +144,10 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
+    /* RPC output */
+    sr_val_t *output = NULL;
+    size_t output_count = 0;
+
```
michalvasko, line 150: Seems there is a problem with you moving this definition because these variables are used in `cleanup`, which can be jumped to even before they are defined. Please keep the definition where it was.

gkatsikas: Done. I never use goto, thus my confusion

michalvasko: One should not overuse it, but it is often handy especially for error handling.

Issue #2108, opened 2020-08-26
Startup Datastore copied over Running after module load


imiller0:
When new modules are loaded into sysrepo the startup datastore is copied to create the running datastore.  As I understand it this should be a one-time per module copy and that the running datastore, once created, is no longer overwritten unless an explicit copy_config is performed.  However, in the following sequence the running datastore for all modules is overwritten by the startup:

1. Connect to sysrepo
1. Load module A into sysrepo
1. Set items in startup and apply
1. Disconnect and re-connect (copy from startup to running happens)
1. Set items in model A in running and apply
1. Load module B into sysrepo
1. Disconnect all connections so that B is loaded
1. Connect and get modified items from model A in running -- The changes made in step 5 have been overwritten when the startup of model A is copied over running a second time.

imiller0:
Fix merged.  Thank you!

PR #2109, opened 2020-08-26
Proposed bug fix for issue #2108

imiller0:
Included is a test module RunningDSTest.py which demonstrates the loss (overwrite) of data in the RUNNING datastore when the STARTUP is copied again.  

The proposed fix is in shm_main.c and uses the absolute filesystem path for testing sr_file_exists.  This ensures that the existence of the running datastore (file) is correctly determined.  The path is then changed back to the relative path for use in the copy.

The NotificationTest appears to depend on the extra copy from startup to running, so I added an explicit copy to TestModule to fix that test against the new behavior.


michalvasko:
Hi,
I believe you are right, this is a bug. The fix seems correct, thanks, merging.

PR #2110, opened 2020-08-27
python binding BUGFIX Should be taking global lock before calling back to python code

johnbeckettn2e:
Should have python global lock before executing function `ly_module_imp_clb`, as this calls python code and handles python reference counts


michalvasko:
Hi,
I do not know much about these Python bindings but what you are saying makes sense, merged.

PR #2112, opened 2020-08-28
docker REFACTOR reorganize dockerfiles

jsmolic:
- Add sysrepo Dockerfile along with sysrepo-netopeer2
- Add legacy and latest image for sysrepo and sysrepo-netopeer2


deploy/docker/sysrepo/latest/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+      apt-get install libyang1 libyang-dev
+
+# sysrepo
+RUN \
+      cd /opt/dev && \
+      git clone https://github.com/sysrepo/sysrepo.git && cd sysrepo && \
+      mkdir build && cd build && \
+      cmake -DENABLE_TESTS=OFF -DCMAKE_BUILD_TYPE="Release" -DCMAKE_INSTALL_PREFIX:PATH=/usr .. && \
+      make -j2 && \
+      make install
```
michalvasko, line 66: You can download it from the same repository as *libyang*.

michalvasko (commented):
I cannot directly comment the code, but in `latest` `netopeer2` `Dockerfile` you should be able to just download all the packages from the repo.

deploy/docker/sysrepo/latest/Dockerfile
jsmolic: Great, thanks :)

I tried to install `libnetconf2` from there, but I got
```
The following packages have unmet dependencies:
 libnetconf2 : Depends: libyang but it is not installable
E: Unable to correct problems, you have held broken packages.
```
Looking at https://download.opensuse.org/repositories/home:/liberouter/xUbuntu_18.04/Packages, should `libnetconf2` maybe depend on `libyang1` instead of `libyang` as `libyang1` is the name of the package?

michalvasko: Oh, the latest packages of *libnetconf2* are not built on Ubuntu 18.04, that is the reason for the dependency problem. Never mind then, it must be installed from source. Please include a comment about why it is not installed from a package (just something generic is fine, that packages are not available for the specific distro).

jsmolic: Thanks, done!

PR #2116, opened 2020-09-01
fix log macros to allow empty variable arguments

hthiery:
With this the SRP_LOG_xxxMSG macros aren't needed anymore.

see: https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

Signed-off-by: Heiko Thiery <heiko.thiery@gmail.com>


hthiery:
the provided change seems not to work on the used compiler in travis

hthiery:
ups ... there seem to be no portable solution for variadic macros with zero arguments

michalvasko:
Hi,
yep, I have already gone through the same process as you just now and have reached the same conclusion.

hthiery:
ok .. sorry for the noise

PR #2119, opened 2020-09-01
fix log macros to allow empty variable arguments

hthiery:
With this the SRP_LOG_xxxMSG macros aren't needed anymore.

see: https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

Signed-off-by: Heiko Thiery <heiko.thiery@gmail.com>


hthiery:
@michalvasko could you please take a look on that .. I was not happy with my previous conclusion.

"There should be a better way" ;-)

michalvasko:
Hi,
okay, but now you can call the macro with no arguments, right? Not sure what that would do, probably nothing good.

hthiery:
Yes, you're right. The macro can be called without arguments. But I think then the compiler/preprocessor should throw an error when the macro is used in that case. 

I tried this
```
#include <stdio.h>

#define PRINTF_ERR(...) fprintf(stderr, __VA_ARGS__)

int main(void)
{
	PRINTF_ERR();
	return 0;
}
```

and the output is:

```
test.c: In function ‘main’:
test.c:3:52: error: expected expression before ‘)’ token
 #define PRINTF_ERR(...) fprintf(stderr, __VA_ARGS__)
                                                    ^
test.c:7:2: note: in expansion of macro ‘PRINTF_ERR’
  PRINTF_ERR();
  ^~~~~~~~~~
```

michalvasko:
Hi,
okay then, I suppose. So please create a pull request for the current `devel` and it can be merged.

hthiery:
OK. Should I close this PR  and create a nee one? Or do you accept with the merge commit?

PR #2130, opened 2020-09-04
sysrepo BUGFIX fix undefined behavior with qsort

syyyr:
Fixes UB in https://github.com/CESNET/netopeer2/issues/712


syyyr:
I found out that the undefined behavior appears in all of the qsort calls in shm_mod.c so I applied the same fix to all of them. I'm not sure if you're fine with this, or if I should refactor this into some sort of a function? I'm fine with changing this to your liking.

PR #2136, opened 2020-09-09
tests BUGFIX increase timeout in test_process

syyyr:
Under TSan, this test sometimes hangs on wait(), because the forked
process never finishes and spins on some sort an init and/or cleanup
procedure, that I'm not able to debug. This is reproducible consistently
on our CI (a heavily multi-threaded environment). On my work laptop, I
was able to reproduce this by setting CPU clock speed to a low value.
TSan creates a huge overhead (and slows the forking process), which
seems to be mandatory for this issue to happen. Lowering the timeout to
a very low value (like 1) also increases the chance this appears.

This is definitely not a proper fix as I'm not sure what exactly is the
reason this happens (it could be a deeper problem), but increasing the
timeout solves it for our CI (or at least most of the time, I haven't
gotten the issue since I increased the timeout).

The test now runs for two seconds longer because of this, but
that shouldn't be a problem since the patch that enables the tests to be
run in parallel.

I also am not sure about the commit message title.


michalvasko:
I have nothing against merging this and I think it makes sense. However, for this test obviously every test case runs in 2 processes and in case of parallel execution, it can happen that the scheduler will not continue execution of the other process that should handle the event and the original process will still time out. No synchronization is possible in this specific use-case either. So, maybe running all the tests in parallel is not such a good idea after all.

syyyr:
Thank you for merging. 

> So, maybe running all the tests in parallel is not such a good idea after all.

You mean as in, `ctest -j10`? I could reproduce this even in my dev environment without the `-j` flag and no other sysrepo processes running. Although I did have to reduce the timeout to a very small number and also slow down my computer (and also enable TSan).

michalvasko:
Well, like I said, the test depends on some concurrency and since direct synchronization cannot be used, there will always be a slow enough machine (for whatever reason) that will time out.

PR #2165, opened 2020-10-01
Replace SRP_LOGxxxMSG in example

hthiery:


michalvasko:
I plan to remove the `*MSG` macros altogether but since it will be a non-backwards compatible change, I wanted to wait until there are more such changes. I do not think it is critical. So, I would probably prefer to leave it the way it is for now but if you feel strongly about this, I do not have a problem with merging it.

hthiery:
I do not have a problem with that ... but thought it is good to change the example to show the "new" way of using the log mechanism. That the "old" fashion is still there is ok for me.

What about wrapping the old macros with deprecation warnings?

michalvasko:
Well, I am hoping to remove them fairly soon (month or so) so I was hoping it could stay the way it is

hthiery:
no problem for me

michalvasko:
Fine.

PR #2166, opened 2020-10-01
C++: Rework subcriptions

syyyr:
Detailed info is in the commit message. Naming (especially `DelegatedSubscriptions`) can/should be changed.

In case you like the new std::function callbacks, I could also change the current `sysrepo::Subscribe` to use that (instead of overriding `sysrepo::Callback`).

cc @jktjkt

### DIFF
--- a/bindings/cpp/src/Session.cpp
+++ b/bindings/cpp/src/Session.cpp
@@ -43,7 +43,7 @@ Session::Session(S_Connection conn, sr_datastore_t datastore)
     int ret;
     _conn = nullptr;
     _sess = nullptr;
-    S_Deleter deleter(new Deleter(_sess));
+    auto deleter = std::make_shared<Deleter>(_sess);
 
     /* start session */
     ret = sr_session_start(conn->_conn, datastore, &_sess);
@@ -97,7 +97,7 @@ void Session::session_notif_buffer()
 
 S_Errors Session::get_error()
 {
-    S_Errors errors(new Errors());
+    auto errors = std::make_shared<Errors>();
 
     sr_get_error(_sess, &errors->_info);
     if (errors->_info == nullptr) {
@@ -149,7 +149,7 @@ libyang::S_Context Session::get_context()
 
 S_Val Session::get_item(const char *path, uint32_t timeout_ms)
 {
-    S_Val value(new Val());
+    auto value = std::make_shared<Val>();
 
     int ret = sr_get_item(_sess, path, timeout_ms, &value->_val);
     if (SR_ERR_OK == ret) {
@@ -164,7 +164,7 @@ S_Val Session::get_item(const char *path, uint32_t timeout_ms)
 
 S_Vals Session::get_items(const char *xpath, uint32_t timeout_ms, const sr_get_oper_options_t opts)
 {
-    S_Vals values(new Vals());
+    auto values = std::make_shared<Vals>();
 
     int ret = sr_get_items(_sess, xpath, timeout_ms, opts, &values->_vals, &values->_cnt);
     if (SR_ERR_OK == ret) {
@@ -315,7 +315,7 @@ void Session::unlock(const char *module_name)
 
 S_Iter_Change Session::get_changes_iter(const char *xpath)
 {
-    S_Iter_Change iter(new Iter_Change());
+    auto iter = std::make_shared<Iter_Change>();
 
     int ret = sr_get_changes_iter(_sess, xpath, &iter->_iter);
     if (SR_ERR_OK == ret) {
@@ -329,7 +329,7 @@ S_Iter_Change Session::get_changes_iter(const char *xpath)
 
 S_Iter_Change Session::dup_changes_iter(const char *xpath)
 {
-    S_Iter_Change iter(new Iter_Change());
+    auto iter = std::make_shared<Iter_Change>();
 
     int ret = sr_dup_changes_iter(_sess, xpath, &iter->_iter);
     if (SR_ERR_OK == ret) {
@@ -343,7 +343,7 @@ S_Iter_Change Session::dup_changes_iter(const char *xpath)
 
 S_Change Session::get_change_next(S_Iter_Change iter)
 {
-    S_Change change(new Change());
+    auto change = std::make_shared<Change>();
 
     int ret = sr_get_change_next(_sess, iter->_iter, &change->_oper, &change->_old, &change->_new);
     if (SR_ERR_OK == ret) {
@@ -357,7 +357,7 @@ S_Change Session::get_change_next(S_Iter_Change iter)
 
 S_Tree_Change Session::get_change_tree_next(S_Iter_Change iter)
 {
-    S_Tree_Change change(new Tree_Change());
+    auto change = std::make_shared<Tree_Change>();
 
     int ret = sr_get_change_tree_next(_sess, iter->_iter, &change->_oper, &change->_node, &change->_prev_value, \
             &change->_prev_list, &change->_prev_dflt);
@@ -374,7 +374,7 @@ Session::~Session() {}
 
 S_Vals Session::rpc_send(const char *path, S_Vals input, uint32_t timeout_ms)
 {
-    S_Vals output(new Vals());
+    auto output = std::make_shared<Vals>();
 
     int ret = sr_rpc_send(_sess, path, input->_vals, input->_cnt, timeout_ms, &output->_vals, &output->_cnt);
     if (ret != SR_ERR_OK) {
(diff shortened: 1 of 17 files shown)


syyyr:
I'll also probably write some tests for this: I'm not sure if this implentation is correct. I posted this here mainly for discussion

michalvasko:
Hi,
firstly, let me start by saying that the C++ bindings are open for (major) modifications because with the migration to *libyang2*, there will be new *libyang* C++ bindings, which then must be used in the *sysrepo* C++ bindings. So, with that there will certainly be at least some refactoring done. However, these C++ bindings are not a priority so I am not sure when it will all take place, probably some time next year. But in the meantime the current C++ bindings can be improved, the changes will not be lost because they will be sued as the basis for the new bindings.

Having said that, I have nothing against changing how the current subscriptions work and having only a single class for them, seems like the cleaner solution since backwards compatibility is not that much of an issue.

A final note, while it is true that you cannot create an empty subscription structure in C, you can get one if you then call only `sr_disconnect()` and not `sr_unsubscribe()`. You then have a valid subscription structure (which can be used as the subscription context for later subscriptions) that does not listen for any events.

syyyr:
I changed my mind, for a test I would have to write the infrastructure for testing c++ and also write a custom event loop, which seems like it would take some time. However, I'm testing this in our app which uses this (and it has already found a bug in https://github.com/sysrepo/sysrepo/pull/2171).

michalvasko:
Okay, I suppose, but I would still like there not to be redundant classes and this one seems that way. If it is to be merged, please keep all the required functionality in a single subscription class.

syyyr:
Hi, I merged my "DelegatedSubscription" into the Subcribe class. I reworked the entire API to a more C++11 style. I have yet to try it in my app, but I posted it here for discussion. Let me know what you think.

I also included some cleanup commits.

michalvasko:
In general it looks good to me but I am wondering, was it really necessary to make all the structures shared pointers? Even the local variables in the examples?

syyyr:
Are you referring to the "get rid of new" commit? That commit doesn't change variables into shared pointers, all of them already were shared pointers, it's just that it's a good practice to create them with make_shared instead of new.

By the way, I think the API can be improved a little bit more, I'll be discussing that with @jktjkt today, and then I will post the results.

michalvasko:
I did not look at separate commits but the whole PR. But you are right, the current API requires them to be pointers so never mind.

And good, l am sure there are more improvements that can be done.

bindings/cpp/src/Session.cpp
```diff
@@ -439,8 +454,26 @@ Subscribe::~Subscribe()
     }
 }
 
+void Subscribe::check_custom_loop_options(sr_subscr_options_t opts)
+{
+    if (uses_custom_loop && !(opts & SR_SUBSCR_NO_THREAD)) {
```
jktjkt, line 459: let's check this the other way round as well -- if the ctor that created this was not for a custom event loop, error out if `SR_SUBSCR_NO_THREAD` is enabled

bindings/cpp/src/Session.cpp
```diff
@@ -439,8 +454,26 @@ Subscribe::~Subscribe()
...
+        reg_called = true;
+    }
+}
+
 void Subscribe::module_change_subscribe(const char *module_name, ModuleChangeCb cb, const char *xpath, uint32_t priority, sr_subscr_options_t opts)
 {
+    check_custom_loop_options(opts);
     module_change_cbs.emplace_back(cb);
 
     opts |= SR_SUBSCR_CTX_REUSE;
```
jktjkt, line 479: I understand that's how the old code worked, but it's different from how `SR_SUBSCR_NO_THREAD` is enforced. My suggestion is to make this consistent, meaning to error out unless `SR_SUBSCR_CTX_REUSE` is passed by the user.

bindings/cpp/src/Session.hpp
```diff
@@ -208,6 +209,14 @@ class Subscribe
 
     S_Session sess;
     S_Deleter sess_deleter;
+
+    bool uses_custom_loop;
```
jktjkt, line 213: The "with event loop" constructor takes a pair of two `std::functions`, which will get called later on, but their validity is not checked. I propose to check that they are defined, and use that in place of `uses_custom_loop`, it's gonna be simpler that way.

bindings/cpp/src/Session.hpp
```diff
@@ -191,47 +168,45 @@ class Subscribe
...
-    void event_notif_subscribe_tree(const char *module_name, S_Callback callback, const char *xpath = nullptr, \
-            time_t start_time = 0, time_t stop_time = 0, void *private_data = nullptr, sr_subscr_options_t opts = SUBSCR_DEFAULT);
+    void event_notif_subscribe_tree(const char *module_name, EventNotifTreeCb cb, const char *xpath = nullptr, time_t start_time = 0, time_t stop_time = 0, sr_subscr_options_t opts = SUBSCR_DEFAULT);
     /** Wrapper for [sr_oper_get_items_subscribe](@ref sr_oper_get_items_subscribe) */
-    void oper_get_items_subscribe(const char *module_name, const char *path, S_Callback callback, \
-            void *private_data = nullptr, sr_subscr_options_t opts = SUBSCR_DEFAULT);
-    std::vector<S_Callback > cb_list;
+    void oper_get_items_subscribe(const char *module_name, OperGetItemsCb cb, const char *path, sr_subscr_options_t opts = SUBSCR_DEFAULT);
 
     /** Wrapper for [sr_get_event_pipe](@ref sr_get_event_pipe) */
```
jktjkt, line 185: this method can be removed now; in case an evil user really needs this, they can use that private API (`swig_sub()`)

jktjkt (changes requested):
I'm happy with the general approach, just some nitpicks below.

bindings/cpp/src/Session.hpp
syyyr: done

bindings/cpp/src/Session.hpp
syyyr: done

bindings/cpp/src/Session.cpp
syyyr: done

jktjkt (changes requested):
The CI is failing because the Python bindings (and I also assume the example code) were not updated with the API changes.

syyyr:
Okay, so the tests were failing because I removed the unsubscribe method from the bindings. Now I have also removed it from the tests, unsubscription should still happen from the destructor. Funnily enough, there are actually TWO subscribe.unsubscribe calls in the tests, but the other one doesn't seem to make any problems (so I've preserved it).

What do you think about this? If I understand correctly, what the python bindings do is extend Subscribe with new overloads for subscription methods that take python callbacks is arguments. That means that the sub API rework didn't really affect Python (apart from the unsub method). The main reason of the rework was to get rid of sysrepo::Callback in favor of the more flexible std::function. In Python, you can directly use functions, which is fine (by me).

The internal implemention could be reworked though, the current state is that there are two implementations of all the subcribe methods (I'm not sure how much this can be improved). Also `private_date` is available even though you can use captures in Python (the same as in C++, where I've gotten rid of private_data).

michalvasko:
You do not have to concern yourself with these C++ based SWIG Python bindings, they should be deprecated fairly soon. It is enough if the current functionality of them is preserved.

jktjkt:
Just a suggestion on the Python bindings -- we had quite a nice experience with [pybind11](https://github.com/pybind/pybind11) ([docs](https://pybind11.readthedocs.io/en/stable/)). For example, [here are Python bindings for our C++ NETCONF-talking wrapper](https://gerrit.cesnet.cz/plugins/gitiles/CzechLight/netconf-cli/+/refs/heads/master/src/python_netconf.cpp), and here's [an example on how to use them from the Python code](https://gerrit.cesnet.cz/plugins/gitiles/CzechLight/netconf-cli/+/refs/heads/master/tests/python_netconfaccess.py). It's a part of a basic smoke test of the bindings themselves, so it's not terribly idiomatic Python.

Anyway, just sharing my favorite SW tools. That reminds me that we should contribute our [C++ bindings for libnetconf2](https://gerrit.cesnet.cz/plugins/gitiles/CzechLight/netconf-cli/+/refs/heads/master/src/netconf-client.cpp) upstream; yup, we'll do that once they get cleaned up, which depends on some libssh improvements, etc.

jktjkt (approved):
I went through these changes, and I think they make sense, thanks. They also make the CESNET-internal code easier to follow, IMHO, so good job, @syyyr :).

michalvasko:
> Just a suggestion on the Python bindings

Thanks, but we are simply not able to maintain all these bindings so we have taken up the offer of @rjarry and his Python bindings implementation using CFFI. They can already be used and are available at https://github.com/sysrepo/sysrepo-python. They will be the main Python bindings after the current ones are deprecated.

> That reminds me that we should contribute our C++ bindings for libnetconf2 upstream

Sounds great, like I said, we just do not have the people for all this.

So, these C++ improvements are appreciated and can be merged it seems, thanks!

PR #2171, opened 2020-10-06
c++ BUGFIX oper_data output

syyyr:
If subscribing to a list type top-level node, all siblings need to be
duplicated.

Also, I deleted a line which was (probably) supposed to create a nullptr
shared_ptr, but instead created a valid libyang::Data_Node with an
invalid lyd_node inside. That broke the duping algorithm if the `parent`/`tree`
variable wasn't reassigned inside the callback.

Sorry, I don't have a test, but I found out about this when porting to new sysrepo.

### DIFF
--- a/bindings/cpp/src/Session.cpp
+++ b/bindings/cpp/src/Session.cpp
@@ -473,10 +473,9 @@ static int oper_get_items_cb(sr_session_ctx_t *session, const char *module_name,
         tree = std::make_shared<libyang::Data_Node>(*parent);
         ret = wrap->oper_get_items(sess, module_name, path, request_xpath, request_id, tree, wrap->private_data["oper_get_items"]);
     } else {
-        tree = std::make_shared<libyang::Data_Node>(nullptr);
         ret = wrap->oper_get_items(sess, module_name, path, request_xpath, request_id, tree, wrap->private_data["oper_get_items"]);
         if (tree) {
-            *parent = lyd_dup(tree->swig_node(), LYD_DUP_OPT_RECURSIVE);
+            *parent = lyd_dup_withsiblings(tree->swig_node(), LYD_DUP_OPT_RECURSIVE);
         }
     }
     return ret;


michalvasko:
Okay, I may be hitting my C++ knowledge limits here. When you remove the `tree` assignment, what exactly is left in it? Is it a proper shared pointer, just without value? Because even though it is passed by reference to the callback, it must be valid because `reset()` method is supposed to be called on it. As for the duplication, you are always supposed to provide *subtrees* and that is true for top-level subscriptions as well so the fix should be redundant.

syyyr:
In the C api, the callback argument is `lyd_node**`. In the C++ bindings, this translates to the reference, which is the outer indirection and the shared_ptr, which is the inner indirection (`lyd_node*`). If I understand the callbacks correctly, the inner indirection can be `NULL`, when working with top-level nodes. That translates to C++ as a `nullptr` of type `libyang::Data_Node`. That means that the shared_ptr has to be initialized with `nullptr` (or default constructed, it's the same behavior). The reference itself will always point to a valid shared_ptr object (reference can't be NULL). So, it is still valid to call `reset()` on it, if you want to (or reassign it or do whatever you want to and it will propagate back to the C api).

What the previous code did was not initialize a shared_ptr with a null, but construct a `libyang::Data_Node` with a `nullptr` `lyd_node*` inside it. `make_shared` constructs objects (as opposed to assigning pointer values).

About the duplication: what if I'm subscribed to a top-level list? If I return more instances, all of them the same list, but different instances, then there won't be any subtrees (this is something that we do in our app).

michalvasko:
> The reference itself will always point to a valid shared_ptr object (reference can't be NULL).

Of course, but if I applied the C initialization rules, `tree` would not be `nullptr` but some garbage instead and if you called `reset()` on it, it would most likely crash. That is obviously not true for C++, which I thought so but wanted to make sure.

> what if I'm subscribed to a top-level list?

Ideally, you should not need to do this because it is not a good practice. Moreover, *libyang* is not able to use hashes for such lists (or leaf-lists) so in case there were many such instances expect severely degraded performance. However, it is a valid use-case and seems to be supported in C code, so okay. I have mentioned this use-case in `sysrepo.h` as well.

So okay, this can be merged, thanks.

syyyr:
>That is obviously not true for C++, which I thought so but wanted to make sure.
I think there is some value in initializing with `nullptr`, for clarity, but the problem was using std::make_shared vs shared_ptr's constructor. I removed that line not because it was redundant, but because it resulted in a bug. Just wanted to point that out.

Thanks for merging!

PR #2177, opened 2020-10-10
Duplicate package of "apt-get install" in Dockerfile

EzekielTsui:
Delete duplicate package of 'apt-get install'. ( **vim** and **build-essential** )


michalvasko:
Right, thanks.

PR #2184, opened 2020-10-15
sysrepo DOC add plugin development documentation

jvijtiuk:
This adds an initial version of plugin development documentation in plugin_dev_doc.md.

The goal of the document is to provide an introduction to sysrepo plugin development, with an overview of the required background knowledge, like YANG and NETCONF so that someone unfamiliar with the process can start developing plugins.

If there are any comments or changes that should be made, please let me know.

Thanks,
Juraj


michalvasko:
It seems fine to me but at the beginning there is quite specific information about versions and the current state of development. This may easily lead to outdated information so we must be careful about that.

jvijtiuk:
I agree, we could remove references to current versions. However it might be useful to have it all available in one place, so that someone who is just getting started with plugin development can find all the relevant information in one place. Do you have any suggestions on what we should do?

michalvasko:
Well, I am not that sure about the usefulness of that section at all. The current versions are always available here on GitHub so it should be easy enough for anyone interested to learn what they want. Still, information that will not age (such as the legacy versions) can remain.

jvijtiuk:
> The current versions are always available here on GitHub so it should be easy enough for anyone interested to learn what they want. Still, information that will not age (such as the legacy versions) can remain.

I've pushed a commit that removes the unnecessary information, and refers to the projects on GitHub, so I think this should be fine now.

michalvasko:
Okay, seems good now, merging.

PR #2195, opened 2020-10-26
docker FEATURE use libssh stable-0.9 branch in sysrepo-netopeer2 latest Dockerfile

jsmolic:


michalvasko:
Okay, checking-out a specific version is a reasonable approach but then I would opt for the latest 0.9.5, no reason not to anymore (since it has finally fixed all the damn regression bugs).

jsmolic:
@michalvasko Sure, that makes sense. I have added the appropriate changes

jsmolic:
Let's actually use `stable-0.9` branch

jsmolic:
Thanks

PR #2199, opened 2020-10-29
Do not let SR_SUBSCR_DONE_ONLY mask out SR_EV_ENABLED events

jktjkt:
The `SR_SUBSCR_DONE_ONLY` event is documented to control whether `SR_EV_CHANGE`/`SR_EV_ABORT` are sent. Since the rewrite, it used to also skip generating the initial `SR_EV_ENABLED` as well. That's surely a mistake because that one is supposed to be gated by the `SR_EV_ENABLED` flag.

Fixes: f36f312e ("sysrepo CHANGE add special enabled event")

### DIFF
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -3260,7 +3260,7 @@ sr_module_change_subscribe_running_enable(sr_session_ctx_t *session, const struc
     tmp_sess.ds = SR_DS_RUNNING;
     tmp_sess.dt[tmp_sess.ds].diff = enabled_data;
 
-    if (!(opts & SR_SUBSCR_DONE_ONLY)) {
+    if (opts & SR_SUBSCR_ENABLED) {
         tmp_sess.ev = SR_SUB_EV_ENABLED;
         SR_LOG_INF("Triggering \"%s\" \"%s\" event on enabled data.", ly_mod->name, sr_ev2str(tmp_sess.ev));
 
--- a/tests/test_operational.c
+++ b/tests/test_operational.c
@@ -814,7 +814,7 @@ test_enabled_partial(void **state)
 
     called = 0;
     ret = sr_module_change_subscribe(st->sess, "ietf-interfaces", "/ietf-interfaces:interfaces/interface[name='eth128']",
-            enabled_change_cb, &called, 0, SR_SUBSCR_ENABLED, &subscr);
+            enabled_change_cb, &called, 0, SR_SUBSCR_ENABLED | SR_SUBSCR_DONE_ONLY, &subscr);
     assert_int_equal(ret, SR_ERR_OK);
     assert_int_equal(called, 2);
 


michalvasko:
Actually, the way it is now is intentional. You are right that the `SR_EV_ENABLED` even is skipped but the following `SR_EV_DONE` with the same data is not. You are just not able to distinguish between the initial and later DONE events. It is implemented this way so that you have 2 events following each other just like in the other cases but maybe it would be better to use just one, I am not sure.

jktjkt:
I think that the use cases behind `SR_EV_ENABLED` and behind `SR_EV_CHANGE` are different from each other. The `SR_EV_CHANGE` is useful for two-phase commit where you want to have some application logic for checking if the proposed changes are actually OK, and possibly rejecting them even after the usual YANG validation has succeeded. In my scenario, I have no use for that (yet) because our YANG models express all the requires constraints via `must`, etc. So, we listen for the `SR_EV_DONE` and propagate that configuration state to the HW. We do not subscribe to `SR_EV_CHANGE`. We could have subscribed and (silently) ignore that event, sure.

On the other hand, we'd like to use the `SR_EV_ENABLED` to signal "yes, we're now up and running" to `systemd`. Without sysrepo, we can initialize the HW to some initial, factory-default state, but we need to fully propagate the content of the `startup` DS to the device's operational state to be able to report back to the user "yes, now it's done, and we're ready for operation". If this propagation fails for whatever reason, it means that there's been some fatal error, and the device is **not** up and running. We might want to roll back to the previous FW release, etc (and we're already doing that).

Some of our models cannot be empty, and for these we *could* just wait for the first callback to hit our code, but then there are other models where an empty state is OK.

I think that you're saying that your intention is to have this mode of operation:

subscription flags | event(s) send on the initial copy (`startup` -> `running`) | events on subsequent modifications of `running`
---|---|---
`SR_SUBSCR_DEFAULT` | just `SR_EV_DONE` | `SR_EV_CHANGE` and `SR_EV_DONE`
`SR_SUBSCR_DONE_ONLY` | just `SR_EV_DONE` | just `SR_EV_DONE`
`SR_SUBSCR_ENABLED` | `SR_EV_ENABLED` and `SR_EV_DONE` | `SR_EV_CHANGE` and `SR_EV_DONE`
 `SR_SUBSCR_ENABLED \| SR_SUBSCR_DONE_ONLY` | just `SR_EV_DONE` | only `SR_EV_DONE`

From the current documentation it is unclear to me that this is the expected behavior.

That is not sufficient for our use case I'm afraid, and it has a bug -- because the user's callbacks have no way of somehow tying that `SR_EV_DONE` (third row above, middle column) to the previous `SR_EV_ENABLED` (same table cell), the code will quite happily call `sr_get_change_iter`, which will fail if there are no changes in there.

So we need an event which always fires upon the initial copy, even if the diff is empty, and to be able to call `sr_get_change_iter` from within that event. How do I do that with sysrepo?

### 1: `SR_EV_ENABLED` instead of `SR_EV_DONE` for the first change event

A behavior like this would be perfectly fine for us, and the benefit is that you can keep that error checking for an empty diff for all other event types than `SR_EV_ENABLED` in place. The drawback is that this API might be surprising for some people -- suddenly you need to handle both `SR_EV_DONE` and `SR_EV_ENABLED`, but that's a one-liner change to make.

subscription flags | event(s) send on the initial copy (`startup` -> `running`) | events on subsequent modifications of `running`
---|---|---
`SR_SUBSCR_ENABLED` |  just `SR_EV_ENABLED` | `SR_EV_CHANGE` and then `SR_EV_DONE`
`SR_SUBSCR_ENABLED \| SR_SUBSCR_DONE_ONLY` | just `SR_EV_ENABLED` | only `SR_EV_DONE`

You cannot have `SR_EV_CHANGE` for the initial copy because calling `sr_get_change_iter` would fail.

### 2: relax error checking of `sr_get_change_iter`

This needs two changes:
- don't error out if there's no diff within `SR_EV_CHANGE` and `SR_EV_DONE`, 
- always generate the initial event even when there's no diff.

I've checked the code and I'm not sure if the second change needs to be done -- is there any active filtering which prevents the `SR_EV_DONE` on a first "copy" even if the datastore is empty? I think all my models actually have some default values, so it's hard for me to check at the moment.

michalvasko:
This is currently the expected behaviour
subscription flags | event(s) send on the initial copy (`startup` -> `running`) | events on subsequent modifications of `running`
---|---|---
`SR_SUBSCR_DEFAULT` | none | `SR_EV_CHANGE` and `SR_EV_DONE`
`SR_SUBSCR_DONE_ONLY` | none | just `SR_EV_DONE`
`SR_SUBSCR_ENABLED` | `SR_EV_ENABLED` and `SR_EV_DONE` | `SR_EV_CHANGE` and `SR_EV_DONE`
 `SR_SUBSCR_ENABLED \| SR_SUBSCR_DONE_ONLY` | just `SR_EV_DONE` | only `SR_EV_DONE`

Also, if an event(s) should be generated for the initial copy (this is not completely accurate but it should not matter now), they should be even if there are no changes (current data are empty). The idea was to make the initial events similar to the standard events (so they can be handled the same way) but at the same time be able to distinguish the initial events if needed (special `SR_EV_ENABLED`).

Based on your feedback I realized that it does not work as intended with flags `SR_SUBSCR_ENABLED | SR_SUBSCR_DONE_ONLY` because you 1) cannot distinguish between the events as you always get only `SR_EV_DONE` and 2) because of that you may call `sr_get_changes_iter` on no changes that will fail.

You suggested to change the behavior to
subscription flags | event(s) send on the initial copy (`startup` -> `running`) | events on subsequent modifications of `running`
---|---|---
`SR_SUBSCR_DEFAULT` | none | `SR_EV_CHANGE` and `SR_EV_DONE`
`SR_SUBSCR_DONE_ONLY` | none | just `SR_EV_DONE`
`SR_SUBSCR_ENABLED` | `SR_EV_ENABLED` | `SR_EV_CHANGE` and `SR_EV_DONE`
 `SR_SUBSCR_ENABLED \| SR_SUBSCR_DONE_ONLY` | `SR_EV_ENABLED` | only `SR_EV_DONE`

which should provide the versatility I had originally intended and was actually my initial idea, too.

Do you think anything I wrote is not correct meaning the behavior currently differs? Does the solution really work for you?

jktjkt:
Thanks for your summary. My patch is broken, and having that table helped me realize that I "fixed" it in a wrong way.

So, which way do we go?

- A quck-n-dirty way is just relaxing that check in `sr_get_changes_iter`. The downside is that the `SR_EV_ENABLED` event type feels a bit useless because it's going to be followed up by `SR_EV_DONE`. The user's code therefore loses that magic flag on the final commit event. On the other hand, it's a compatible change.

- Another way is inventing a better name for these two events. If we could disregard the backward compatibility concerns, my proposal would be to use something like `SR_EV_ENABLED_CHECK` followed by `SR_EV_ENABLED_COMMIT` for the initial copy, and `SR_EV_CHECK` followed by `SR_EV_COMMIT` for regular changes. That way, the `SR_SUBSCR_DONE_ONLY` can just filter out `SR_EV_ENABLED_CHECK` and `SR_EV_CHECK` (i.e., disable that two-phase commit), and the lack of `SR_SUBSCR_ENABLED` filters out `SR_EV_ENABLED_CHECK` and `SR_EV_ENABLED_COMMIT` (i.e., ignore the initial notifications about copy-config).

Or perhaps it can be made even more consistent. By default, assume that the users' code starts from some empty, possibly invalid state, and that the user doesn't really care *why* these modifications are being made. Have a flag which turns on that distinction between "initial copy" and "subsequent changes", and another flag which activates a two-phase commit for either of these two situations. This will require removing the error check in `sr_get_changes_iter` unconditionally:

subscription flags | event(s) send on the initial copy (`startup` -> `running`) | events on subsequent modifications of `running`
---|---|---
`SR_SUBSCR_DEFAULT` | `SR_EV_COMMIT` | `SR_EV_COMMIT`
`SR_SUBSCR_TWO_PHASE_COMMIT` | `SR_EV_CHECK` + `SR_EV_COMMIT` | `SR_EV_CHECK` + `SR_EV_COMMIT`
`SR_SUBSCR_TWO_PHASE_COMMIT \| SR_SUBSCR_ENABLED` | `SR_EV_ENABLED_CHECK` + `SR_EV_ENABLED_COMMIT` | `SR_EV_CHECK` + `SR_EV_COMMIT`
`SR_SUBSCR_ENABLED` | `SR_EV_ENABLED_COMMIT` | `SR_EV_COMMIT`

michalvasko:
Firstly, I would really avoid renaming the events (again) because I intentionally chose vague terms to not hint the users what to do on each event and leave it up to them (in contrast with legacy *sysrepo*).

Then, I think I should clarify what exactly the enabled event represents because I believe your second solution is way too complicated, the reason may be these details missing. So, actual copy from `startup` to `running` happens when the first connection is created after system start (when there are no SHM files). All the `SR_SUBSCR_ENABLED` flag does is generate a fake event (meaning `running` datastore is the same before and after the event) that simply publishes the current `running` data of the specific subscription for the callback as being created. Because that is what is essentially happening for a starting plugin, it needs to configure the device (or whatever) to be consistent with the `running` data in *sysrepo*.

That is also why I suggested the previous solution of using only a single event `SR_EV_ENABLED` to not overcomplicate things since they are even simpler than one would expect in this case. So again, can you implement the functionality you need with the change (solution) I mentioned in my previous reply?

jktjkt:
It turns out that this works: `_sr_get_changes_iter` does *not* fail. It seems that we both misinterpreted the `if`-clause. The error condition fires only if:
- it is something else than `SR_EV_APPLY` (`true`)
- it is something else than `SR_EV_DONE` (`false` in my case, so the whole `if` expression is `false`)
- the diff is `nullptr` (`true`)

I'll see if I can improve the documentation so that the different types of events are clearly documented. This is what one gets for, your words, "vague terms".

michalvasko:
I see. I did not check what you said because it made sense to me, I guess I should have.

PR #2212, opened 2020-11-09
travis FEATURE add sanitizer test job

jvijtiuk:
Hello,

this adds a Travis CI job to run unit tests with address and undefined behaviour sanitizers enabled.

It is based on the sanitizer job patch that was added to [libyang](https://github.com/CESNET/libyang/pull/1261). Valgrind tests are disabled when sysrepo is built with sanitizers. As the jobs are defined via the job matrix, and a single script is used, I've had to enable sanitizer tests with an ENV variable. If there is a better way to do that please let me know, and I will change the patch.

Thanks,
Juraj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,22 +7,32 @@ matrix:
             dist: bionic
             sudo: required
             compiler: gcc
+            env: SANITIZE="OFF"
         -   os: linux
             dist: bionic
             sudo: required
             compiler: clang
+            env: SANITIZE="OFF"
+        -   os: linux
+            dist: bionic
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
I do not think there is a better way. But I would prefer to keep the structure of the `script` the way it is now meaning no `if .. else`. Also, please rename the variable to `SANITIZE`, for example, just something more specific than `CHECK` and best to set it in every build to `TRUE`/`FALSE` or something, to make the logic obvious.

michalvasko:
Okay, thanks.

PR #2235, opened 2020-12-01
Qualify lock directory with repository prefix

imiller0:
Bug Fix: Qualify connection lock directory with prefix for unique locks per sysrepo instance.

The environment variable SYSREPO_SHM_PREFIX can be used to create unique instances of the sysrepo repository. The lock directory (and thus lockfiles) should be qualified by this prefix.

The calculation of lock file path is moved to common.c so that it is co-located with other file/path calculations.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1673,6 +1673,25 @@ sr_path_yang_file(const char *mod_name, const char *mod_rev, char **path)
     return err_info;
 }
 
+sr_error_info_t *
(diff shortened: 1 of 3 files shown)


michalvasko (approved):

michalvasko:
Okay, it seems fine, thanks.

imiller0:
Thank you!

Issue #2304, opened 2021-02-02
sysrepo/src/shm_mod.c:464: void sr_shmmod_unlock(struct sr_mod_lock_s *, int, sr_lock_mode_t, sr_cid_t, sr_sid_t): Assertion `!memcmp(&shm_lock->sid, &sid, sizeof sid)' failed.


jktjkt:
Hey, we just got this via our CI:
```
sysrepo/src/shm_mod.c:464: void sr_shmmod_unlock(struct sr_mod_lock_s *, int, sr_lock_mode_t, sr_cid_t, sr_sid_t):
 Assertion `!memcmp(&shm_lock->sid, &sid, sizeof sid)' failed.
```

It's triggered by a test which constantly pushes ops data to sysrepo. Unfortunately I don't have a stacktrace for you. It's with commit e7dc2ee4.

jktjkt:
Oh, the stack trace is apparently rather simple, here's the relevant part (after removing the TSAN bits and the event loop stack that is huuuuuge):
```
 #8 sr_shmmod_modinfo_unlock sysrepo/src/shm_mod.c:662:13 (libsysrepo.so.5+0x8e251)
 #9 sr_apply_changes sysrepo/src/sysrepo.c:2126:5 (libsysrepo.so.5+0x11288)
 #10 sysrepo::Session::apply_changes(unsigned int, int) sysrepo/bindings/cpp/src/Session.cpp:261:15 (libsysrepo-cpp.so.5+0x2a840)
```

pfeige:
I've got a similar assertion in [sr_shmmod_lock()](https://github.adtran.com/opensource/sysrepo/blob/e7dc2ee4321e00e6b4030416bdd664ad9b1b80f0/src/shm_mod.c#L451). This seem to be caused by an shm_lock->sid.user == NULL. The value is not always available as it can be seen in [sr_shmsub_notify_rite_event()](https://github.adtran.com/opensource/sysrepo/blob/e7dc2ee4321e00e6b4030416bdd664ad9b1b80f0/src/shm_sub.c#L333) and in [sr_shmsub_multi_notify_write_event()](https://github.adtran.com/opensource/sysrepo/blob/e7dc2ee4321e00e6b4030416bdd664ad9b1b80f0/src/shm_sub.c#L378).
I guess it would be better to use 
`assert(shm_lock->sid.sr == sid.sr && shm_lock->sid.nc == sid.nc);`
[...]

PR #2306, opened 2021-02-02
Do not use memcmp on sr_sid_t

jktjkt:
As shown in #2304 by @pfeige , the `username` can well be `NULL`. In that case, it would be an error to compare these `sr_sid_t` instances via `memcmp`.

### DIFF
--- a/src/shm_mod.c
+++ b/src/shm_mod.c
@@ -396,6 +396,11 @@ cleanup:
     sr_errinfo_free(&err_info);
 }
 
+inline static int sr_sid_compare(sr_sid_t *a, sr_sid_t *b)
+{
+    return a->sr == b->sr && a->nc == b ->nc;
+}
+
 /**
  * @brief Lock or relock a main SHM module.
  *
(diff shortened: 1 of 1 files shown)


jktjkt:
The [CI failure](https://travis-ci.org/github/sysrepo/sysrepo/jobs/757228459) appears to be a regression introduced by fb04659a08dbca37631bd73051a46cd0d6b4a8a3 which has left one test commented out.

jktjkt:
With this patch in place, I'm getting occasional failures in later stages of our integration tests, such as:

- an internal error in `src/shm_mod.c:744`
- in libyang's `src/parse_lyb.c:1175`, `Invalid first magic number byte "0x00"`
- in libyang's `src/parser_lyb.c:974`, `Failed to find matching hash for a top-level node from ...`

At this time I think these are merely uncovered by this patch, not really introduced by it, but I could be wrong.

michalvasko:
This is mostly a workaround like I mentioned in #2304 so I would like to look at it and fix it myself.

Issue #2080, opened 2020-08-04
Ubuntu 20.04 LTS permission denied error


michalvasko:
On Ubuntu 20.04, there is an error with yet unknown cause tracked in cesnet/netopeer2#621. In *sysrepo* it manifests in `permission denied` errors on most operations even when using the `root` user. The only solution we can suggest right now is not to use this distribution.

gunterw:
Any updates on this issue? 
I would like to move to 20.04, but require sysrepo2, so currently in a holding pattern.

michalvasko:
There are links to all the bug reports we did in the linked `netopeer2` issue. Sadly, there still was no real answer and there is nothing more we can do. If you can make someone from Ubuntu take an interest in the issue, it would help a great deal.

jktjkt:
This is not a Ubuntu-specific issue. It's caused by recent enough kernel (4.19+) and systemd (>= 241). The `/dev/shm` directory has the sticky bit set, [this kernel patch](https://lwn.net/Articles/748141/) added support for the `fs.protected_regular` sysctl knob, and this knob in turn [gets enabled via `systemd`](https://github.com/systemd/systemd/commit/2732587540035227fe59e4b64b60127352611b35). As a result, one cannot use `O_CREAT` on "somebody else's files" on those world-writable sticky directories.

I'll follow up on a random kernel mailing list to find out if this is the expected behavior here (my personal opinion is "WTF?").

michalvasko:
Um, okay, thanks for learning more about this and keep us posted. Also, it seems I am using kernel `5.8.15` and systemd `246`, everything is working fine.

jktjkt:
How does the output of `sysctl fs.protected_regular` look for you? If it returns `1`, you "should" have problems. If it's at `0`, then this security mechanism is not used, and it should indeed work fine.

michalvasko:
Okay, it is `0` so I am wondering whether it is the default or it is set some other way.

jktjkt:
Looks like Ubuntu might have [reverted to the older behavior](https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1845637). But from that bugreport it seems that they might be willing to implement the new defaults via the `kernel` package anyway in future -- I don't know.

michalvasko:
The config files on OpenSUSE are provided by special `aaa_base` package so that is probably not helpful. However, I have not found `fs.protected_regular` option set in any of those config files meaning it should be set to the default value.

[...]

PR #2319, opened 2021-02-11
Work around a paranoid kernel protection for sticky-bit directories

jktjkt:
Since the Linux kernel 4.19+ and systemd v241, there's a new sysctl knob which prevents users from overwriting "someone else's files" that are located in world-writable directories with the sticky bit set. That goes directly against sysrepo's permission model (which might become irrelevant anyway because it sits in parallel to the NACM).

When this change was made upstream, they cited a list of CVEs which are supposed to be preventable via that change. I believe that it turns out that all of them are actually *other*, serious bugs, but I acknowledge that it's true that `fs.protected_regular = 1` can indeed help against something stupid like `exec /tmp/install.sh`:

- CVE-2000-1134: Insecure file creation under /tmp.
- CVE-2007-3852: Insecure file creation of /tmp/sysstat.run.
- CVE-2008-0525: Symlinks below /tmp.
- CVE-2009-0416: Symlinks below /var/tmp.
- CVE-2011-4834: Symlink attack, and then content spoofing via a FIFO.
- CVE-2015-1838: Download & execute /tmp/install.sh.
- CVE-2015-7442: Insecure file creation in /tmp.
- CVE-2016-7489: Insecure file creation in /tmp.

Looking at that list, I still think that the upstream decision is nonsense, but hey, I've been postponing that mail to linux-fsdevel for four months already, so let's solve this within sysrepo. Just provide a drop-in file for systemd to undo this nonsensical (IMHO, anyway) policy change.

Bug: https://github.com/sysrepo/sysrepo/issues/2080
Cc: @syyyr @peckato1

### DIFF
--- a/66-sysrepo-disable-fs-protected_regular.conf
+++ b/66-sysrepo-disable-fs-protected_regular.conf
@@ -1,0 +1,4 @@
+# Make sure that systemd's and kernel's settings still allow other users to
+# access sysrepo SHM files
+# Bug: https://github.com/sysrepo/sysrepo/issues/2080
+fs.protected_regular = 0
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,6 +284,8 @@ install(TARGETS sysrepoctl sysrepocfg sysrepo-plugind DESTINATION ${CMAKE_INSTAL
 
 install(FILES "${PROJECT_BINARY_DIR}/sysrepo.pc" DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
 
+install(FILES "${PROJECT_SOURCE_DIR}/66-sysrepo-disable-fs-protected_regular.conf" DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/sysctl.d/")
+
 # examples
 if(BUILD_EXAMPLES)
     add_subdirectory(examples)


michalvasko:
You mean I do not have to keep fixing bugs and the kernel will fix them for me? Neat.

On a more serious note, I am not sure this really is okay even though I cannot really say why not. What exactly is required for the file to take effect, machine reboot or restarting *systemd* (or something else) is enough? Also, please put the file somewhere else (only package-exclusive files are in `packages`) although I am not sure where. Maybe the root is fine or even better a separate directory (`sysctl_conf` maybe?).

jktjkt:
I agree that this is not an optimal change, we're doing this to override the system policy, after all. In this case, however, I think the system policy is just silly, and the stated goals do not match the implementation. At the same time, reverting this upstream will, realistically, take years at best, so the best option we can do (IMHO) is to try to minimize the damage and reduce the support load. Unfortunately, we cannot even issue a warning at build time because one cannot even *read* that sysctl knob as a non-root. Also, I suspect that any complaint about this behavior will trigger a "you're doing it wrong" response, and I do not think that you would feel happy with abandoning the SHM approach either, so, well, meh, this is where we are.

You'll have to reboot or restart the [`systemd-sysctl.service`](https://man7.org/linux/man-pages/man8/systemd-sysctl.service.8.html) for this change to take effect. This file location only really works with systemd (see [`sysctl.d(5)`](https://man7.org/linux/man-pages/man5/sysctl.d.5.html)). The alternative would be pushing stuff directly to `/etc/sysctl.d/`, but then you would still hit some crazy setups that only have `/etc/sysctl.conf`, *and* `/etc` is usually not for packages to put their stuff in directly, so this is probably the sanest thing one can do.

Whom does this affect -- I would argue that developers do not care at all because they are unlikely to pass explicit `--user` or `--group` flags to `sysrepoctl`. So the only affected people (IMHO) are those who install sysrepo as a systemwide thingy. Presumably, those people are building something around sysrepo, so it makes sense for them to tweak their settings in a way which makes it possible to let sysrepo work. That is, by the way, the reason I put that file to `packages/`.

The alternative would be abandoning the UNIX permission model altogether. The current SHM code looks like something which *trusts* the in-memory data that are in the SHM files. As such, an attacker that is able to modify these SHM files directly, without using the sysrepo APIs -- which is **anyone** in the default config because `/dev/shm` is world-writable -- can probably cause any other sysrepo user to crash or possibly execute arbitrary code. Perhaps it would make sense to recognize this general vulnerability and move these files from `/dev/shm` to something "safe" like `/run/sysrepo`, have that one owned by `sysrepo:sysrepo`, permissions `0770`, and mandate all sysrepo-calling processes to use the same `sysrepo` EUID. Yes, that would be super-inconvenient for *other* bits, such as netopeer2 access to SSH `authorized_keys`, or for any daemon which is designed to run under a particular UID of its own, but hey, that's what you get with using shared memory I guess.

I've now moved the file to the top level directory based on your request.

michalvasko:
> Presumably, those people are building something around sysrepo, so it makes sense for them to tweak their settings in a way which makes it possible to let sysrepo work. That is, by the way, the reason I put that file to packages/.

Okay, but I still would not make this file package-exclusive. What I think would be the safest solution is only optional installing of this file, off by default. As for packages, they can be tuned so that the affected distributions include it, others do not. For manual build, the user would need to enable it himself. While we may then still get issues complaining about the permission denied error, it makes the user responsible for installing the file, not us. Of course, this problem and the solution should be mentioned in the README, at the beginning somewhere to cover the users that look there, at least (you can add that if you want, or I will).

> The current SHM code looks like something which trusts the in-memory data that are in the SHM files. As such, an attacker that is able to modify these SHM files directly, without using the sysrepo APIs -- which is anyone in the default config because /dev/shm is world-writable -- can probably cause any other sysrepo user to crash or possibly execute arbitrary code. Perhaps it would make sense to recognize this general vulnerability and move these files from /dev/shm to something "safe" like /run/sysrepo

The SHM data files are not implicitly trusted, this vulnerability is know and hence there is proper file permissions management so that you can easily set the correct permissions for the files and avoid any security problems. `/dev/shm` is used because of the efficiency, to avoid using persistent files when they are not needed.

jktjkt:
> Okay, but I still would not make this file package-exclusive. What I think would be the safest solution is only optional installing of this file, off by default.

Feel free to take this one from here and handle as you deem appropriate, then. If this is hidden behind an extra flag I will just add this one-liner via our build scripts.

michalvasko:
Yes, that is the idea, another `cmake` variable. Okay, I will finish it and merge it (soon).

oerdnj:
I think it's wrong to change system-wide settings for security settings. This change won't even give the user any warning at all.

I would suggest that by default the directory would be set only with setgid and group writable only. Then anybody who needs access to the repository would have to be added to the group.  Using world-writable directories with sticky bit is not very good security model anyway.

michalvasko:
> I think it's wrong to change system-wide settings for security settings. This change won't even give the user any warning at all.

So what exactly do you propose? We would like to find a way for *sysrepo* to work out-of-the-box on distributions that have this bit set, which makes *sysrepo* not function properly.

> I would suggest that by default the directory would be set only with setgid and group writable only. 

Which directory do you mean? The persistent repository (`/etc/sysrepo` for release)?

> Then anybody who needs access to the repository would have to be added to the group. Using world-writable directories with sticky bit is not very good security model anyway.

What else can we do if we want to use SHM files instead of regular ones? I do not think they can be placed anywhere else than `/dev/shm`.

jktjkt:
I think that depends on what you can reasonably demand form the system. One option is documenting that packages should provide a `tmpfs` at something like `/run/sysrepo`, and using that. A developer build can continue with `/dev/shm` as before because it's unlikely that they will use multiple UIDs to access sysrepo anyway.

However, even if we have this `/run/sysrepo` owned by `root:sysrepo` with that group-write bit on, that would only help with `fs.protected_regular = 1`. With `fs.protected_regular = 2`, the kernel won't allow `O_CREAT` by another user than the one which owns the target file.

michalvasko:
Okay, it should be merged but disabled by default so it should do no harm. It can be further improved later.

nickbroon:
Debian/Ubuntu by default ship a /usr/lib/sysctl.d/protect-links.conf

```
fs.protected_regular = 2
```

Installing a /usr/lib/sysctl.d/66-sysrepo-disable-fs-protected_regular.conf with `fs.protected_regular = 0`

is not going to be applied by `systemd-sysctl.service` as files are sorted in lexicographic order ([sysctl.d(5)](https://man7.org/linux/man-pages/man5/sysctl.d.5.html)), and protected-links.conf will be sorted after 66-sysrepo-disable-fs-protected_regular.conf

michalvasko:
Seems Debian is not using number prefixes but other distributions are. Not sure what to do about this, it obviously cannot fit all the use-cases.

nickbroon:
I've submitted https://github.com/sysrepo/sysrepo/pull/2558 which should help on distributions like debian/ubuntu.

PR #2322, opened 2021-02-12
[WIP] tests: apply_changes: use proper atomics instead of `volatile`

jktjkt:
At this time I would appreciate your input if you are OK with this approach. If so, I can mass-change the other tests as well. Some of these patterns are still inherently racy (there's essentially a double-load everywhere that variable is used in the `switch` block a second time), but this is rather tedious work, so I took this shortcut.


michalvasko:
It was written the most simple way that I knew it will work because using atomics here seems rather excessive. The value is always incremented in a single thread while another one is simultaneously reading it, so the only problem that can occur as far as I can tell is reading the value while it is being written and hence reading some junk. Whether that is actually possible to happen (whether using volatile, atomics, or even a standard variable) I doubt. But I am not sure how else to silence TSAN, maybe just write a suppression file? Easier than rewriting all of the tests

jktjkt:
I understand that "this is just a test". The problem is that TSAN's suppression system is based on pattern matching some strings against the symbol names (and file names, and...) on stack. One *could* exclude, say, "the whole test file", but that's too big of a hammer because there will always be that particular file at the top of each stack. I'm afraid I don't know how to exclude just this particular memory location from data race reporting.

I'm not a C guy, if this was C++, it would be just a matter of changing the variable type to `std::atomic<int>` at the place it's declared, but that's of course not feasible here.

michalvasko:
I have looked at the suppressions available for TSAN and you are right, it would probably not even be possible to ignore it this way as you can only ignore whole functions (`valgrind` is better in this). So, I do not have any better idea than just using atomics everywhere. One could write some small API special for tests that would implement these counters but there is likely no advantage compared to your solution. So please, rewrite all the tests this way and I will merge it, thanks.

jktjkt:
> So please, rewrite all the tests this way and I will merge it, thanks.

Done (it was just a bunch of `sed`s, and it works). I've listed some of the real issues that I'm getting in #2323.

PR #2340, opened 2021-02-25
lyd mods CHANGE more efficient data storage

lePici:
Unchanged data is not rewritten during context change.

### DIFF
--- a/src/lyd_mods.c
+++ b/src/lyd_mods.c
@@ -1053,6 +1053,35 @@ sr_lydmods_ctx_load_modules(const struct lyd_node *sr_mods, struct ly_ctx *ly_ct
 }
 
 /**
+ * @brief Compare two unlinked Data tree.
+ * @param[in] new unlinked Data tree.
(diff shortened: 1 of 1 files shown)


src/lyd_mods.c
```diff
@@ -1187,25 +1218,40 @@ sr_lydmods_sched_update_data(const struct lyd_node *sr_mods, const struct ly_ctx
         goto cleanup;
     }
 
-    /* print all modules data with the updated module context and free them, no longer needed */
+    /* Print all modules data with the updated module context if the new data is different from the old ones.
+     * Then free them, no longer needed.
+     */
     for (idx = 0; idx < set->number; ++idx) {
+        int diff;
```
michalvasko, line 1225: Rename as well.

michalvasko (changes requested):
As for the function parameters, now it no longer matter which data are old and which are new so the original names would be fine. But these are fine too, so no change needed.

michalvasko (approved):

PR #2362, opened 2021-03-11
test NEW update data deviation test

lePici:
Attempts to install and uninstall modules containing a deviation
statement, triggering changes in context.

### DIFF
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -841,7 +841,7 @@ API int
 sr_install_module(sr_conn_ctx_t *conn, const char *schema_path, const char *search_dirs, const char **features)
 {
     sr_error_info_t *err_info = NULL;
-    struct ly_ctx *tmp_ly_ctx = NULL;
+    struct ly_ctx *tmp_ly_ctx = NULL, *sr_mods_ctx = NULL;
     struct lyd_node *sr_mods = NULL;
     const struct lys_module *ly_mod, *ly_iter, *ly_iter2;
     LYS_INFORMAT format;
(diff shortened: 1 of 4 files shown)


src/sysrepo.c
```diff
@@ -854,12 +854,20 @@ sr_install_module(sr_conn_ctx_t *conn, const char *schema_path, const char *sear
...
+    /* create a bond with the second temporary context */
+    if ((err_info = sr_lydmods_parse(tmp_ly_ctx2, &sr_mods))) {
         goto cleanup;
     }
+    /* use first temporary context to load modules */
     if ((err_info = sr_lydmods_ctx_load_modules(sr_mods, tmp_ly_ctx, 1, 1, 0, NULL))) {
         goto cleanup;
     }
+    lyd_free_all(sr_mods);
+    ly_ctx_destroy(tmp_ly_ctx2, NULL);
```
michalvasko, line 870: Always be careful about your cleanup, in this case both these variables will leak for any `cleanup` above them.

tests/test_modules.c
```diff
@@ -1520,6 +1520,72 @@ test_feature_dependencies_across_modules(void **state)
...
+    ret = sr_connection_count(&conn_count);
+    assert_int_equal(ret, SR_ERR_OK);
+    assert_int_equal(conn_count, 0);
+    ret = sr_connect(SR_CONN_ERR_ON_SCHED_FAIL, &st->conn);
+    assert_int_equal(ret, SR_ERR_OK);
+
+    /* clean up - remove first and third module */
+    ret = sr_remove_module(st->conn, "test-cont");
+    assert_int_equal(ret, SR_ERR_OK);
+    ret = sr_remove_module(st->conn, "defaults");
```
michalvasko, line 1585: I am guessing this module name is wrong.

michalvasko (changes requested):

michalvasko:
Also, ignore `travis`, it is not compiling with `libyang2` so it will always fail.

tests/test_modules.c
lePici: I think the module name is correct.

michalvasko (approved):

PR #2383, opened 2021-03-30
sysrepo BUGFIX overwrite data when changing ctx

lePici:


src/lyd_mods.c
```diff
@@ -1069,6 +1069,42 @@ sr_lydmods_ctx_load_modules(const struct lyd_node *sr_mods, struct ly_ctx *ly_ct
     return NULL;
 }
 
+/**
+ * @brief Learn whether any module in a set augments/deviates a specific module.
+ *
+ * @param[in] old_base_mod Module that can be augmented/deviated.
+ * @param[in] old_mod_set Set with modules possibly augmenting/deviating @p mod.
```
michalvasko, line 1076: s/mod/old_base_mod

src/lyd_mods.c
```diff
@@ -1069,6 +1069,42 @@ sr_lydmods_ctx_load_modules(const struct lyd_node *sr_mods, struct ly_ctx *ly_ct
...
+ * @return Whether any module in @p mod_set was augmenting/deviating @p mod.
```
michalvasko, line 1077: s/mod_set/old_mod_set; s/mod/old_base_mod

src/lyd_mods.c
```diff
@@ -1069,6 +1069,42 @@ sr_lydmods_ctx_load_modules(const struct lyd_node *sr_mods, struct ly_ctx *ly_ct
...
+    const struct lys_module *old_mod;
+    uint32_t i;
+    LY_ARRAY_COUNT_TYPE u;
+
+    for (i = 0; i < old_mod_set->count; ++i) {
+        old_mod = old_mod_set->objs[i];
+
+        /* deviates */
+        LY_ARRAY_FOR(old_base_mod->deviated_by, u) {
+            if (!strcmp(old_base_mod->deviated_by[u]->name, old_mod->name)) {
```
michalvasko, line 1091: It should not be needed to compare their names, comparing pointers should work since both modules must be from the `old` context.

src/lyd_mods.c
```diff
@@ -1069,6 +1069,42 @@ sr_lydmods_ctx_load_modules(const struct lyd_node *sr_mods, struct ly_ctx *ly_ct
...
+            if (!strcmp(old_base_mod->augmented_by[u]->name, old_mod->name)) {
+                return 1;
+            }
+        }
+    }
+
+    return 0;
+}
+
+
```
michalvasko, line 1107: Extra line.

michalvasko (changes requested):

src/lyd_mods.c
```diff
@@ -1069,6 +1069,41 @@ sr_lydmods_ctx_load_modules(const struct lyd_node *sr_mods, struct ly_ctx *ly_ct
...
+    const struct lys_module *old_mod;
+    uint32_t i;
+    LY_ARRAY_COUNT_TYPE u;
+
+    for (i = 0; i < old_mod_set->count; ++i) {
+        old_mod = old_mod_set->objs[i];
+
+        /* deviates */
+        LY_ARRAY_FOR(old_base_mod->deviated_by, u) {
+            if (old_base_mod->deviated_by[u]->name == old_mod->name) {
```
michalvasko, line 1091: Um, sorry for not being more clear. I mean, this should work, too, but what I meant was comparing the module pointers directly, no need to compare names, that is just redundant and confusing.

michalvasko (changes requested):

src/lyd_mods.c
lePici: ok

michalvasko (approved):
Should be fine now but it cannot be merged until *libyang2* PR is merged so while waiting for that, you can try to refactor `sr_lydmods_sched_update_data()` by splitting it into functions somehow because it is way too long. But do it in a separate commit!

PR #2417, opened 2021-05-16
Do no routinely emit "info" messages

jktjkt:
Our devices push operational data into sysrepo periodically, often several times a second. Here's how a log usually looks like on a production box:
```
May 16 15:56:07 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 891 priority 0 for 1 subscribers.
May 16 15:56:07 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 892 priority 0 for 1 subscribers.
May 16 15:56:07 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 893 priority 0 for 1 subscribers.
May 16 15:56:08 line-qr79 sysrepo[459]: No datastore changes to apply.
May 16 15:56:09 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 894 priority 0 for 1 subscribers.
May 16 15:56:09 line-qr79 sysrepo[459]: No datastore changes to apply.
May 16 15:56:10 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 895 priority 0 for 1 subscribers.
May 16 15:56:10 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 896 priority 0 for 1 subscribers.
May 16 15:56:10 line-qr79 sysrepo[459]: Published event "done" "czechlight-roadm-device" with ID 897 priority 0 for 1 subscribers.
```
These are produced via the sysrepo library and stored via our application code because of the `info` message severity.

Reduce the log verbosity a bit down to `debug` for those debugging messages.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -529,7 +529,7 @@ sr_shmsub_multi_notify_write_event(sr_multi_sub_shm_t *multi_sub_shm, uint32_t r
     }
 
     if (event) {
-        SR_LOG_INF("Published event \"%s\" \"%s\" with ID %u priority %u for %u subscribers.", sr_ev2str(event),
+        SR_LOG_DBG("Published event \"%s\" \"%s\" with ID %u priority %u for %u subscribers.", sr_ev2str(event),
                 event_desc, request_id, priority, subscriber_count);
     }
     return NULL;
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -1949,7 +1949,7 @@ sr_changes_notify_store(struct sr_mod_info_s *mod_info, sr_session_ctx_t *sessio
     *cb_err_info = NULL;
 
     if (!mod_info->diff) {
-        SR_LOG_INF("No datastore changes to apply.");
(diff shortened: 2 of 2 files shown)


michalvasko:
Generally, you would solve similar problems by decreasing the verbosity. Why did you not simply do that?

jktjkt:
This is probably a bikeshedding discussion :), but in the past we've seen cases where `info` messages contained information that were very useful for debugging an actual failure. I don't know how is the situation today, and based on a quick grep, it looks like these info-level messages are in fact a debug ones. So we will probably just map sysrepo's `debug` to our internal `trace`, and sysrepo's `info` to our internal `debug`.

michalvasko:
I am not saying you do not have a point, I was just wondering why did you decide to do a PR instead of a simple change in your code :) I am often not sure what level to assign for a message and, more importantly, the messages were changing a lot. So I may take an explicit look at the log levels later but the change you did is probably the right one.

PR #2440, opened 2021-05-27
cmake BUGFIX don't install ${LIB_HEADERS} in /usr/include

jsmolic:
* This shouldn't be installed there as `values.h` causes file collision with
  glibc, and we are already installing `sysrepo.h` and `sysrepo_types.h`
  in `/usr/include` anyway, so let's just remove it from the install
  command

Signed-off-by: Jakov Smolic <jakov.smolic@sartura.hr>


michalvasko:
Oh, right, merging.

PR #2446, opened 2021-05-31
BUGFIX: Correct permissions check when mode_t is uint16_t.

snar:
Currently permissions (of type mode_t) are initialized with (mode_t)-1 and then 
(int)perms checked against -1. Unfortunately, when mode_t is defined as uint16_t 
(FreeBSD 11), perms became 65535, and (int)65535 never equals -1.


michalvasko:
No problem, thanks.

PR #2476, opened 2021-06-11
BUGFIX: build failure on FreeBSD

snar:
Setting  _POSIX_C_SOURCE to 199309L removes <netinet6/in6.h> from inclusion,
leading to build failure.


michalvasko:
Okay, build seems to be passing without it.

PR #2487, opened 2021-06-21
shm mod BUGFIX fix lock ordering for writes

### LINKED ISSUE #2486: Multiple simultaneous sr_apply_changes() can result in deadlock
If there are two modules, say alpha and beta, and one imports the other such that it has a dependency on it, then when changes are written using sr_apply_changes() or sr_copy_config() the read lock will be acquired for the other module.

The semantics of the read(upgradable) locks are that they can be taken whilst read locks are held, but not when a write lock is held or when another read(upgradable) lock is held, and only the required modules (those being written into) use read(upgradable) locks. The steps that sr_apply_changes() takes are the following from a high level:

1. Take locks. Take a read(upgradable) lock for alpha, read lock for beta.
2. Issue some notifications
3. Upgrade read(upgradable) lock to write lock for alpha
4. Write into datastore.
5. Downgrade write lock to read(upgradable) lock for alpha.
6. More notifications
7. Release locks

Similar will occur when writing to beta, except the read(upgradable) lock for beta will be taken and the read lock for alpha. Since a read(upgradable) is not mutually exclusive with a read lock then a write to beta can occur concurrently with a write to alpha and get down to (4). If the thread for writing into alpha runs concurrently with that writing into beta then they can deadlock (with the timed-wait get-out-of-jail-mostly-free card) because each wants to acquire the write lock for the lock that the other has the read lock for.

This is a well-known problem with reader/writer locks and so supporting being upgraded is generally done in most implementations, although the problems aren't insurmountable, at least in this case here.

By extending the unit tests to hit this case then the following is seen:
```
[INF]: Published event "change" "when2" with ID 5 priority 0 for 1 subscribers.
[INF]: Processing "when2" "change" event with ID 5 priority 0 (remaining 1 subscribers).
[INF]: Successful processing of "change" event with ID 5 priority 0 (remaining 0 subscribers).
[INF]: Event "change" with ID 5 priority 0 succeeded.
[ERR]: Waiting on a conditional variable failed (sr_shmmod_lock: Connection timed out).
0xd != 0
[   LINE   ] --- /home/rob/graphnos/sysrepo/tests/test_apply_changes.c:6301: error: Failure!
```

Whilst control does return to the application after a short while and it may retry the failed transaction (ignoring the impact to subscribers that duplicate notifications may have), this soon becomes unworkable since it results in more writes potentially getting backed up and more contention and more chance of hitting the race and applications potentially having to retry indefinitely.

rshearman:
Change the semantics of read locks taken during module writes so they
don't co-exist with other read locks taken for module writes to avoid
(timed) deadlocks when locks are taken in opposite way.

Test included.

Fixes #2486

### DIFF
--- a/src/shm_mod.c
+++ b/src/shm_mod.c
@@ -502,7 +502,7 @@ sr_shmmod_modinfo_rdlock(struct sr_mod_info_s *mod_info, int upgradeable, uint32
     if (upgradeable) {
         /* read-upgr-lock main DS */
         if ((err_info = sr_shmmod_modinfo_lock(mod_info, mod_info->ds, MOD_INFO_RLOCK | MOD_INFO_RLOCK_UPGR |
-                MOD_INFO_WLOCK, MOD_INFO_REQ, SR_LOCK_READ_UPGR, MOD_INFO_RLOCK_UPGR, sid))) {
+                MOD_INFO_WLOCK, 0, SR_LOCK_READ_UPGR, MOD_INFO_RLOCK_UPGR, sid))) {
             return err_info;
         }
     }
@@ -558,8 +558,9 @@ sr_shmmod_modinfo_rdlock_upgrade(struct sr_mod_info_s *mod_info, uint32_t sid)
         mod = &mod_info->mods[i];
         shm_lock = &mod->shm_mod->data_lock_info[mod_info->ds];
 
-        /* upgrade only read-upgr-locked modules */
-        if (mod->state & MOD_INFO_RLOCK_UPGR) {
+        /* upgrade only required read-upgr-locked modules and leave others read-upgr-locked to prevent their locking causing potential dead-lock */
+        if ((mod->state & (MOD_INFO_RLOCK_UPGR | MOD_INFO_REQ)) ==
+                (MOD_INFO_RLOCK_UPGR | MOD_INFO_REQ)) {
             /* MOD WRITE UPGRADE */
             if ((err_info = sr_shmmod_lock(mod->ly_mod, mod_info->ds, shm_lock, SR_MOD_LOCK_TIMEOUT, SR_LOCK_WRITE,
                     mod_info->conn->cid, sid, 1))) {
--- a/tests/test_apply_changes.c
+++ b/tests/test_apply_changes.c
@@ -6280,6 +6280,116 @@ test_change_enabled(void **state)
     pthread_join(tid[1], NULL);
 }
 
+#define APPLY_ITERATIONS 100
+
+static void *
+apply_when1_thread(void *arg)
+{
+    struct state *st = (struct state *)arg;
+    sr_session_ctx_t *sess;
+    int ret;
+
+    ret = sr_session_start(st->conn, SR_DS_RUNNING, &sess);
+    assert_int_equal(ret, SR_ERR_OK);
+
+    for (int i = 0; i < APPLY_ITERATIONS; i++) {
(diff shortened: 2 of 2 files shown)


src/shm_mod.c
```diff
@@ -558,8 +558,9 @@ sr_shmmod_modinfo_rdlock_upgrade(struct sr_mod_info_s *mod_info, uint32_t sid)
         mod = &mod_info->mods[i];
         shm_lock = &mod->shm_mod->data_lock_info[mod_info->ds];
 
-        /* upgrade only read-upgr-locked modules */
-        if (mod->state & MOD_INFO_RLOCK_UPGR) {
+        /* upgrade only required read-upgr-locked modules */
```
michalvasko, line 561: Please, explain the fix (at least briefly) somewhere, here, for example. Something like
```
... and leave others read-upgr-locked to prevent their locking causing potential dead-lock
```

src/shm_mod.c
```diff
@@ -558,8 +558,9 @@ sr_shmmod_modinfo_rdlock_upgrade(struct sr_mod_info_s *mod_info, uint32_t sid)
...
+        if ((mod->state & (MOD_INFO_RLOCK_UPGR|MOD_INFO_REQ)) ==
```
michalvasko, line 562: `uncrustify` probably expects spaces around `|`.

tests/test_apply_changes.c
```diff
@@ -6280,6 +6280,116 @@ test_change_enabled(void **state)
...
+    (void)session;
+    (void)sub_id;
+    (void)module_name;
+    (void)xpath;
+    (void)event;
+    (void)request_id;
+    (void)private_data;
+
+    /* yield to make any race conditions more evident */
+    pthread_yield();
```
michalvasko, line 6362: I have usually used a barrier for deterministic synchronization but that is probably not possible here, so fine.

michalvasko (changes requested):
Thanks for the fix and a test! Just some minor details to tune and please set the target branch to `devel`, we do not accept PRs to `master`.

src/shm_mod.c
```diff
@@ -558,8 +558,9 @@ sr_shmmod_modinfo_rdlock_upgrade(struct sr_mod_info_s *mod_info, uint32_t sid)
         mod = &mod_info->mods[i];
         shm_lock = &mod->shm_mod->data_lock_info[mod_info->ds];
 
-        /* upgrade only read-upgr-locked modules */
-        if (mod->state & MOD_INFO_RLOCK_UPGR) {
+        /* upgrade only required read-upgr-locked modules and leave others read-upgr-locked to prevent their locking causing potential dead-lock */
+        if ((mod->state & (MOD_INFO_RLOCK_UPGR | MOD_INFO_REQ)) ==
+            (MOD_INFO_RLOCK_UPGR | MOD_INFO_REQ)) {
```
michalvasko, line 563: Here I think it should have one indent more.

rshearman: Ok, done. My distro doesn't have a version of uncrustify new enough for cmake to use.

michalvasko: Neither does the Ubuntu 18.04 the CI is running on, it is compiled there from [source](https://github.com/sysrepo/sysrepo/blob/devel/.github/workflows/ci.yml#L127).

michalvasko (approved):

Reshad-Rahman:
Does/can this fix have to go in libyang1 too?

michalvasko:
Yes, it seems the exact same fix can be merged, so you can make another PR.

PR #2534, opened 2021-08-10
Improve lock recovery to avoid endless while loop

gunterw:
I was getting lockups on my system where the pthread_cond_timedwait function would return success(0) and yet there where still
cids in the readers structure, this would spin in the while loop forever.

fyi - not sure how/why this happens on my system but we use the SR_SUBSCR_NO_THREAD for all of our subscriptions,
and I only have one connection per process.


gunterw:
I added some debug logic and discovered that the lockup for me was occurring in the sr_rwrelock function

michalvasko:
So what happened, did you learn more? But your fixes seem to be just workarounds at best.

PR #2558, opened 2021-09-01
Install a sysctl.d file that is more like to apply last

nickbroon:
systctl.d files are applied in lexicographical order.
Some distributions, like debian/ubuntu, don't follow convention and
prefix files with digits to give easy to understand order.
So use, zz, which is much more likely to be last, and thus override
distribution shipped files.


michalvasko:
Fine, even though it is not that nice for distributions with numbered conf files.

nickbroon:
Fab, thanks.

For what it's worth I've opened this with Debian: https://salsa.debian.org/debian/procps/-/merge_requests/3
Though even if that's merged, it will only go into Debian Bookworm, which is unlikely to release for a couple of years.

michalvasko:
Well, that is a minor issue, really. The main problem is the default parameter value it is changing, which should be `0`.

PR #2556, opened 2021-08-31
CI: TSAN builds

jktjkt:
There was an ASAN+UBSAN build job, but it was running on an old clang, and UBSAN failures were being silently ignored. Make sure that we're on the latest & greatest so as not to hit too many compiler errors, and then enforce ASAN+UBSAN. We've been doing a similar thing for the libyang-v1+sysrepo internally, so let's make this available to the wider community. And while we're at that, enable TSAN as well now that the build is clean.

The last commit currently fails due to #2555. Once a fix for that UB is merged, this should start passing. All other commits [pass](https://github.com/jktjkt/sysrepo/commits/830964c26ad73545c1aeb77c34dc5de873196073) (there should be a green tick by all of them, hopefully).

PR #2562, opened 2021-09-03
CI: ASAN+UBSAN

jktjkt:
A version of #2556 which only implements ASAN+UBSAN, see the other PR for context. TSAN is left due to a possible false positive related to sysrepo's hand-rolled locking.


michalvasko:
Fine, no problem.

PR #2566, opened 2021-09-06
Change some typedef structs into normal structs

syyyr:
Hi, while working on my C++ wrapper for sysrepo, I found out that some of the structs are defined as typedefs. This means that you can't forward declare them. I think this change should not have an effect on existing code, but I'm not sure. What do you think?

I'm creating this as a draft, because I'm not sure what structs I'll have to change.

### DIFF
--- a/src/sysrepo_types.h
+++ b/src/sysrepo_types.h
@@ -129,20 +129,24 @@ typedef enum {
 /**
  * @brief A single, detailed error message. Used in sr_error_info_t
  */
-typedef struct {
+struct sr_error_info_err_s {
     sr_error_t err_code;    /**< Error code. */
     char *message;          /**< Error message. */
     char *error_format;     /**< Error format identifier. */
     void *error_data;       /**< Opaque error data specific for @p error_format. */
-} sr_error_info_err_t;
+};
+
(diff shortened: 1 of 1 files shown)


syyyr:
Ideally, I would like all structs to be in the C "struct namespace" and then possibly in the global "typedef" namespace.

src/sysrepo_types.h
```diff
@@ -244,8 +244,9 @@ typedef struct {
 
     /** Data of an element (if applicable), properly set according to the type. */
     sr_data_t data;
+} ;
```
michalvasko, line 247: Extra space.

michalvasko (changes requested):

michalvasko:
I have no problem with the change and any like this. However, please rebase your branch to the current `devel`, you seem to have quite an old revision.

src/sysrepo_types.h
syyyr: done

syyyr:
>I have no problem with the change and any like this.

Great! I changed all (or most) of them.

>However, please rebase your branch to the current devel, you seem to have quite an old revision.

The patch is now rebased to devel.

michalvasko (approved):

PR #2630, opened 2021-11-11
Add a helper for NETCONF- and RESTCONF-compliant errors

jktjkt:
Sysrepo uses a flexible internal mechanism which client applications can use for storing arbitrary error information. So far, however, the only conventions for how these error messages are supposed to be created were defined in Netopeer2's `README` file. This patch builds on top of that format.

I changed serialization of the `error-data` construct. Previously, it would be a sequence of pairs of strings which was used for building `anyxml` elements and their string values, respectively. The proposed format accepts a libyang data tree and uses the `LYB` formatting for passing that around.

Cc: @syyyr

### DIFF
--- a/src/sysrepo_types.h
+++ b/src/sysrepo_types.h
@@ -168,6 +168,16 @@ struct sr_error_info_s {
 typedef struct sr_error_info_s sr_error_info_t;
 
 /**
+ * @brief Error type as used in the NETCONF and RESTCONF protocols
+ */
+typedef enum {
+    SR_NC_ERR_TYPE_TRANSPORT = 0,
+    SR_NC_ERR_TYPE_RPC = 1,
+    SR_NC_ERR_TYPE_PROTOCOL = 2,
+    SR_NC_ERR_TYPE_APPLICATION = 3
+} sr_netconf_error_type_t;
+
+/**
  * @brief Callback to be called before applying a diff. Set it using ::sr_set_diff_check_callback.
  *
  * @param[in] session Implicit session (do not stop) with information about the event originator session IDs.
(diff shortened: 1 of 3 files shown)


src/sysrepo_types.h
```diff
@@ -167,6 +167,16 @@ struct sr_error_info_s {
 
 typedef struct sr_error_info_s sr_error_info_t;
 
+/**
+ * @brief Error type as used in the NETCONF and RESTCONF protocols
+ */
+typedef enum {
```
jktjkt, line 173: I would be OK with using a string for this thing if you prefer to bake less NETCONF/RESTCONF info into sysrepo.

michalvasko: Yes, I think I would prefer that.

michalvasko (commented):
I think I would prefer to take this further. If the errors are to be defined in sysrepo, which I have agreed to, I guess, then it should be complete. That means moving the exact error description from netopeer2 README to sysrepo, somewhere. The idea would probably be to write a brief doc section on these errors and define a list of well-know error formats, which will include helper functions by sysrepo. And that is where the `NETCONF` error format would be specified. I am fine with writing this myself, up to you.

jktjkt:
> I think I would prefer to take this further. If the errors are to be defined in sysrepo, which I have agreed to, I guess, then it should be complete. That means moving the exact error description from netopeer2 README to sysrepo, somewhere. The idea would probably be to write a brief doc section on these errors and define a list of well-know error formats, which will include helper functions by sysrepo. And that is where the `NETCONF` error format would be specified. I am fine with writing this myself, up to you.

Please go for it :). As long as there's "something" with an API that's similar to the function I proposed (and no matter if there are `enum`s or magic strings), I'm OK with whatever shape it takes.

jktjkt:
Another version of this has been merged (thanks), closing this one.

PR #2644, opened 2021-11-25
sysrepo BUGFIX build with musl libc fails due to unknown type name 'm…

hthiery:
while updating buildroot package this failure was observed with musl libc


michalvasko:
Fine, thanks.

PR #2647, opened 2021-11-27
Fix build when not using CMake's make generator

jktjkt:
CMake is a generator which can be used with other tools than just `make` -- `ninja` is a good candidate which tends to build a little faster. It's therefore a bug to try invoking `make` on some target directly.

Test fixtures were introduced to CMake in 3.7 (which means that the version comparison operator was too pessimistic). Since they only work on top of other tests, not targets, the `tests_clean` needs to become a test. However, in that case it might get happily executed in parallel to the usual testing process, which needs to be avoided via that `RUN_SERIAL` test property. That is not an equivalent to fixtures, but at least it's enough to serialize its execution with the rest of the test suite. As long as it is listed as the last test in this file, it should work reasonably even on older CMake.

Fixes: 08659bdf tests FEATURE clean all files created by tests

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -92,6 +92,5 @@ endif()
 
 # phony target for clearing all sysrepo test data
 add_custom_target(test_clean
-    COMMAND rm -rf ${PROJECT_BINARY_DIR}/test_repositories
-    COMMAND rm -rf /dev/shm/_tests_sr_*
+    COMMAND rm -rf ${PROJECT_BINARY_DIR}/test_repositories /dev/shm/_tests_sr_*
 )
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -28,9 +28,12 @@ endif()
 # generate config
 configure_file("${PROJECT_SOURCE_DIR}/tests/config.h.in" "${PROJECT_BINARY_DIR}/tests/config.h" ESCAPE_QUOTES @ONLY)
 
+set(TEST_CLEAR_STATE_COMMAND rm -rf ${PROJECT_BINARY_DIR}/test_repositories /dev/shm/_tests_sr_*)
 if(${CMAKE_VERSION} VERSION_GREATER "3.7")
     # tests cleanup fixture
-    add_test(NAME tests_done COMMAND make test_clean)
+    add_test(NAME tests_done
+        COMMAND ${TEST_CLEAR_STATE_COMMAND}
+    )
     set_tests_properties(tests_done PROPERTIES FIXTURES_CLEANUP tests_cleanup)
 endif()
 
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -59,7 +53,7 @@ if(ENABLE_TESTS)
             "SYSREPO_REPOSITORY_PATH=${PROJECT_BINARY_DIR}/test_repositories/${test_name}"
             "SYSREPO_SHM_PREFIX=_tests_sr_${test_name}"
         )
-        if(${CMAKE_VERSION} VERSION_GREATER "3.7")
+        if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.7")
```
michalvasko, line 56: I know `cmake` version 3.7 supports this already but `cmake` version < 3.7 does not support `VERSION_GREATER_EQUAL` hence me using `VERSION_GREATER`. I do not really consider it a problem.

michalvasko (changes requested):
Alright, so the problem for you is invoking `make` directly. However, I would prefer following the principle of as few changes as possible and so the whole change would consist of replacing the `make test_clean` command with the actual commands.

If nothing else, it will cover the situation when the tests are manually interrupted on `cmake` < 3.7, there would still be the target `test_clean` available (which I often do except I am using `cmake` >= 3.7).

tests/CMakeLists.txt
jktjkt: Ah, good catch, thanks, I'll fix this one.

jktjkt:
> Alright, so the problem for you is invoking `make` directly. However, I would prefer following the principle of as few changes as possible and so the whole change would consist of replacing the `make test_clean` command with the actual commands.

That would mean that these commands are present twice, which would be a Bad Thing™, but fortunately this thing supports variables.

tests/CMakeLists.txt
```diff
@@ -28,9 +28,12 @@ endif()
 # generate config
 configure_file("${PROJECT_SOURCE_DIR}/tests/config.h.in" "${PROJECT_BINARY_DIR}/tests/config.h" ESCAPE_QUOTES @ONLY)
 
+SET(CLEAR_ALL_COMMAND rm -rf ${PROJECT_BINARY_DIR}/test_repositories /dev/shm/_tests_sr_*)
```
michalvasko, line 31: Please use `set` instead of `SET` for consistency. Also, I think I would prefer variable name `TEST_CLEAN_COMMAND` to be more specific and go with the custom target.

michalvasko (changes requested):
Minor problems.

michalvasko:
> That would mean that these commands are present twice, which would be a Bad Thing

I agree I just thought it cannot be avoided but not sure why I did not think of simple variables. Almost fine now, just some minor annoyances for me, sorry.

michalvasko (approved):

PR #2652, opened 2021-11-30
Do not trigger pull callbacks with SR_SUBSCR_ENABLED

### LINKED ISSUE #2651: Operational pull callbacks are invoked during module_change_subscribe()
When subscribing to a module in the operational data store, using SR_SUBSCR_ENABLED, any/all pull callbacks for that module are invoked.

This seems incorrect, since it is not possible to subscribe to pull data, and callbacks are therefore unlikely to be expecting to use/access the pull data.

I've only verified this with 1.4.148, but I suspect the same behaviour would be observed in 2.x: https://github.com/sysrepo/sysrepo/blob/08659bdfb8a5c2f7ac179bc64e0e5714b4aea337/src/sysrepo.c#L3661

deastoe:
Currently if module_change_subscribe() is used to subscribe to a module
in the operational datastore, all operational data pull callbacks are
invoked for the module (even if not included by the subscription XPath)
if SR_SUBSCR_ENABLED is used.

This is not very intuitive since you cannot subscribe to changes to
pull data, only push data, and the change callbacks are unlikely to
be expecting the pull data. So this causes un-necessary system load
to collect data that is not used.

Therefore we prevent any pull callbacks from being invoked by using
SR_OPER_NO_SUBS when collecting data for the enabled callback.

Fixes #2651

### DIFF
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -2905,7 +2905,7 @@ sr_module_change_subscribe_enable(sr_session_ctx_t *session, struct sr_mod_info_
      * (avoid dead-lock) */
     ly_set_add(&mod_set, (void *)ly_mod, 0);
     if ((err_info = sr_modinfo_add_modules(mod_info, &mod_set, 0, SR_LOCK_READ, SR_MI_PERM_NO, session->sid, NULL,
-            0, 0))) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Right, the ideas seem correct to me, merging. Thanks.

PR #2662, opened 2021-12-02
sysrepoctl BUGFIX init-data argument order

syyyr:
It turns out, that the data_path is actually the second to last
argument, not the third to last (which is memory data).


michalvasko:
Right, thanks.

PR #2676, opened 2021-12-17
build: format-check: invoke `make` in a portable way

jktjkt:
I've added `uncrustify` to my build environment, and as a result the build started failing for me because it tried to invoke `make format-check`. I use the ninja generator for CMake, which means that there's no `Makefile` generated in my build directory.

Fix this by invoking the relevant target via CMake.


michalvasko:
No problem, merging.

PR #2677, opened 2021-12-17
Doc oper subscr sr subscr oper merge

bbergquist0930:
I struggled in how to implement an operational callback that covers multiple leafs of a container in which the container also contained configuration leafs.   I wanted sysrepo to use the data in the running datastore for the configuration leafs at the same time to have the operation callback handle the state leafs.   And I wanted the operational callback to be called only once for the state leafs if the subscription was for the container and the callback provided all of the state leafs. 

I turns out that sysrepo can handle this with the SR_SUBSCR_OPER_MERGE subscription option, just that it is not documented well (at all?) it the developers guide.

Here is my attempt to provide some documentation for this and hopefully help someone else to not struggle like I did.

### DIFF
--- a/doc/subs.dox
+++ b/doc/subs.dox
@@ -107,6 +107,11 @@ they can also be used to provide _configuration_ data nodes. In that case they s
 configuration that is actually in-use by the managed device. Much more details and explanations can be found
 in the [NMDA RFC](@ref rfcs).
 
+Note that if a an operational subscription XPath covers _configuration_ data nodes, then the subscription callback
+ _must_ return those _configuration_ data nodes unless the ::SR_SUBSCR_OPER_MERGE subscription option is used
+ in which case the _configuration_ data nodes from the _running_ datastore will be merged with _configuration_
+ data nodes provded by the subscription callback.
+
 When using these [pull operational data](@ref oper_data) subscriptions, the subscription must be **actively handling events**
 whenever **operational datastore data of the module are requested** (so these 2 operations cannot occur in a single thread,
 so they must have individual subscription structures) to **avoid a time out**. Note that operational datastore data are
--- a/examples/examples.yang
+++ b/examples/examples.yang
@@ -6,9 +6,6 @@ module examples {
         leaf l {
             type string;
         }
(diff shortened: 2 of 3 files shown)


michalvasko:
I am sorry you could not figure this out right away but it is a good practice to go through the whole documentation (including options) of a function. I am fine with any documentation improvements but the text you wrote focuses way too much on the fact that some nodes are configuration, some state. This does not make such a difference, you can provide configuration data, too, in your callback and can "overwrite" (not using the flag) configuration/state data provided by the previous callback, for example. So please adjust the text and if you are not sure how, I will do it.

bbergquist0930:
I am open for suggestions on the text Michal or the pull request can be rejected.  

I do know that I had been through the documentation quite a bit and did not realize the importance of this option especially when handling YANG modules which mix configuration and state data within the same container.   Often the state data within the container will be handled by a callback that provides all of the state data for the container in one callback.  Subscribing for the state data with an operational callback at the container level causes one to have to supply the configuration data as well without this option being set and often, or at least in my case, I did not want my callback to also have to supply the configuration data.  I struggled with this use case and went down the path having separate subscriptions for the individual state data items but then ran into the performance problem of having my callbacks called repeatedly even though I supplied all of the state data.

From the current documentation that is not apparent at least to me looking at the documentation from the outside.   Don't forget you are intimately familiar with how Sysrepo works and what might seem to be apparent sometimes is not. 
 
Don't take that as a criticism because really, the documentation and the actual product are actually quite good.  I would just like to see if I could somehow help the next person along that struggles with understanding how this option truly does help in the use case of YANG modules with mixed configuration and state data within a container.  If that can be done, I think it might  be one less question on the issues list that will come up.

michalvasko:
I have nothing against improving the docs but the text needs to remain correct and avoid misleading the reader. Your text was too focused on your exact use-case which is why I refused to merge it, that is all. As you said, my perspective is completely different from that of an average user so it is the users that must point out the unclear parts. I have tried to improve it.

bbergquist0930:
You document update looks good and I think will really help out!

PR #2679, opened 2021-12-22
Fix perf test

mzagozen:
The performance measurement test mentioned at the end of the readme does not compile with latest version of sysrepo found in the *devel* branch due to some API changes. I fixed it and added the test to CI. The numbers we get out of the test running on github runners probably won't be consistent, but at least we will know if the test works 🤷

### DIFF
--- a/tests/perf.c
+++ b/tests/perf.c
@@ -426,7 +426,7 @@ static int
 test_get_tree(struct test_state *state, struct timespec *ts_start, struct timespec *ts_end)
 {
     int r;
-    struct lyd_node *data;
+    sr_data_t *data;
     char path[64];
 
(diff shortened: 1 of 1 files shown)


mzagozen:
You can see the new CI job running in my fork: https://github.com/mzagozen/sysrepo/runs/4608012574?check_suite_focus=true

michalvasko:
Right, we forgot to update this special test. Thanks, the code changes look fine but I do not like the rest so if you want this merged, please remove them. Otherwise I will just fix the compilation myself.

mzagozen:
Sure, I removed all but the actual test code changes.

michalvasko:
Alright, thanks.

PR #2703, opened 2022-01-25
Port NACM from Netopeer2

syyyr:
### Generic implementation of NACM for sysrepo

Hi, this is my PR for NACM in sysrepo.

- [x] Import the source verbatim
- [x] Make it buildable
- [ ] Add helper functions that use NACM automatically (e.g. `sr_nacm_apply_changes`)
- [ ] Figure out tests
- [x] Make this usable in an actual sysrepo app
- [x] Port Netopeer2 to use this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,6 +180,7 @@ set(LIB_SRC
     src/shm_ext.c
     src/shm_mod.c
     src/shm_sub.c
+    src/netconf_acm.c
     src/plugins/ds_lyb.c
     src/plugins/ntf_lyb.c
     src/plugins/common_lyb.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,6 +379,7 @@ install(FILES ${PROJECT_SOURCE_DIR}/src/sysrepo.h ${PROJECT_SOURCE_DIR}/src/sysr
 install(FILES ${PROJECT_SOURCE_DIR}/src/plugins_datastore.h ${PROJECT_BINARY_DIR}/version.h
             ${PROJECT_SOURCE_DIR}/src/utils/values.h ${PROJECT_SOURCE_DIR}/src/utils/xpath.h
             ${PROJECT_SOURCE_DIR}/src/utils/error_format.h
+            ${PROJECT_SOURCE_DIR}/src/netconf_acm.h
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/sysrepo)
 install(FILES ${PROJECT_BINARY_DIR}/sysrepo.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
 
--- a/src/netconf_acm.c
+++ b/src/netconf_acm.c
@@ -31,11 +31,11 @@
 #include "compat.h"
 #include "log.h"
 
-static struct ncac nacm;
+static struct sr_nacm nacm;
 
 /* /ietf-netconf-acm:nacm */
 int
-ncac_nacm_params_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UNUSED(module_name), const char *xpath,
+sr_nacm_nacm_params_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UNUSED(module_name), const char *xpath,
         sr_event_t UNUSED(event), uint32_t UNUSED(request_id), void *UNUSED(private_data))
 {
     sr_change_iter_t *iter;
@@ -118,7 +118,7 @@ ncac_nacm_params_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const ch
 
 /* /ietf-netconf-acm:nacm/denied-* */
 int
-ncac_oper_cb(sr_session_ctx_t *UNUSED(session), uint32_t UNUSED(sub_id), const char *UNUSED(module_name), const char *path,
+sr_nacm_oper_cb(sr_session_ctx_t *UNUSED(session), uint32_t UNUSED(sub_id), const char *UNUSED(module_name), const char *path,
         const char *UNUSED(request_xpath), uint32_t UNUSED(request_id), struct lyd_node **parent, void *UNUSED(private_data))
 {
     LY_ERR lyrc;
@@ -151,8 +151,8 @@ ncac_oper_cb(sr_session_ctx_t *UNUSED(session), uint32_t UNUSED(sub_id), const c
     return SR_ERR_OK;
 }
 
-static struct ncac_group *
-ncac_group_find(const char *group_name)
+static struct sr_nacm_group *
+sr_nacm_group_find(const char *group_name)
 {
     uint32_t i;
 
@@ -167,14 +167,14 @@ ncac_group_find(const char *group_name)
 
 /* /ietf-netconf-acm:nacm/groups/group */
 int
-ncac_group_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UNUSED(module_name), const char *xpath,
+sr_nacm_group_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UNUSED(module_name), const char *xpath,
         sr_event_t UNUSED(event), uint32_t UNUSED(request_id), void *UNUSED(private_data))
 {
     sr_change_iter_t *iter;
     sr_change_oper_t op;
     const struct lyd_node *node;
     const char *group_name, *user_name;
-    struct ncac_group *group = NULL;
+    struct sr_nacm_group *group = NULL;
     uint32_t i;
     char *xpath2;
     int rc;
@@ -221,7 +221,7 @@ ncac_group_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UN
                 break;
             case SR_OP_DELETED:
                 /* find it */
-                group = ncac_group_find(group_name);
+                group = sr_nacm_group_find(group_name);
                 assert(group && nacm.group_count);
 
                 /* delete it */
@@ -251,7 +251,7 @@ ncac_group_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UN
         } else {
             /* name must be present */
             assert(!strcmp(node->parent->child->schema->name, "name"));
-            group = ncac_group_find(lyd_get_value(node->parent->child));
+            group = sr_nacm_group_find(lyd_get_value(node->parent->child));
 
             if (!strcmp(node->schema->name, "user-name")) {
                 if ((op == SR_OP_DELETED) && !group) {
@@ -315,9 +315,9 @@ ncac_group_cb(sr_session_ctx_t *session, uint32_t UNUSED(sub_id), const char *UN
  * @param[in,out] list Rule list to remove from.
  */
 static void
-ncac_remove_rules(struct ncac_rule_list *list)
+sr_nacm_remove_rules(struct sr_nacm_rule_list *list)
 {
-    struct ncac_rule *rule, *tmp;
+    struct sr_nacm_rule *rule, *tmp;
 
     LY_LIST_FOR_SAFE(list->rules, tmp, rule) {
         free(rule->name);
@@ -349,7 +349,7 @@ ncac_remove_rules(struct ncac_rule_list *list)
  * @return > 0 if ptr1 > ptr2.
  */
 static int
-ncac_sort_strcmp_cb(const void *ptr1, const void *ptr2)
+sr_nacm_sort_strcmp_cb(const void *ptr1, const void *ptr2)
 {
     const char **str1, **str2;
 
@@ -371,7 +371,7 @@ ncac_sort_strcmp_cb(const void *ptr1, const void *ptr2)
  * @return -1 if no matching item was found.
  */
 static int32_t
-ncac_strarr_sort_find(const char **item, size_t item_size, char **items, uint32_t item_count)
+sr_nacm_strarr_sort_find(const char **item, size_t item_size, char **items, uint32_t item_count)
 {
     const char **m;
     int32_t idx = -1;
@@ -380,7 +380,7 @@ ncac_strarr_sort_find(const char **item, size_t item_size, char **items, uint32_
         return idx;
     }
 
-    m = bsearch(item, items, item_count, item_size, ncac_sort_strcmp_cb);
+    m = bsearch(item, items, item_count, item_size, sr_nacm_sort_strcmp_cb);
     if (m) {
         idx = ((uintptr_t)m - (uintptr_t)items) / item_size;
     }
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -114,6 +114,8 @@ set(SYSREPO_UMASK "00000" CACHE STRING "Umask used for any files created by sysr
 # global group settings
 set(SYSREPO_GROUP "" CACHE STRING "System group that will own all sysrepo-related files. If empty, the specific process group will be kept.")
 
+set(NACM_RECOVERY_UID 0 CACHE STRING "NACM recovery session UID that has unrestricted access")
```
michalvasko, line 117: Not sure if it needs to be separate, I would just unite it with `SYSREPO_SUPERUSER_UID`.

src/log.h
```diff
@@ -33,6 +33,7 @@
         sr_errinfo_new(err_info, rc, "Callback \"%s\" of plugin \"%s\" for module \"%s\" failed.", \
         cb_name, plugin_name, mod_name)
 
+#define SR_LOG_ERR(...) sr_log(SR_LL_ERR, __VA_ARGS__)
```
michalvasko, line 36: There is a reason why it is missing, you should be using `sr_errinfo_new()` for errors.

src/netconf_acm.h
```diff
@@ -0,0 +1,172 @@
...
+ * Copyright (c) 2017 - 2021 CESNET, z.s.p.o.
+ *
+ * This source code is licensed under BSD 3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#ifndef NP2SRV_NETCONF_ACM_H_
```
michalvasko, line 17: Obviously, needs to be updated.

src/netconf_acm.h
```diff
@@ -0,0 +1,172 @@
...
+};
+
+enum ncac_access {
+    NCAC_ACCESS_DENY = 1,           /**< access to the node is denied */
+    NCAC_ACCESS_PARTIAL_DENY = 2,   /**< access to the node is denied but it is a prefix of a matching rule */
+    NCAC_ACCESS_PARTIAL_PERMIT = 3, /**< access to the node is permitted but any children must still be checked */
+    NCAC_ACCESS_PERMIT = 4          /**< access to the node is permitted with any children */
+};
+
+#define NCAC_ACCESS_IS_NODE_PERMIT(x) ((x) > 2)
```
michalvasko, line 102: It will now be part of *sysrepo* so it needs `sr_` prefix just like everything else, all the function in the internal API at least.

syyyr: done

src/netconf_acm.h
syyyr: done

src/log.h
syyyr: done

CMakeLists.txt
syyyr: done

src/netconf_acm.c
```diff
@@ -0,0 +1,1777 @@
...
+    LY_ERR lyrc;
+    char num_str[11];
+
+    assert(*parent);
+
+    /* NACM LOCK */
+    pthread_mutex_lock(&nacm.lock);
+
+    if (!strcmp(path, "/ietf-netconf-acm:nacm/denied-operations")) {
+        sprintf(num_str, "%u", nacm.denied_operations);
```
michalvasko, line 137: My fault but you can fix it, should be `"%" PRIu32` instead.

michalvasko (commented):
The current changes seem fine to me.

src/netconf_acm.c
```diff
@@ -0,0 +1,1833 @@
...
+
+    /* 1) NACM is off */
+    if (!nacm.enabled) {
+        return 1;
+    }
+
+    /* 2) recovery session allowed */
+    err_info = sr_nacm_getpwnam(user, &user_uid, NULL);
+    if (err_info) {
+         /* FIXME: What if sr_nacm_getpwnam returned an error? What to do now? */
```
michalvasko, line 1030: This is a valid situation for TLS sessions, for example, when the username is not a system user. Why did you change the functionality compared to *netopeer2*? As I have said before, assume it all works correctly unless you have a test that does not work.

It seems I will have to go through the whole file and compare it to what is now in *netopeer2* before merging, you are breaking things with these changes.

src/netconf_acm.c
```diff
@@ -0,0 +1,1833 @@
...
+    SR_OPER_SUBSCR(session, mod_name, xpath, sr_nacm_oper_cb);
+
+    xpath = "/ietf-netconf-acm:nacm/denied-notifications";
+    SR_OPER_SUBSCR(session, mod_name, xpath, sr_nacm_oper_cb);
+
+    return error_info;
+}
+
+API void
+sr_nacm_destroy(void)
```
michalvasko, line 930: Just to be safe, it should be possible to call this function several times, so please zero all the freed data.

src/netconf_acm.c
```diff
@@ -0,0 +1,1833 @@
...
+
+#define SR_OPER_SUBSCR(session, mod_name, xpath, cb) \
+    rc = sr_oper_get_subscribe(session, mod_name, xpath, cb, NULL, 0, sub); \
+    if (rc != SR_ERR_OK) { \
+        sr_errinfo_new(&error_info, rc, "Subscribing for providing \"%s\" state data failed.", mod_name); \
+        return error_info; \
+    }
+
+API sr_error_info_t *
+sr_nacm_init_callbacks(sr_session_ctx_t *session, sr_subscription_ctx_t **sub)
```
michalvasko, line 896: Is there a reason why this is not part of `sr_nacm_init()`? Also, it would probably be useful to allow using `NO_THREAD` flag for the subscription.

michalvasko (changes requested):

src/netconf_acm.c
syyyr: done

src/netconf_acm.c
syyyr: I assumed that adding more error checking wouldn't hurt, but I probably should've added that in a separate change. I didn't know it would be a problem. Before, `sr_nacm_getpwnam` would just return an int, but I think it does make sense for it to return a standard `sr_error_info_t`, since there can be, for example, allocation errors:
```c
    if (!buf) {
        EMEM;
        return -1;
    }
```
I understand that in the old code, errors would be skipped, because in some situations, it's fine that `sr_nacm_getpwnam` failed. But it also means that the allocation error is also skipped. So, maybe the function should be changed to somehow distinguish when something is an actual error? In the end, the change might be as simple as changing this code:
```c
    if (ret) {
        sr_errinfo_new(&err_info, SR_ERR_OPERATION_FAILED, "Getting user \"%s\" pwd entry failed (%s).", user, strerror(ret));
        free(buf);
        return err_info;
```
to something like this:
```c
    if (ret) {
        SR_LOG_WRN("Getting user \"%s\" pwd entry failed (%s).", user, strerror(ret));
        free(buf);
        return err_info;
```
And then introduce the old return value as an output argument. I would also invert the meaning, the output parameter would be called "found" (1 would be "found" and 0 would be "not found").

The caller would then look something like this:
```c
err_info = sr_nacm_getpwnam(user, &user_uid, NULL, &found);
if (err_info) {
    return 0;
}

if (found && user_uid == SR_SU_UID) {
    return 1;
}
```

Also: even `sr_nacm_allowed_tree` can have an error to return. So maybe the return value should also be `sr_error_info_t` (and its old return value should changed to an output argument)?

michalvasko: Yes, memory allocation errors were ignored before. I suppose it makes sense to check them now (because *sysrepo* is a library) but then I would expect the whole `sr_nacm_allowed_tree()` to return err_info and have an output argument `allowed` or something.

src/netconf_acm.c
syyyr: No, there is no particular reason. I have merged the functions. I have also added custo subscription options

src/netconf_acm.c
syyyr: Okay, I have changed the return values now and (hopefully) restored the old functionality. I have made a separate commit so that the changes can be seen, but I can squash it into the "NACM: Use sr_error_info_t for returns" commit, if you preferred that.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+ * Copyright (c) 2017 - 2021 CESNET, z.s.p.o.
+ *
+ * This source code is licensed under BSD 3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://opensource.org/licenses/BSD-3-Clause
+ */
+#define _GNU_SOURCE
+#define _DEFAULT_SOURCE
```
michalvasko, line 17: Please remove this, it should not be possible and using `_GNU_SOURCE` should be always enough.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+        sr_session_set_error_message(session, "Getting next change failed (%s).", sr_strerror(rc));
+        return rc;
+    }
+
+    return SR_ERR_OK;
+}
+
+#define SR_CONFIG_SUBSCR(session, mod_name, xpath, opts, cb) \
+    rc = sr_module_change_subscribe(session, mod_name, xpath, cb, NULL, opts, \
+            SR_SUBSCR_DONE_ONLY | SR_SUBSCR_ENABLED, sub); \
```
michalvasko, line 876: Instead of `opts` there should be `0`, it is the priority and instead of the following fixed flags there should be `opts`. Also, I would like `sub` to be part of the macro, this is dangerous.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+        sr_errinfo_new(&error_info, rc, "Subscribing for providing \"%s\" state data failed.", mod_name); \
+        return error_info; \
+    }
+
+API sr_error_info_t *
+sr_nacm_init(sr_session_ctx_t *session, sr_subscr_options_t opts, sr_subscription_ctx_t **sub)
+{
+    pthread_mutex_init(&nacm.lock, NULL);
+
+    sr_error_info_t *error_info = NULL;
```
michalvasko, line 894: I **always** use just `err_info` so please stick to that convention.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+ * @brief Get passwd entry of a user, specifically its UID and GID.
+ *
+ * @param[in] user User to learn about.
+ * @param[out] uid User UID, if set.
+ * @param[out] gid User GID, if set.
+ * @param[out] found 1 if user found, 0 if user not found
+ * @return err_info, NULL on success.
+ */
+static sr_error_info_t *
+sr_nacm_getpwnam(const char *user, uid_t *uid, gid_t *gid, int* found)
```
michalvasko, line 971: Formatting, s/int\* /int \*/

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+sr_nacm_allowed_tree(const struct lysc_node *root, const char *user, int* allowed)
+{
+    int user_found;
+    sr_error_info_t *err_info = NULL;
+    uid_t user_uid;
+
+    /* 1) NACM is off */
+    if (!nacm.enabled) {
+        *allowed = 1;
+        return err_info;
```
michalvasko, line 1033: If returning always on success I prefer just `return NULL;`, it makes the fact obvious.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+        err_info = sr_nacm_getpwnam(user, NULL, &user_gid, &found);
+        if (err_info) {
+            goto cleanup;
+        }
+
+        if (!found) {
+            /* no user, no more groups */
+            sr_errinfo_free(&err_info);
+            err_info = NULL;
+        }
```
michalvasko, line 1116: Do not understand the purpose of this code. There should probably just be `goto cleanup;` because there is nothing else to do.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+
+            /* add, if not already there */
+            if (sr_nacm_strarr_sort_add((const char **)&grp.gr_name, sizeof **groups, 1, groups, group_count)) {
+                goto cleanup;
+            }
+        }
+    }
+
+    /* success */
+    err_info = NULL;
```
michalvasko, line 1160: Redundant, please remove.

src/netconf_acm.c
```diff
@@ -0,0 +1,1862 @@
...
+                return err_info;
+            }
+        }
+    }
+
+    return err_info;
+}
+
+API sr_error_info_t *
+sr_nacm_check_diff(const struct lyd_node *diff, const char *user, const struct lyd_node **denied_node)
```
michalvasko, line 1773: As we talked about, please replace functionality with `sr_nacm_apply_changes()` with a new parameter `user`.

src/netconf_acm.h
```diff
@@ -0,0 +1,160 @@
...
+    SR_NACM_ACCESS_DENY = 1,           /**< access to the node is denied */
+    SR_NACM_ACCESS_PARTIAL_DENY = 2,   /**< access to the node is denied but it is a prefix of a matching rule */
+    SR_NACM_ACCESS_PARTIAL_PERMIT = 3, /**< access to the node is permitted but any children must still be checked */
+    SR_NACM_ACCESS_PERMIT = 4          /**< access to the node is permitted with any children */
+};
+
+#define SR_NACM_ACCESS_IS_NODE_PERMIT(x) ((x) > 2)
+
+sr_error_info_t *sr_nacm_init(sr_session_ctx_t *session, sr_subscr_options_t opts, sr_subscription_ctx_t **sub);
+void sr_nacm_destroy(void);
```
michalvasko, line 105: Missing docs.

src/netconf_acm.c
michalvasko: It would probably be best to squash all your commits before merging so do not worry about it now.

src/netconf_acm.c
syyyr: Done

src/netconf_acm.c
syyyr: Okay, I have fixed the error and sub is now part of the macro. However, aren't the fixed flags needed for proper functionality of the callbacks? As in, they assume these flags will be used. Or maybe your point is that I should OR those flags at the location the marco is used?

michalvasko: I sort-of stopped thinking when I say the parameter problem. Yes, the 2 flags need to be set and `SR_SUBSCR_PASSIVE` must not be. The only flag that should actually be supported is `SR_SUBSCR_NO_THREAD` so maybe put that in doxygen and check the `opts` param accordingly returning inval arg is some other options are set.

syyyr: Yeah, that makes sense. `SR_CHECK_ARG_APIRET` macro returns an `int`, I wonder if all the functions in the API should return an int? What would happen with the errors then

src/netconf_acm.c
syyyr: Will do

src/netconf_acm.c
michalvasko: Oh, right, I have missed that. Yes, so these functions should all return `int`, `error_info` is used only internally and set to `session` when passing API. But `session` is not passed to these functions even though `user` is. One solution would be to pass `session` instead of `user` but then the NACM user would need to be stored in the `session` like I was suggesting before. I do not really have a problem with this.

PR #2730, opened 2022-02-21
Draft: Filter subscription notification at producer side

irfanHaslanded:
When a producer creates some changes for a certain module
sysrepo notifies all the subscriber threads, which then
filter based on the xpath, whether the change is relevant or not

By doing the filtering at the producer side, we can reduce the
overhead of notifying many threads and optimize resource use.

This should reduce the overhead of socket notification, locks
and improve commit times.

This is enabled through a new option SR_SUBSCR_PRODUCER_FILTERING
that must be used at the time of subscription.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -33,6 +33,8 @@
 #include <time.h>
 #include <unistd.h>
 
+static int sr_shmsub_change_listen_has_diff(const char *xpath, const struct lyd_node *diff);
+
 sr_error_info_t *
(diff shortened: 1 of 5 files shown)


michalvasko:
Looking at the code, I would say you have added the filtering on the producer side but have **not** removed it on the subscriber side so it works as before except you do some additional redundant filtering. If you remove it, I think it will stop working and the problem I have mentioned before will occur. Not sure why you decided to start implementing some changes with me not agreeing them being possible (the way described) but that is up to you.

PR #2744, opened 2022-03-04
Let sysrepo-plugind write a pid file

jorgensigvardsson:
In systems where daemons depend on each other, it is useful to leave pid
files under /var/run so that init (PID 1) can track dependencies between
daemons, and thus orchestrate their startup times.

The implementation is inspired by the implementation found in netopeer2.

Signed-off-by: Jörgen Sigvardsson <jorgen.sigvardsson@gmail.com>

### DIFF
--- a/src/executables/sysrepo-plugind.c
+++ b/src/executables/sysrepo-plugind.c
@@ -438,17 +438,23 @@ main(int argc, char **argv)
     sr_session_ctx_t *sess = NULL;
     sr_log_level_t log_level = SR_LL_ERR;
     int plugin_count = 0, i, r, rc = EXIT_FAILURE, opt, debug = 0;
+    int pidfd = -1;
+    char pid[8] = { 0 };
+    int pid_len;
+    const char *pidfile = SRPD_PIDFILE;
+    int pidfile_specifid = 0;
     struct option options[] = {
         {"help",      no_argument,       NULL, 'h'},
         {"version",   no_argument,       NULL, 'V'},
         {"verbosity", required_argument, NULL, 'v'},
         {"debug",     no_argument,       NULL, 'd'},
+        {"pid-file",  required_argument, NULL, 'p'},
         {NULL,        0,                 NULL, 0},
     };
 
     /* process options */
     opterr = 0;
-    while ((opt = getopt_long(argc, argv, "hVv:d", options, NULL)) != -1) {
+    while ((opt = getopt_long(argc, argv, "hVv:dp:", options, NULL)) != -1) {
         switch (opt) {
         case 'h':
             version_print();
@@ -479,6 +485,10 @@ main(int argc, char **argv)
             break;
         case 'd':
             debug = 1;
+            break;
+        case 'p':
+            pidfile = strdup(optarg);
+            pidfile_specifid = 1;
             break;
         default:
             error_print(0, "Invalid option or missing argument: -%c", optopt);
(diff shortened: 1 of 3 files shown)


jorgensigvardsson:
I had worked against an older version of the master branch, hence I got conflicts. When I rebased my branch onto an updated master, I could see that similar changes had been made for systemd. Should I take systemd into account? If running systemd, you probably don't want pid files left and right? Let me know, and I'll adjust for systemd.

src/executables/sysrepo-plugind.c
```diff
@@ -518,6 +528,34 @@ main(int argc, char **argv)
...
+            goto cleanup;
+        }
+
+        if (lockf(pidfd, F_TLOCK, 0) < 0) {
+            if ((errno == EACCES) || (errno == EAGAIN)) {
+                error_print(0, "Another instance of the sysrepo-plugind is running.");
+            } else {
+                error_print(0, "Unable to lock the PID file \"%s\" (%s).", pidfile, strerror(errno));
+            }
+            goto cleanup;
```
jktjkt, line 555: This change will effectively make it impossible to run `sysrepo-plugind` as non-`root` by default. Since `/var/run` is not writable by all, unless people take an explicit action to build with an option to put the PID file into a writable location, `sysrepo-plugind` with no arguments will fail to start on a developer's box.

Instead of a build-time option for a default PID file location, and then a runtime option for overriding that, why not just do nothing by default, *and* allow the user to pass a flag like `--pid-file=/var/run/sysrepo-plugind.pid` which will activate this mechanism?

src/executables/sysrepo-plugind.c
```diff
@@ -480,6 +486,10 @@ main(int argc, char **argv)
         case 'd':
             debug = 1;
             break;
+        case 'p':
+            pidfile = strdup(optarg);
```
jktjkt, line 490: There's no matching call to `free(pidfile)`, so this leaks memory.

src/executables/sysrepo-plugind.c
```diff
@@ -438,17 +438,23 @@ main(int argc, char **argv)
     sr_session_ctx_t *sess = NULL;
     sr_log_level_t log_level = SR_LL_ERR;
     int plugin_count = 0, i, r, rc = EXIT_FAILURE, opt, debug = 0;
+    int pidfd = -1;
+    char pid[8] = { 0 };
```
jktjkt, line 442: This is fragile. On Linux, the upper bound on the maximal PID values [has been increased before](https://lwn.net/Articles/10424/), and the fact that it's currently limited to 4M, and therefore will *currently* fit into seven digits plus the trailing `NULL` is just an implementation detail.

jktjkt (changes requested):
Hi, the development happens in the `devel` branch, so PRs should go there.

Since I'm personally very happy with `systemd`, I think that implementing this daemonization and PID-file-writing code in each and every daemon is a bit tedious, but that's probably my personal bias -- and this binary will already daemonize by default anyway.

That said, I think that the code will be simpler and more straightforward if the PID file writing is explicitly activated by a runtime option -- with no default location.

src/executables/sysrepo-plugind.c
```diff
@@ -553,6 +606,11 @@ main(int argc, char **argv)
     rc = EXIT_SUCCESS;
 
 cleanup:
+    if (pidfile >= 0) {
```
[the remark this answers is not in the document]
jorgensigvardsson: Thanks for spotting that one!

src/executables/sysrepo-plugind.c
jorgensigvardsson: Great idea, thanks!

src/executables/sysrepo-plugind.c
jorgensigvardsson: I was going to rely on the OS to reclaim memory, but I can see memory checking tools report it.

jorgensigvardsson:
Should I make the pid file a compile time option, just like the systemd integration? Either way works for me (I'm building from source).

Also, I'll rebase onto the devel branch.

By the way, do you know the state of all components involved (to get netopeer2, sysrepo and libyang) on the devel branch(s)? I tried verify my changes by building from master on each, but got some bad error messages about contexts being invalid and whatnot. It could've been me though, because it was Friday evening after a long week. 🥴  Ended up having to comment out irrelevant code paths in srpd and test just the pidfile code path.

Thanks for the reviewn

jktjkt:
> Should I make the pid file a compile time option, just like the systemd integration? Either way works for me (I'm building from source).

Better wait for @michalvasko's reply, he's the maintainer.

> By the way, do you know the state of all components involved (to get netopeer2, sysrepo and libyang) on the devel branch(s)? I tried verify my changes by building from master on each, but got some bad error messages about contexts being invalid and whatnot. It could've been me though, because it was Friday evening after a long week. woozy_face Ended up having to comment out irrelevant code paths in srpd and test just the pidfile code path.

Our team tends to build from `devel` of everything quite regularly, so I don't really care that much about `master` or about releases, but I think the maintainers do care :). In general, it seems to me that `master` is typically only updated prior to a release, so it "should" be stable. In the `devel` there's an occasional breakage, especially when more intrusive changes are merged -- and the NACM refactoring which went in this Friday looks like a good example of a possible breaking change.

src/executables/sysrepo-plugind.c
jktjkt: It looks like I forgot to mention that I suggest to use `asprintf()` since it's something that's already used throughout the code.

jorgensigvardsson:
Sorry, had to reopen a new pull request: https://github.com/sysrepo/sysrepo/pull/2745

PR #2745, opened 2022-03-06
Let sysrepo-plugind write a pid file

jorgensigvardsson:
In systems where daemons depend on each other, it is useful to leave pid
files under /var/run so that init (PID 1) can track dependencies between
daemons, and thus orchestrate their startup times.

The implementation is inspired by the implementation found in netopeer2.

Signed-off-by: Jörgen Sigvardsson <jorgen.sigvardsson@gmail.com>

### DIFF
--- a/src/executables/sysrepo-plugind.c
+++ b/src/executables/sysrepo-plugind.c
@@ -444,6 +444,55 @@ apply_plugind_module(int plugin_count, sr_session_ctx_t *sess, struct srpd_plugi
     return rc;
 }
 
+static int
+open_pidfile(const char *pidfile)
+{
+    int pidfd;
+    
+    pidfd = open(pidfile, O_RDWR | O_CREAT, 0640);
+    if (pidfd < 0) {
+        error_print(0, "Unable to open the PID file \"%s\" (%s).", pidfile, strerror(errno));
+        return -1;
+    }
+
+    if (lockf(pidfd, F_TLOCK, 0) < 0) {
+        if ((errno == EACCES) || (errno == EAGAIN)) {
+            error_print(0, "Another instance of the sysrepo-plugind is running.");
+        } else {
+            error_print(0, "Unable to lock the PID file \"%s\" (%s).", pidfile, strerror(errno));
+        }
+        close(pidfd);
+        return -1;
+    }
+
+    return pidfd;
+}
+
+static int
+write_pidfile(int pidfd)
+{
+    char pid[30] = { 0 };
+    int pid_len;
+
+    if (ftruncate(pidfd, 0)) {
+        error_print(0, "Failed to truncate pid file (%s).", strerror(errno));
+        return -1;
+    }
+
+    if (snprintf(pid, sizeof(pid) - 1, "%ld\n", (long) getpid())) {
+        error_print(0, "Failed to allocate memory for pid (%s).", strerror(errno));
+        return -1;
+    }
+
+    pid_len = strlen(pid);
+    if (write(pidfd, pid, pid_len) < pid_len) {
+        error_print(0, "Failed to write PID into pid file (%s).", strerror(errno));
+        return -1;
+    }
+
+    return 0;
+}
+
 int
 main(int argc, char **argv)
 {
@@ -460,12 +509,16 @@ main(int argc, char **argv)
         {"verbosity",      required_argument, NULL, 'v'},
         {"debug",          no_argument,       NULL, 'd'},
         {"plugin-install", required_argument, NULL, 'P'},
-        {NULL,        0,                 NULL, 0},
+        {"pid-file",       required_argument, NULL, 'p'},
+        {NULL,             0,                 NULL, 0},
     };
+
+    int pidfd = -1;
+    const char *pidfile = NULL;
 
     /* process options */
     opterr = 0;
-    while ((opt = getopt_long(argc, argv, "hVv:dP:", options, NULL)) != -1) {
+    while ((opt = getopt_long(argc, argv, "hVv:dP:p:", options, NULL)) != -1) {
         switch (opt) {
         case 'h':
             version_print();
@@ -515,6 +568,9 @@ main(int argc, char **argv)
 
             rc = EXIT_SUCCESS;
             goto cleanup;
+        case 'p':
+            pidfile = optarg;
+            break;
         default:
             error_print(0, "Invalid option or missing argument: -%c", optopt);
             goto cleanup;
@@ -524,6 +580,10 @@ main(int argc, char **argv)
     /* check for additional argument */
     if (optind < argc) {
         error_print(0, "Redundant parameters");
+        goto cleanup;
+    }
+
+    if (pidfile && (pidfd = open_pidfile(pidfile)) < 0) {
         goto cleanup;
     }
 
@@ -567,6 +627,11 @@ main(int argc, char **argv)
     sd_notify(0, "READY=1");
 #endif
 
+    /* Now this process is ready, so let's go ahead and update pid file */
+    if (pidfile && write_pidfile(pidfd) < 0) {
+        goto cleanup;
+    }
+
     /* wait for a terminating signal */
     pthread_mutex_lock(&lock);
     while (!loop_finish) {
@@ -588,6 +653,11 @@ main(int argc, char **argv)
     rc = EXIT_SUCCESS;
 
 cleanup:
+    if (pidfd >= 0) {
+        close(pidfd);
+        unlink(pidfile);
+    }
+
     for (i = 0; i < plugin_count; ++i) {
         dlclose(plugins[i].handle);
         free(plugins[i].plugin_name);


jorgensigvardsson:
Reopened this PR https://github.com/sysrepo/sysrepo/pull/2744 as a new one. Could not change target branch of PR, and with my poor git skills, I had to redo my own branch (which dropped the old PR).

jorgensigvardsson:
This PR does not always create a PID-file at a default path, instead you have to specify PID with `-p` on the command line.

jorgensigvardsson:
Will replace the `snprintf` of pid to stack buffer with an `asprintf` to heap buffer to avoid surprises when/if PID size changes.

src/executables/sysrepo-plugind.c
```diff
@@ -527,6 +534,34 @@ main(int argc, char **argv)
...
+     * considered up and running. No need to create a pid file if we're in debug
+     * mode (we're not a daemon)
+     */
+
+    if (debug && pidfile) {
+        SRPLG_LOG_WRN("sysrepo-plugind", "Ignoring pidfile in debug mode.");
+    }
+
+    if (!debug && pidfile) {
+        /* make sure we are the only instance - lock the PID file and write the PID */
```
jktjkt, line 548: this comment is misleading because *this* code doesn't write anything; it only happens later on, after the plugins have been loaded

src/executables/sysrepo-plugind.c
```diff
@@ -527,6 +534,34 @@ main(int argc, char **argv)
...
+        pidfd = open(pidfile, O_RDWR | O_CREAT, 0640);
```
jktjkt, line 549: You can use `O_TRUNC` here and save a later `ftruncate()`. Also, since the comments suggest that the code is trying to only write out a PID once all plugins have loaded, that's one more reason to truncate  that file at the earliest possible time. (But the code is not syncing the IO anyway, so it probably doesn't matter that much.)

src/executables/sysrepo-plugind.c
```diff
@@ -567,6 +602,30 @@ main(int argc, char **argv)
...
+            error_print(0, "Failed to truncate PID file (%s).", strerror(errno));
+            goto cleanup;
+        }
+
+        if (asprintf(&pid, "%d\n", getpid()) < 0) {
+            error_print(0, "Failed to allocate memory for pid (%s).", strerror(errno));
+            goto cleanup;
+        }
+
+        pid_len = (int) strlen(pid);
```
jktjkt, line 618: - `strlen()` returns `size_t` and `write(2)` accepts a `size_t`, so why is this casting to an `int`?

src/executables/sysrepo-plugind.c
```diff
@@ -527,6 +534,34 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
+    /* open and lock pid file before plugins are loaded, but don't update it yet 
+     * plugins may fail, and do some initialization that must finish before we are
+     * considered up and running. No need to create a pid file if we're in debug
+     * mode (we're not a daemon)
```
jktjkt, line 540: nit: In general, I prefer comments at the place where they "match the code". I.e., in this case the explanation why using the debug mode and asking for a PID writeout doesn't make sense should IMHO go to that `SRPLG_LOG_WRN` branch (and it should not just say the same thing as the code, but rather explain *why*, otherwise I suggest to simply remove that comment)

src/executables/sysrepo-plugind.c
```diff
@@ -567,6 +602,30 @@ main(int argc, char **argv)
...
+
+        pid_len = (int) strlen(pid);
+        if (write(pidfd, pid, pid_len) < pid_len) {
+            error_print(0, "Failed to write into PID file (%s).", strerror(errno));
+            goto cleanup;
+        }
+        close(pidfd);
+        pidfd = -1;
+        free(pid);
+        pid = NULL;
```
jktjkt, line 626: why not rely on the `cleanup`?

jktjkt (changes requested):

src/executables/sysrepo-plugind.c
jorgensigvardsson: Because of the comparison with the return value of `write` below. I figured the cast there is more justifiable than in the condition expression below. Together with the line(s) above it, it's obvious that the cast is safe, while
```c
if (write(pidfd, pid, pid_len) < (int) pid_len) {
```
would cause more cognitive work I would imagine...

src/executables/sysrepo-plugind.c
jorgensigvardsson: I don't want to trigger observers like `finit` (an init package: https://github.com/troglobit/finit) that allows the use pid files for conditions.

src/executables/sysrepo-plugind.c
```diff
@@ -567,6 +602,30 @@ main(int argc, char **argv)
...
 
+    /* update pid file (if daemon) */
+    if (!debug && pidfile) {
+        int pid_len;
+        if (ftruncate(pidfd, 0)) {
+            error_print(0, "Failed to truncate PID file (%s).", strerror(errno));
+            goto cleanup;
+        }
+
+        if (asprintf(&pid, "%d\n", getpid()) < 0) {
```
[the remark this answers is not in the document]
jorgensigvardsson: Yeah, I read that man page too. It's probably very vague because it isn't part of any standard. I couldn't imagine that it would leave a dangling pointer in `*strp`, that would just be nuts!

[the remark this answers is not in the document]
jorgensigvardsson: I dunno about the ANSI C file streams. The trade off is more code. The `0640` permission flags in the call to `open` has to be set in yet another function call `fchmod`. It and and `lockf` requires the use of `fileno` in order to get to the file descriptor. `fchmod` also requires error checking. I'm not sure all that extra code adds any value because of the `asprintf` which likely isn't going to fail ever.

src/executables/sysrepo-plugind.c
jorgensigvardsson: Correction: I do want to trigger observers like that, but not until _after_ plugins have been initialized.

src/executables/sysrepo-plugind.c
jorgensigvardsson: Comment(s) updated. I also removed the `close(pidfd)` as it removes the lock!

src/executables/sysrepo-plugind.c
jorgensigvardsson: Well, I figured that there's no need to hang on to memory if it's not needed. Old habit... of course, this is a very small chunk of memory so it doesn't really matter in most systems.

src/executables/sysrepo-plugind.c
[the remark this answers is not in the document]
jorgensigvardsson: (and when it does fail, it really shouldn't return a dangling pointer in `*strp`, and I think that would've been called upon by now!)

src/executables/sysrepo-plugind.c
```diff
@@ -527,6 +534,34 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
+    /* No need to create a pid file if we're in debug mode (we're not a daemon) */
+
+    if (debug && pidfile) {
+        SRPLG_LOG_WRN("sysrepo-plugind", "Ignoring pidfile in debug mode.");
+    }
```
michalvasko, line 541: This seems redundant to me. Users specify all the parameters and if they use both `-d` and `-p` I would assume it is what they wanted to do.

src/executables/sysrepo-plugind.c
```diff
@@ -515,6 +519,9 @@ main(int argc, char **argv)
 
             rc = EXIT_SUCCESS;
             goto cleanup;
+        case 'p':
+            pidfile = strdup(optarg);
```
michalvasko, line 523: Totally redundant and leaking `strdup()`.

src/executables/sysrepo-plugind.c
```diff
@@ -527,6 +534,34 @@ main(int argc, char **argv)
         goto cleanup;
     }
 
+    /* No need to create a pid file if we're in debug mode (we're not a daemon) */
+
+    if (debug && pidfile) {
+        SRPLG_LOG_WRN("sysrepo-plugind", "Ignoring pidfile in debug mode.");
+    }
+
+    if (!debug && pidfile) {
```
michalvasko, line 543: This and the later new code block could be put into functions to avoid cluttering main with all this.

michalvasko (changes requested):
Agreeing with @jktjkt and believing you should use `systemd` for process synchronization, I am fine with adding this feature if properly organized since you made the effort of making a PR.

src/executables/sysrepo-plugind.c
jorgensigvardsson: Is it really redundant? `optarg` is a global variable defined by `getopt`, and I don't want to just copy the pointer as it will most likely change over time. I did miss the `free(pidfile)` call, but i'ts been added now.

src/executables/sysrepo-plugind.c
jorgensigvardsson: The idea was that if you're running it in debug mode, it's most likely not running in the context of `init` or some other daemon orchestrating process. You know the pid in debug mode. I have however removed it per request.

src/executables/sysrepo-plugind.c
michalvasko: The pointer does change but the string itself does not, the parsed parameters are kept for the whole lifetime of the process. It is safe to use it directly.

src/executables/sysrepo-plugind.c
michalvasko: I do not think trying to guess what the users wants to do is a game you can win, it is best to simply do what the user requests.

src/executables/sysrepo-plugind.c
jorgensigvardsson: Oh, ok, I did not know that!

jorgensigvardsson:
> Agreeing with @jktjkt and believing you should use `systemd` for process synchronization, I am fine with adding this feature if properly organized since you made the effort of making a PR.

In the project I'm working on, there is unfortunately no `systemd`, and I have no mandate to change to using it either.

jorgensigvardsson:
Please let me know if there is anything else I should change! :)

src/executables/sysrepo-plugind.c
```diff
@@ -44,6 +44,11 @@ int loop_finish;
 pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
 pthread_cond_t cond = PTHREAD_COND_INITIALIZER;
 
+/** Variables for PID file handling */
+int pidfd = -1;
+char *pidfile = NULL;
+char *pid = NULL;
+
```
michalvasko, line 51: Never use global variables unless necessary. They should be local just like before and passed as parameters to the functions.

src/executables/sysrepo-plugind.c
```diff
@@ -444,6 +449,64 @@ apply_plugind_module(int plugin_count, sr_session_ctx_t *sess, struct srpd_plugi
...
+        } else {
+            error_print(0, "Unable to lock the PID file \"%s\" (%s).", pidfile, strerror(errno));
+        }
+        return 0;
+    }
+
+    return 1;
+}
+
+static int commit_pidfile()
```
michalvasko, line 475: For a bit more clarity rename to `write_pidfile()`.

src/executables/sysrepo-plugind.c
```diff
@@ -44,6 +44,11 @@ int loop_finish;
 pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
 pthread_cond_t cond = PTHREAD_COND_INITIALIZER;
 
+/** Variables for PID file handling */
+int pidfd = -1;
+char *pidfile = NULL;
```
michalvasko, line 49: Should be `const char *`.

michalvasko (changes requested):
Except for the comments, the coding style is quite off. You should be able to use `uncrustify` to fix it.

michalvasko (changes requested):

src/executables/sysrepo-plugind.c
```diff
@@ -444,6 +444,55 @@ apply_plugind_module(int plugin_count, sr_session_ctx_t *sess, struct srpd_plugi
...
+
+static int
+write_pidfile(int pidfd)
+{
+    char pid[30] = { 0 };
+    int pid_len;
+
+    if (ftruncate(pidfd, 0)) {
+        error_print(0, "Failed to truncate pid file (%s).", strerror(errno));
+        return 0;
```
michalvasko, line 479: General rule is to return `0` on success and `1` on error so please change ti that way.

src/executables/sysrepo-plugind.c
```diff
@@ -588,6 +653,11 @@ main(int argc, char **argv)
     rc = EXIT_SUCCESS;
 
 cleanup:
+    if (pidfile) {
```
michalvasko, line 656: I appreciate the intention to unify all the conditions but unfortunately here it will not work, `pidfd` may easily be `-1` even though `pidfile` was set, so please revert this back.

michalvasko (changes requested):
Final minor changes and it should be good to go.

src/executables/sysrepo-plugind.c
jorgensigvardsson: Doh. Not sure what I was thinking there. Thanks for spotting!

michalvasko (approved):
Seems the formatting is still off but if you are not able to fix it, I will merge it manually and run `uncrustify` myself.

michalvasko:
Merged manually.

jorgensigvardsson:
> Seems the formatting is still off but if you are not able to fix it, I will merge it manually and run `uncrustify` myself.

I tried figure out how to use `uncrustify`, but got lost in the details. How would you run it (for future reference)?

michalvasko:
We have added special support for it so if it is found, you can run `make format-check` to check that everything is formatted well and `make format` to actually adjust the formatting of all the source files.

jorgensigvardsson:
Thank you, I'll keep that in mind!

PR #2780, opened 2022-04-11
sysrepo: allow use of SR_SUBSCR_ENABLED with SR_SUBSCR_PASSIVE

deastoe:
This appears to be an API level restriction only, with passive
subscriptions correctly processing enabled events.


michalvasko:
This limitation seems to be based only on principle because passive subscriptions do not enable any data. The terminology was inherited from *sysrepo* v0.7.x and now there are no data being enabled, instead they are present in the `operational` datastore. In any case, I have nothing against allowing these options to be used together.

PR #2782, opened 2022-04-12
Check sr_diff_ly2sr() return to avoid segfault

deastoe:
- sr_diff_ly2sr: initially set return pointer to NULL Help protect against
   memory corruption by ensuring the returned pointer value does not contain
   garbage.

 - sr_modinfo_replace: don't link diff if there was an error Check for an
   error generating a sysrepo diff from a libyang diff, before linking the
   diff with the sr_mod_info_s.

   In this error case the sysrepo diff pointer may not be valid, causing 
   memory corruption and possible segfault.

Diff generation and segfault was observed due to https://github.com/CESNET/libyang/pull/1833


michalvasko:
Thanks, merging.

PR #2785, opened 2022-04-13
sr_diff_set_getnext: add missing set bounds checking

deastoe:
Prevent a segfault observed when (skip) iterating list keys by
checking we are still within the bounds of the set.


michalvasko:
Right, thanks.

PR #2823, opened 2022-05-23
install: create SRPD_PLUGINS_PATH directory

peckato1:
Sysrepo should install the directories it needs, especially when they
are by default in /usr which can be read-only.

Also the directory should be owned by sysrepo packages.


jktjkt (approved):

michalvasko (approved):

michalvasko:
No problem, merged.

PR #2845, opened 2022-06-19
Fix segfault when an NACM group is deleted

yakbay:


michalvasko:
Thanks, but fixed differently.

yakbay:
No problem, thanks!


12 Tem 2022 Sal 10:20 tarihinde Michal Vasko ***@***.***>
şunu yazdı:

PR #2873, opened 2022-07-16
CI: test ASAN/UBSAN with a newer clang

jktjkt:
There's no point in not using the latest & greatest when it comes to diagnostics. The other builds are still on Ubuntu 18.04 for compatibility reasons, I suppose, but for this one we want to latest fixes from upstream.


michalvasko:
No problem, merging.

PR #2884, opened 2022-07-28
modinfo FEATURE add datastore/last-modified to sysrepo-monitoring

gotthardp:
This is just a "draft for comments"

(1) To see if you agree
 - with the approach to show a cummulative timestamp for all modules
 - with the YANG model
 - with the plugins_datastore API function name (it can be timestamp specific like now, or it can be made more generic to retrieve also e.g. repo size)

(2) To ask for a little guidance concerning the module info retrieval in modinfo.c:1321-1327. I feel something is wrong because the list of modules does not have the `mod->ds_plg[SR_DS_STARTUP]` initialized. Am I doing this right, please?


michalvasko:
>  with the approach to show a cummulative timestamp for all modules

I do not see the advantage over showing it for each module separately, so that is what I would do.

> with the YANG model

See previous.

> with the plugins_datastore API function name (it can be timestamp specific like now, or it can be made more generic to retrieve also e.g. repo size)

I would call the function `srds_last_modif` and keep just the timestamp but do not limit the API to `time_t` granularity (seconds), use `struct timespec` instead. Obviously, LYB datastore would always set `tv_nsec` to `0` but some other non-file datastore should be able to report a more precise value.

> To ask for a little guidance concerning the module info retrieval in modinfo.c:1321-1327. I feel something is wrong because the list of modules does not have the mod->ds_plg[SR_DS_STARTUP] initialized. Am I doing this right, please?

Only the datastore plugins needed are resolved, which are `mod_info->ds` and `mod_info->ds2`. So, this needs special handling because you obviously need all the plugins in the specific case of retrieving `operational` data for `sysrepo-monitoring`. It is up to you to find the best place to add this special case handling.

Also, thanks for actually implementing this improvement.

gotthardp:
Thank you. I completed the implementation and renamed the callback to `srds_last_modif` as you suggested.

It resolves all plugins on its own. I was afraid the `mod_info` may even not include all modules.

michalvasko (commented):
Looks fine to me except you kept the summarized last-modification time for the whole datastore instead of returning it for each module separately. Why do you think it is better? It should be quite simple to compute it on your own, if needed. Also, please increase the plugin API version in `plugins_datastore.h`. Lastly, you will need to modify `test_operational` where `sysrepo-monitoring` data are being read and compared.

gotthardp:
> ... you kept the summarized last-modification time for the whole datastore instead of returning it for each module separately. Why do you think it is better? It should be quite simple to compute it on your own, if needed.

Oops, I tried to follow your guidance, but I misunderstood your statement *I do not see the advantage over showing it for each module separately, so that is what I would do.*

In my use-case the user wants to know if a datastore changed. But showing it for each module separately is also fine, just that the calculation of the cumulative value takes more effort, nothing else. I will change it.

gotthardp:
I can add the `datastore` list into each `module` in the `sysrepo-state`. I just feel the pain of the user who whould need to walk through all the modules, for each `module` retrieve the `last-modified` timestamp of the specific datastore and calculate the cummulative value.

How about keeping both the individual and the cummulative?

michalvasko:
I am not sure, *sysrepo* does not fully follow all the YANG principles and splitting the datastores by YANG modules is one of the changes. I do not see the pain of collecting the timestamp for the whole datastore, you just use a simple XPath to get all the relevant nodes and in a trivial loop get the latest value.

src/plugins_datastore.h
```diff
@@ -221,6 +221,17 @@ typedef int (*srds_access_get)(const struct lys_module *mod, sr_datastore_t ds,
  */
 typedef int (*srds_access_check)(const struct lys_module *mod, sr_datastore_t ds, int *read, int *write);
 
+/**
+ * @brief Get the time when the datastore data of the module were last modified.
+ *
+ * @param[in] mod Specific module.
+ * @param[in] ds Specific datastore.
+ * @param[out] mtime Time of last modification.
```
michalvasko, line 229: Mention that `0`s are expected if it is unknown (such as when the file does not exist).

michalvasko (changes requested):

src/plugins_datastore.h
gotthardp: Yes, that ought to be documented.

michalvasko (approved):

michalvasko:
Seems fine to me, thanks.

PR #2893, opened 2022-08-11
modinfo BUGFIX fix null termination of sysrepo-monitoring

gotthardp:
The commit e268fb538cc6e4a15173bcdec01092293e4261d9 dropped the trailing `0x00`, although it probably shouldn't, right?


michalvasko:
Right, it definitely should not.

PR #2958, opened 2022-11-29
Implement api for comparing sysrepo values

irfanHaslanded:


michalvasko (changes requested):
Please use simply `int` instead of `bool`, we do not use that at all for compatibility reasons. Also, run `uncrustify` on the code to format it correctly.

src/utils/values.c
```diff
@@ -374,6 +374,80 @@ sr_dup_values(const sr_val_t *values, size_t count, sr_val_t **values_dup_p)
...
+        return value1->data.uint32_val == value2->data.uint32_val;
+    case SR_UINT64_T:
+        return value1->data.uint64_val == value2->data.uint64_val;
+    case SR_UNKNOWN_T:
+    case SR_CONTAINER_T:
+    case SR_CONTAINER_PRESENCE_T:
+    case SR_LIST_T:
+    case SR_LEAF_EMPTY_T:
+    case SR_NOTIFICATION_T:
+        // No value to compare
```
michalvasko, line 444: We use only `/* */` style comments.

src/utils/values.h
```diff
@@ -117,6 +117,15 @@ int sr_val_build_str_data(sr_val_t *value, sr_val_type_t type, const char *forma
  */
 int sr_dup_val(const sr_val_t *value, sr_val_t **value_dup);
 
+/**
+ * @brief Compare two values.
+ *
+ * @param [in] value1 Sysrepo value to compare
+ * @param [in] value2 Sysrepo value to compare
+ * @param [out] 1 is the values are equal, 0 othewise
```
michalvasko, line 125: Not a `param` but `return`.

src/utils/values.c
```diff
@@ -374,6 +374,80 @@ sr_dup_values(const sr_val_t *values, size_t count, sr_val_t **values_dup_p)
...
+        return 1;
+    }
+
+    return 0;
```
michalvasko, line 448: This is unreachable, right? Please refactor without unreachable code. For example, when there is no value, use just `break` and put `return 1` here.

src/utils/values.h
```diff
@@ -117,6 +117,15 @@ int sr_val_build_str_data(sr_val_t *value, sr_val_type_t type, const char *forma
...
+ */
+int sr_cmp_val(const sr_val_t *value1, const sr_val_t *value2);
```
michalvasko, line 127: `cmp` would suggest that the function returns `0` on equality (`strcmp`) but it does not. So please rename to `sr_equal_val` or invert the return value.

src/utils/values.h
```diff
@@ -117,6 +117,22 @@ int sr_val_build_str_data(sr_val_t *value, sr_val_type_t type, const char *forma
...
+ * @param [in] value2 Sysrepo value to compare
+ * @return 1 if the values are equal, 0 othewise
+ *
+ * @note If any one of the below parameters of the values differ, the values are not consider equal
+ *  1. type
+ *  2. origin
+ *  3. xpath
+ *  4. default flag
+ *  5. actual value
+ */
```
michalvasko, line 133: We generally put this additional information in the `brief` description of the function, before the parameters. Even something like
```
Check if two ::sr_val_t values are equal. They are not considered equal if their type, origin, xpath, default flag, or actual value differ.
```
would be fine.

michalvasko (changes requested):
One last detail, seems fine otherwise.

michalvasko (approved):

PR #2962, opened 2022-12-10
Add SR_SUBSCR_OPER_ONCE option for oper subscr

irfanHaslanded:
This option causes a component subscribing to provide operational pull data to be called no more than once per-subscription for a given get on the operational datastore.

This can be used to optimise certain usecases. For example, if a component is subscribed to provide pull state below a list, the default behaviour causes a component to be called for each element in the list (unless the request was for a specific element in the list).

As the number of elements in the parent list(s) grows it can take a significant amount of time, proportional to the time required to actually collect the data, to issue and handle a callback for each element when data for all (or a large subset of) elements is requested.

This option therefore gives the capability to handle pull requests in a "one shot" fashion where it is more performant to do so.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -1057,10 +1057,10 @@ sr_module_oper_data_update(struct sr_mod_info_mod_s *mod, const char *orig_name,
             }
 
             /* nested data */
-            for (j = 0; j < set->count; ++j) {
+            if (xpath_subs[0].opts & SR_SUBSCR_OPER_ONCE) {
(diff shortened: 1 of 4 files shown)


michalvasko:
I do not really understand the point of this option, to me it seems to only add a fair amount of complexity to the subscription behavior, which is not desired. You can always subscribe to the parent node of the list (as long as it has one but having a top-level list is a bad YANG design), which will be called only once and do whatever you want in the callback, even simulate the behavior of this flag, I think.

PR #2979, opened 2023-01-10
counter based futex for sr_cond_wait

irfanHaslanded:
Avoid busy waiting with EAGAIN due to irrelevant event being broadcast.
The implementation of conditional variable, with just 2 IDLE and READY
values is not efficient.
For example, when a producer is waiting for an event to be processed by
multiple subscribers, the rwlock associated with this subscription will
undergo many cycles of lock, unlock, with each unlock causing a
sr_cond_broadcast. Once the producer receives this, it goes into a
busy spin of sr_cond_wait because the futex is now marked as READY,
although the event (unlock) is not relevant to us.
This can be improved by waiting until futex value changes from the last_val.
By making futex a counter, each thread can wait until a new event occurs,
and use sys_futex_wait to actually wait for a new event.

### DIFF
--- a/src/sr_cond/sr_cond_futex.c
+++ b/src/sr_cond/sr_cond_futex.c
@@ -34,29 +34,16 @@
 #include "log.h"
 #include "sysrepo_types.h"
 
-#define FUTEX_VAL_IDLE  0
-#define FUTEX_VAL_READY 1
-
 sr_error_info_t *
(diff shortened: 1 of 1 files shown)


michalvasko (approved):

michalvasko:
Okay, seems to work well.

PR #2991, opened 2023-01-26
modinfo BUGFIX check NULL return for successive list items

mblacksee:


michalvasko:
I do not think your changes are correct. If you are trying to handle cases when the owner module is `NULL` then your intention is valid but it must be implemented as a separate check. Such nodes are really specific and I am not sure what to do about them, just skipping them does not seems like a proper behavior, maybe even returning an error would be appropriate. Please confirm the use-case and even provide the edit that triggered it if you can and I will try to fix it properly.

mblacksee:
Hi Michal,
Below RPC has triggered the problem. Root cause of the problem is that XML namespace of the private module is provided as `urn:ietf:params:xml:ns:yang:module-gps` in the RPC request. Module's real namespace is actually `http://blabla.com/yang/module-gps`. So that, In `sr_modinfo_collect_edit` function, `strcmp(lyd_owner_module(root)->name` triggers crash.
Thanks.

```
<rpc message-id="m-17" xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
    <edit-config>
        <target>
            <running/>
        </target>
        <default-operation>merge</default-operation>
        <error-option>rollback-on-error</error-option>
        <config>

            <system xmlns="urn:ietf:params:xml:ns:yang:ietf-system"
xmlns:op="urn:ietf:params:xml:ns:netconf:base:1.0" op:operation="merge">

                <hostname>A6-38_test</hostname>

                <ntp>

                    <enabled>true</enabled>

                    <server>

                        <name>NTP Server 1</name>

                        <udp>

                            <address>time.nist.gov</address>

                        </udp>

                    </server>

                </ntp>

            </system>

            <gps xmlns="urn:ietf:params:xml:ns:yang:module-gps"
xmlns:op="urn:ietf:params:xml:ns:netconf:base:1.0" op:operation="merge">

                <config>

                    <user-configuration-enabled>false</user-configuration-enabled>

                </config>

            </gps>

        </config>
    </edit-config>
</rpc>
```

michalvasko:
Okay, should be fixed with a test.

PR #2999, opened 2023-02-06
Lock oper ds while changing running ds

irfanHaslanded:
Add missing WRITE lock in `sr_modinfo_data_store` during update of operational data.
Correct lock order of secondary ds in `sr_shmmod_modinfo_rdlock` and `sr_shmmod_modinfo_wrlock`

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -2560,6 +2560,8 @@ sr_modinfo_data_store(struct sr_mod_info_s *mod_info)
     struct lyd_node *mod_data, *enab_mod_data = NULL, *diff = NULL;
     uint32_t i;
     int change, create_flags;
+    int oper_mod_locked = 0;
+    struct sr_mod_lock_s *shm_lock;
 
     assert(!mod_info->data_cached);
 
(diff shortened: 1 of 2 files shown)


michalvasko (approved):

src/modinfo.c
```diff
@@ -2624,6 +2626,14 @@ sr_modinfo_data_store(struct sr_mod_info_s *mod_info)
                 }
 
                 if (mod_info->ds == SR_DS_RUNNING) {
+                    /* Acquire write lock on oper ds */
+                    shm_lock = &mod->shm_mod->data_lock_info[SR_DS_OPERATIONAL];
+
+                    if ((err_info = sr_rwlock(&shm_lock->data_lock, SR_MOD_LOCK_TIMEOUT,
+                            SR_LOCK_WRITE, mod_info->conn->cid, __func__, sr_shmmod_recover_cb, NULL))) {
```
irfanHaslanded, line 2633: I am not entirely sure about this invocation here.. Is it okay to call this with `NULL` for recovery?

michalvasko: Just look at the code of `sr_shmmod_recover_cb`, obviously not. I have not noticed it.

PR #3062, opened 2023-05-17
cmake FEATURE skip program_check when TAR_BINARY is an argument

gotthardp:
When cross-compiling the program_check finds a wrong tar executable. This patch adds a way to impose a TAR_BINARY path using a cmake argument.


michalvasko:
Based on the [description](https://cmake.org/cmake/help/v3.8/command/find_program.html), it should work this way out-of-the-box. I have also tested it and when I defined the value of `TAR_BINARY`, it was kept even on repeated `cmake` runs. So what exactly is the problem?

gotthardp:
Oh, no, I am an idiot who cannot read properly. :-( Sorry for bothering you.

PR #3064, opened 2023-05-18
Reorder cmocka.h include to fix build error

alexyin1:
On ubuntu 22.04,
After building libyang with -DENABLE_TESTS=ON, 
sysrepo fails to build with -DENABLE_TESTS=ON, with 'make' generating the error
```
[ 32%] Building C object tests/CMakeFiles/test_modules.dir/test_modules.c.o                                                                                                                                                                                                               
In file included from /home/ayin/tmp/sysrepo/tests/test_modules.c:28:                                                                                                                                                                                                                     
/usr/local/include/cmocka.h:2216:42: error: unknown type name 'uintmax_t'
```

Fix was to reorder `#include <cmocka.h>` as suggested by https://github.com/clibs/cmocka/issues/18

I built libyang via
```
cmake -DCMAKE_BUILD_TYPE:String="Release" -DENABLE_TESTS=ON ..
make -Wno -j8
sudo make install
```
and libnetconf2 via
```
cmake -DCMAKE_BUILD_TYPE:String="Release" ..
make -Wno -j8
sudo make install
```
then tried building sysrepo with
```
cmake -DCMAKE_BUILD_TYPE:String="Release" -DENABLE_TESTS=ON ..
make -Wno -j8
sudo make install
```
which worked only after reordering the headers. I ran into a similar problem with libnetconf2 tests, but haven't got a chance to look further.

### DIFF
--- a/tests/test_apply_changes.c
+++ b/tests/test_apply_changes.c
@@ -25,6 +25,7 @@
 #include <string.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <stdint.h>
 
 #include <cmocka.h>
 #include <libyang/libyang.h>
--- a/tests/test_candidate.c
+++ b/tests/test_candidate.c
@@ -21,6 +21,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <stdint.h>
 
 #include <cmocka.h>
 #include <libyang/libyang.h>
--- a/tests/test_context_change.c
+++ b/tests/test_context_change.c
@@ -25,6 +25,7 @@
 #include <string.h>
 #include <sys/types.h>
 #include <unistd.h>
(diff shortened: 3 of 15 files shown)


michalvasko:
The issue you referenced says nothing about header order but rather mentions a missing header so that would be the proper fix. However, I have failed to reproduce the problem on Ubuntu 22.04 so I am wondering what you are doing differently, this would get reported much sooner if it occurred normally.

alexyin1:
Agree with the two above points.
Regarding my specific env,
I'm using a docker image based on Ubuntu-22.04 and built cmocka from source from https://github.com/clibs/cmocka/ via
```
cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Debug ..
make
sudo make install
```

PR #3065, opened 2023-05-19
Include stdint for newer cmocka

alexyin1:
Following up on https://github.com/sysrepo/sysrepo/pull/3064

Root cause of cmocka build issue is that on newer cmocka's `#include <stdint.h>` is no longer present in `cmocka.h`

Newer cmocka releases expect users to include stdint.h themselves prior to including cmocka.h or implement custom types.

[ cmocka 1.1.5](https://github.com/clibs/cmocka/blob/f5e2cd77c88d9f792562888d2b70c5a396bfbf7a/include/cmocka.h#L50)

[ cmocka 1.0.1](https://github.com/clibs/cmocka/blob/12ce9a800c6ff9189c004202f019723405003e1d/include/cmocka.h#L49)

### DIFF
--- a/tests/test_apply_changes.c
+++ b/tests/test_apply_changes.c
@@ -21,6 +21,7 @@
 #include <sched.h>
 #include <setjmp.h>
 #include <stdarg.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
--- a/tests/test_candidate.c
+++ b/tests/test_candidate.c
@@ -18,6 +18,7 @@
 
 #include <setjmp.h>
 #include <stdarg.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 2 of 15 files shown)


michalvasko:
Seems the changed base is not properly propagated into the actions scripts. In any case, the includes need to be ordered within a group for `uncrustify` check to pass so if you want this merged, you would need to update it. Up to you, adding the includes really did not required a PR.

Also, I am missing the point of `cmocka` not including the headers it needs but nothing we can do about that.

alexyin1:
Reordered the headers to be alphabetical. I ran uncrustify manually on the test files via 
`uncrustify -c uncrustify.cfg test/*c`
but there were many unrelated whitespace changes. I also ran with `make format` but seems like files in `/test` directory are not included in the files to format.
This missing header issue exists in libnetconf and possibly netopeer2 as well.

michalvasko:
> files in /test directory are not included in the list of files to format

That is possible, I forgot, sorry about that.

PR #3082, opened 2023-06-15
compat: fix `pthread_mutex_timedlock`

jktjkt:
## refactor: compat: simplify waiting
    
This code looked a bit fishy:
    
- There's no need to check up front whether we're dividing a zero or non-zero number by some constant.
    
- The conversion to milliseconds is actually not needed at all because it gets immediately converted to back to nanoseconds.


michalvasko (approved):

michalvasko:
Okay, seems to be a bit outdated code, not sure what platforms need it.

jktjkt:
> not sure what platforms need it.

TSAN (because it does not have interceptors for these functions, yet).

PR #3154, opened 2023-09-11
bugfix shm_sub new_wrlock race

irfanHaslanded:
For publishing a new event, we use `sr_shmsub_notify_new_wrlock`.
 It does a fake write unlock after acquiring a write lock, to check if
 the last event is still being processed.

 In case of a timeout, we certainly have mutex. But we do not always
 have the write lock. This can happen if,
 A. a reader is still processing the last event.
 B. another publisher wanted to send out its own event and grabbed the
    write lock after our fake unlock.

 If there are no readers, we will take back the write lock and if the
 last publisher crashed, we will recover the sub_shm lock.

 If there are readers, we have lost the write lock.

 We then check if there was a race between our timeout and the
 last_event fully processed.

 If this is the case, we can ignore the timeout and goto event_handled.
 However, this should be done only if
 1. request_id matches
 2. last_event is cleared or is clearable (abort)
 3. there are no readers left.

 Condition 3 was missing and is now added to keep the lock consistent.
 Without it, the following code will attempt to unlock a write lock
 which it doesn't have anymore and fail.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -361,6 +361,7 @@ sr_shmsub_notify_new_wrlock(sr_sub_shm_t *sub_shm, const char *shm_name, sr_sub_
 
     if (!sub_shm->lock.readers[0]) {
         /* FAKE WRITE LOCK */
+        assert(!sub_shm->lock.writer);
         sub_shm->lock.writer = cid;
 
         if (ret == ETIMEDOUT) {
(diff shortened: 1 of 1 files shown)


michalvasko (approved):
Yes, it seems this condition has been skipped when the reader check was added. Thanks.

PR #3160, opened 2023-09-21
edit_diff: fix (leaf-)list instance move checks

lioyerar:
This test is never true except for the first element since the "sibling" is from data tree whereas and the "anchor_node is from "dit_node", use lyd_compare_single to nodes are equal.

Fixes issues 3159

### DIFF
--- a/src/edit_diff.c
+++ b/src/edit_diff.c
@@ -979,7 +979,7 @@ sr_edit_userord_is_moved(const struct lyd_node *data_match, enum insert_val inse
     case INSERT_FIRST:
     case INSERT_AFTER:
         sibling = sr_edit_find_previous_instance(data_match);
-        if (sibling == anchor_node) {
+        if (lyd_compare_single(sibling, anchor_node, LYD_PRINT_WD_ALL) == LY_SUCCESS) {
             /* data_match is after the anchor node (or is the first) */
             return 0;
         }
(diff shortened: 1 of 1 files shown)


jktjkt (changes requested):
Since the corresponding issue has an attached reproducer, can you please add a unit test for this as well?

nit: the commit message is probably missing a word ("*...use lyd_compare_single to **check** that nodes are equal*").

Also, please [link to the issue that you're fixing form the commit message](https://github.blog/2013-01-22-closing-issues-via-commit-messages/).

src/edit_diff.c
```diff
@@ -979,7 +979,7 @@ sr_edit_userord_is_moved(const struct lyd_node *data_match, enum insert_val inse
     case INSERT_FIRST:
     case INSERT_AFTER:
         sibling = sr_edit_find_previous_instance(data_match);
-        if (sibling == anchor_node) {
+        if (lyd_compare_single(sibling, anchor_node, LYD_PRINT_WD_ALL) == LY_SUCCESS) {
```
michalvasko, line 982: What about the flag `LYD_PRINT_WD_ALL`? It is wrong to use here (notice it starts `LYD_PRINT...` and not `LYD_COMPARE...`) and does not map to any relevant flag anyway, so just use `0`. While at it, use just `!lyd_compare_single()` instead of the comparison, redundant.

michalvasko (changes requested):
Thanks for the bugfix but it needs a few changes. The problem is that `anchor_node` can be from the `edit` tree as you correctly observed, so please update also the doxygen of this parameter. Also, what @jktjkt said, having this use-case in the tests would be useful and is quite simple to add, really. Lastly, please change the target branch to `devel`, we do not accept PRs to `master`.

lioyerar:
Close this one, I will open another.

PR #3161, opened 2023-09-22
edit_diff: fix (leaf-)list instance move checks

lioyerar:
This test is never true except for the first element since the "sibling" is from data tree whereas and the "anchor_node" is from "edit_node". Use 'lyd_compare_single' to compare if nodes are equal.

Fixes #3160


lioyerar:
Sorry, wrong number of issue, close it.

michalvasko:
You know, you can edit issues/PRs and/or modify commits and then force-push the PR.

PR #3175, opened 2023-10-05
sysrepo BUGFIX change_sub rd lock iterates wrong

irfanHaslanded:
Upon failure to read lock a single module, we must unlock all the locks already acquired. However the iterator was wrong.


michalvasko:
Yep, thanks.

PR #3176, opened 2023-10-10
sysrepo UPDATE don't check feature '*'

jeremie6wind:
sysrepoctl returned an error when checking the feature '\*' in sr_install_modules_check_features. For example:
sysrepoctl -s yang_directory	-i file.yang -e '\*'
Now the function no longer returns an error anymore.


michalvasko:
Okay, I will also update the documentation where relevant.

PR #3187, opened 2023-10-31
shm_sub bugfix abort notify when subscriber dies

irfanHaslanded:
When a subscriber dies before abort notification is sent, we may not
 have any subscribers left to send the notification.

 So the assertion is not correct.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -1792,7 +1792,11 @@ sr_shmsub_change_notify_change_abort(struct sr_mod_info_s *mod_info, const char 
         notify_subs[notify_count].shm_data_sub.fd = -1;
         ++notify_count;
     }
-    assert(notify_count);
+
+    if (!notify_count) {
+        /* nothing to do , likely because subscriber died recently */
+        goto cleanup;
+    }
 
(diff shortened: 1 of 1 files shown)


irfanHaslanded:
The failure in `test_modules` seems unrelated to this change

michalvasko:
The failed test is fine, obviously some problem with a sanitizer itself. Regarding the change, could you elaborate on the use-case? The SHM subscriptions are being searched and for the whole duration of apply-changes processing a READ lock on them is held so I do not see a way of a `CHANGE` callback failing but the subscription no longer existing for the `ABORT` event.

irfanHaslanded:
> The failed test is fine, obviously some problem with a sanitizer itself. Regarding the change, could you elaborate on the use-case? The SHM subscriptions are being searched and for the whole duration of apply-changes processing a READ lock on them is held so I do not see a way of a `CHANGE` callback failing but the subscription no longer existing for the `ABORT` event.

A subscription process could have died and discovered during `sr_shmsub_change_notify_has_subscription` and removed.

michalvasko (approved):

michalvasko:
Right, fine.

PR #3189, opened 2023-10-31
doc UPDATE make it clear when a subscribed callback will not be called twice

precla:
There is also a note here: https://github.com/sysrepo/sysrepo/blob/master/src/sysrepo_types.h#L471

but if someone misses that note, and only reads the doc for `sr_subscr_flag_t()`, it will be wrongly assumed that `SR_EV_ABORT` is also passed to the callback even if it failed as that is how it's mentioned here


precla:
@michalvasko was faster :)

https://github.com/sysrepo/sysrepo/commit/357c68266342560857a8847383badadbf3ee9796

PR #3191, opened 2023-11-16
NULL check for empty main_shm

irfanHaslanded:
In case a less priviliged process attempts to connect to sysrepo,
 before shm has been properly set up, main_shm file is created but the
 permissions aren't set up correctly. Upon detecting that failure, the
 process leaves an empty main_shm file.

 Any processes which try to connect later to sysrepo must not crash.

 Ideally `sr_open` must not leave a file it could not properly setup.

### DIFF
--- a/src/shm_main.c
+++ b/src/shm_main.c
@@ -489,6 +489,10 @@ sr_shmmain_open(sr_shm_t *shm, int *created)
     }
 
     main_shm = (sr_main_shm_t *)shm->addr;
+
+    /* main_shm can never be NULL */
+    assert(main_shm);
+
     if (creat) {
         /* init the memory */
         main_shm->shm_ver = SR_SHM_VER;
@@ -519,6 +523,10 @@ sr_shmmain_open(sr_shm_t *shm, int *created)
 
 cleanup:
     if (err_info) {
+        if (creat) {
+            /* tried to create but could not setup fully, remove improper shm file */
+            unlink(shm_name);
+        }
         sr_shm_clear(shm);
     } else if (created) {
         *created = creat;


michalvasko:
There is the `CREATE` lock being used when the basic SHM structure is being created that should prevent these situations, can you elaborate more how what you are describing can occur?

irfanHaslanded:
This is not a race condition. 
I suspect it happens, when the first process after boot up to connect to sysrepo does not have the permissions to change the main_shm file permissions after creating it. so, sr_open succeeds in creating the main_shm file, but fails in setting the permissions. (fchmod        EPERM  The effective UID does not match the owner of the file, and the process is not privileged (Linux: it does not have the CAP_FOWNER capability).)

After this, subsequent processes crash dereferencing an empty main_shm

michalvasko:
> first process after boot up to connect to sysrepo does not have the permissions to change the main_shm file permissions after creating it

How could this be? The owner of a file must be able to change its permissions.

irfanHaslanded:
My mistake. It is not `fchmod` but `fchown` that would fail from within `sr_open`

michalvasko:
Okay, that may happen if a `SR_GROUP` is set that a process creating the main SHM does not belong to and is not `root` (which is not expected but can happen), right? In that case I would simply unlink the SHM, which should solve the problem.

michalvasko (changes requested):
Firstly, please remove the original fix, `main_shm` should no longer be ever empty.

As for the new fix, it would remove the file even if it has already existed and has not been created (`O_EXCL` is not used with `O_CREAT`), so please check for the former flag before calling `unlink()` (should work for main SHM creation). Another option is to add unlink to the code creating main SHM, specifically `sr_shmmain_open()` in `shm_main.c` (which I thought you would do), it tracks whether the SHM was created. Both solutions should be fine.

michalvasko (approved):

PR #3192, opened 2023-11-17
sysrepocfg allow shorthand for choosing ds

irfanHaslanded:
Calling `sysrepocfg -d o` is easier than having to type `-d operational`.

### DIFF
--- a/src/executables/sysrepocfg.c
+++ b/src/executables/sysrepocfg.c
@@ -78,8 +78,8 @@ help_print(void)
             "\n"
             "Available options:\n"
             "  -d, --datastore <datastore>  Datastore to be operated on, \"running\" by default (\"running\", \"startup\",\n"
-            "                               \"candidate\", \"operational\", or \"factory-default\"). Accepted by\n"
-            "                               import, export, edit, copy-from op.\n"
+            "                               \"candidate\", \"operational\", or \"factory-default\") or use initials ('r', 's',\n"
+            "                               'c', 'o', or 'f') respectively. Accepted by import, export, edit, copy-from op.\n"
             "  -m, --module <module-name>   Module to be operated on, otherwise it is operated on full datastore.\n"
             "                               Accepted by import, export, edit, copy-from op.\n"
             "  -x, --xpath <xpath>          XPath to select. Accepted by export op.\n"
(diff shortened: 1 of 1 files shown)


samuel-gauthier:
Hi Irfan,

Just an idea, I think you could achieve more or less the same more nicely with shell completion for sysrepo tools.

My two cents,
Samuel

irfanHaslanded:
Hi Samuel,
How do you get shell completion for sysrepo tools?
Although this will probably not be usable from smaller shells such as busybox ash

samuel-gauthier:
For bash, it is explained here: https://github.com/scop/bash-completion
For zsh, here: https://github.com/zsh-users/zsh-completions
You basically have to create a file describing the syntax.
I don't know about ash.

irfanHaslanded:
Thanks Samuel, for the pointers.

IMHO, it is much simpler to just let `sysrepocfg` handle it directly.

src/executables/sysrepocfg.c
```diff
@@ -78,8 +78,9 @@ help_print(void)
             "\n"
             "Available options:\n"
             "  -d, --datastore <datastore>  Datastore to be operated on, \"running\" by default (\"running\", \"startup\",\n"
-            "                               \"candidate\", \"operational\", or \"factory-default\"). Accepted by\n"
-            "                               import, export, edit, copy-from op.\n"
+            "                               \"candidate\", \"operational\", or \"factory-default\").\n"
+            "                               or use initials 'r', 's', 'c', 'o', 'f'\n"
+            "                               Accepted by import, export, edit, copy-from op.\n"
```
michalvasko, line 83: Please keep the correct text flow. There is a leftover period and you continue the sentence, which is not properly finished with a period.

src/executables/sysrepocfg.c
```diff
@@ -709,15 +710,15 @@ arg_is_file(const char *optarg)
...
+    } else if (optarg[0] == 's') {
         *ds = SR_DS_STARTUP;
-    } else if (!strcmp(optarg, "candidate")) {
+    } else if (optarg[0] == 'c') {
         *ds = SR_DS_CANDIDATE;
-    } else if (!strcmp(optarg, "operational")) {
+    } else if (optarg[0] == 'o') {
         *ds = SR_DS_OPERATIONAL;
-    } else if (!strcmp(optarg, "factory-default")) {
+    } else if (optarg[0] == 'f') {
```
michalvasko, line 721: Please restrict the valid values to `running` or `r`, similarly for all the datastores. Accepting any mangled words starting withe the right letters could cause problems in future.

michalvasko (changes requested):
I am fine with this feature and adding auto-completion support seems redundant for this only.

src/executables/sysrepocfg.c
irfanHaslanded: Makes sense, updated

src/executables/sysrepocfg.c
irfanHaslanded:
```sh
Available options:
  -d, --datastore <datastore>  Datastore to be operated on, "running" by default ("running", "startup",
                               "candidate", "operational", or "factory-default") or use initials ('r', 's',
                               'c', 'o', or 'f') respectively. Accepted by import, export, edit, copy-from op.

```

This is how it looks now, Hope that is good.

irfanHaslanded:
Seems like sysrepo UT is broken with latest `libyang`

src/executables/sysrepocfg.c
```diff
@@ -709,15 +709,15 @@ arg_is_file(const char *optarg)
...
+    } else if (!strcmp(optarg, "s") || !strcmp(optarg, "startup")) {
         *ds = SR_DS_STARTUP;
-    } else if (!strcmp(optarg, "candidate")) {
+    } else if (!strcmp(optarg, "c") || !strcmp(optarg, "candidate")) {
         *ds = SR_DS_CANDIDATE;
-    } else if (!strcmp(optarg, "operational")) {
+    } else if (!strcmp(optarg, "o") || !strcmp(optarg, "operational")) {
         *ds = SR_DS_OPERATIONAL;
-    } else if (!strcmp(optarg, "factory-default")) {
+    } else if (!strcmp(optarg, "f") || !strcmp(optarg, "factory")) {
```
michalvasko, line 720: The full name is `factory-default` so please stick to it.

michalvasko (changes requested):

michalvasko (approved):

PR #3205, opened 2023-12-06
Skip dead subscriptions in sr-mon

irfanHaslanded:
sysrepo-monitoring should not report dead subscriptions. Dead subscriptions will be recovered when encountered, but until then reporting them as existing can lead to errors in application code.


michalvasko:
Okay, fine.

PR #3218, opened 2023-12-27
downgrade mod locks to read after storing changes

irfanHaslanded:
In `sr_changes_notify_store`, the mod locks we need are

Step 1. READ_UPGR lock until sr_modinfo_data_store.
    a. update, change and abort notifications.
Step 2. WRITE lock during sr_modinfo_data_store.
Step 3. READ (instead of READ_UPGR) lock after sr_modinfo_data_store.
    a. wait for done notifications

sr_shmmod_modinfo_wrlock_downgrade is called after Step 2. but we only downgrade to READ_UPGR, but only READ lock is necessary.

To enable a full downgrade to READ lock, we must prevent the below race.
1. `_sr_shmsub_notify_wait_wr` of a process P1 for done events. and
2. `sr_shmsub_notify_new_wrlock` of a process P2.

After P1 has notified subscribers, and they have finished processing, if P2 has acquired the sub_shm lock, before P1 has a chance to check.

In this case P2 will go ahead and publish it's own event, causing P1 to timeout incorrectly.

To prevent this, P2 must check that the previous event is not only processssed by subscribers but also that P1 has finished checking the result. We can use the orig_cid field of sr_sub_shm_t for this.

we clear the `orig_cid` field of `sub_shm` after P1 is finished waiting for the event. The next publisher P2 will wait for orig_cid to be 0, before publishing new events.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -2831,6 +2831,15 @@ sr_rwrelock(sr_rwlock_t *rwlock, uint32_t timeout_ms, sr_lock_mode_t mode, sr_ci
         return NULL;
     }
 
+    if (!rwlock->writer) {
+        /* downgrade from upgr lock to read lock */
+        assert(rwlock->upgr == cid);
+        rwlock->upgr = 0;
+        /* broadcast on condition so waiters can grab upgr lock */
+        sr_cond_broadcast(&rwlock->cond);
+        goto cleanup_unlock;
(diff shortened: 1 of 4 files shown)


michalvasko (approved):
Thanks, I think this should work and hopefully cause no additional issues.

PR #3233, opened 2024-01-24
ds json BUGFIX don't rename log files with old notification.

jeremie6wind:
sysrepo uses notification timestamps to rename the log files. The first timestamp in the name is the timestamp of the first notification and the second one is the timestamp of the last notification. However, if the timestamp is in the past, the new file name will be invalid. To avoid losing notification files, don't rename the file with an older timestamp. The file may contain notification with old timestamps.

### DIFF
--- a/src/plugins/ntf_json.c
+++ b/src/plugins/ntf_json.c
@@ -331,9 +331,10 @@ srpntf_rename_file(const char *mod_name, time_t old_from_ts, time_t old_to_ts, t
     sr_error_info_t *err_info = NULL;
     char *old_path = NULL, *new_path = NULL;
 
-    assert(old_to_ts <= new_to_ts);
-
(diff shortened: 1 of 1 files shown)


michalvasko:
Could you please explain the problem in more detail? I do not get it. Maybe using an example.

jeremie6wind:
Yes, this can happen when you jump back in time. Ok I will replace the assert.

michalvasko (changes requested):
Seems good now, thanks. Just please include a note in the comment that it can happen when there is a jump in realtime (because of NTP, for example).

jeremie6wind:
I have changed the comment as you suggested.

michalvasko (approved):

michalvasko:
Alright, thanks.

PR #3238, opened 2024-02-01
Fix race after timeout of high prio done sub

irfanHaslanded:
After a timeout of a higher priority subscriber S1, the event is cleared and the next priority subscribers are notified.
The event and request ID remains the same. Only priority is changed.

So, S1 can still assume the SHM is valid for it, when it is not.

`sr_shmsub_change_listen_relock` should also check for priority when determining SHM is still valid, in addition to event and request ID.

### DIFF
--- a/tests/test_apply_changes.c
+++ b/tests/test_apply_changes.c
@@ -7191,6 +7191,71 @@ test_mult_update(void **state)
     sr_session_stop(sess);
 }
 
+static int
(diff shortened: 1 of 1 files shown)


michalvasko (changes requested):
Have noticed the PR after I committed a fix, so please remove the fix and I can merge the test.

irfanHaslanded:
Updated the commit to just the test.

michalvasko (approved):

michalvasko:
Okay, thanks.

PR #3248, opened 2024-02-20
sysrepo FEATURE sorting of data nodes

lePici:
The Sysrepo should now be adapted to work with sorted data nodes.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -3312,8 +3312,8 @@ sr_shmsub_change_listen_process_module_events(struct modsub_change_s *change_sub
     }
 
     /* parse event diff */
-    if ((err_info = sr_lyd_parse_data(conn->ly_ctx, shm_data_ptr, NULL, LYD_LYB, LYD_PARSE_ONLY | LYD_PARSE_STRICT, 0,
-            &diff))) {
+    if ((err_info = sr_lyd_parse_data(conn->ly_ctx, shm_data_ptr, NULL, LYD_LYB,
+            LYD_PARSE_ONLY | LYD_PARSE_STRICT | LYD_PARSE_ORDERED, 0, &diff))) {
         SR_ERRINFO_INT(&err_info);
         goto cleanup;
     }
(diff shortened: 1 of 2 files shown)


src/context_change.c
```diff
@@ -362,6 +376,9 @@ sr_lycc_del_module(sr_conn_ctx_t *conn, const struct ly_ctx *ly_ctx, const struc
     LY_ERR lyrc;
     int r;
 
+    /* sort removed modules by name because the modules in sr_del_mods are also sorted by name */
+    qsort(mod_set->objs, mod_set->count, sizeof *mod_set->objs, sr_mod_set_qsort_cmp);
```
michalvasko, line 380: I see no reason for this function to be sorting the modules. If it is required and expected, the set should already be created sorted (the function creating the set should return it ordered).

tests/test_modules.c
```diff
@@ -140,12 +140,12 @@ test_install_module(void **state)
...
             "<enabled-feature>feat</enabled-feature>"
-            "<plugin><datastore>ds:startup</datastore><name>JSON DS file</name></plugin>"
-            "<plugin><datastore>ds:running</datastore><name>JSON DS file</name></plugin>"
             "<plugin><datastore>ds:candidate</datastore><name>JSON DS file</name></plugin>"
-            "<plugin><datastore>ds:operational</datastore><name>JSON DS file</name></plugin>"
             "<plugin><datastore>fd:factory-default</datastore><name>JSON DS file</name></plugin>"
             "<plugin><datastore>notification</datastore><name>JSON notif</name></plugin>"
+            "<plugin><datastore>ds:operational</datastore><name>JSON DS file</name></plugin>"
+            "<plugin><datastore>ds:running</datastore><name>JSON DS file</name></plugin>"
+            "<plugin><datastore>ds:startup</datastore><name>JSON DS file</name></plugin>"
```
michalvasko, line 148: Looking at this I am wondering whether identities should not be ordered based on their order in the module, it certainly expresses something in this case (from the simplest and most obvious to more obscure ones). Ideally, the inheritance hierarchy should also be expressed in "DFS" order. But I am not sure it is worth it, would probably not be that simple and may potentially take time to figure out.

michalvasko (changes requested):

michalvasko (approved):

Issue #3255, opened 2024-03-06
How/if to implement a Datastore Plugin


nullromo:
## Background

I just found out about the [datastore plugin feature](https://netopeer.liberouter.org/doc/sysrepo/master/html/sr_plugins.html), and I think it may suit my application. I have a system with multiple user interfaces, including an SSH CLI and a Web GUI, in addition to NETCONF. Right now, I am using (a slightly old version of) sysrepo as the datastore. Whenever a user makes a change on the Web GUI or CLI, it uses the sysrepo library on the backend to write to the running datastore, then calls `sr_copy_config` to sync the changes to the startup datastore. I want to keep the startup datastore and the running datastore in sync at all times, because that's how my CLI and Web GUI users will expect the system to behave.

However, the YANG validation is taking a long time. For instance, when I get an HTTP call from the Web GUI, I can validate the data myself in the javascript backend and write to a JSON file in about 1-2ms, but if I write the data into sysrepo, it takes 200ms to call `sr_apply_changes` because of the validation. Furthermore, it takes an additional 200ms to call `sr_copy_config`. Removing all the `must` statements from the YANG model as a test results in the `sr_apply_changes` call taking about 25ms, which is still a lot slower than just writing the JSON file. So it seems I am taking a big performance hit on my CLI and Web GUI for the sake of supporting NETCONF.

## Questions

There's a few things going on. These are kind of separate issues, but they are related because of the performance issue.

### 1
Is there a way to make changes and apply them to the running and startup datastores at the same time? I don't think there is. However, it seems I may be able to create a Datastore Plugin that causes the startup datastore to just use the running datastore as its backing data. The other options are A) do what I am currently doing, which is to do 2 writes for every write; or B) periodically sync the running data to the startup datastore, which I don't like too much; C) skip validation on the sync, which I don't suspect is allowed; or D) something else (?). If I could eliminate the sync step, it would be a big performance improvement.

### 2
How does one implement a Datastore Plugin? I don't see a clear example. I came across a couple old issues referencing [this (DEAD LINK)](https://github.com/sysrepo/sysrepo/blob/master/src/plugins/ds_lyb.c), but I guess that file was removed. I see there is [some kind of default JSON plugin](https://github.com/sysrepo/sysrepo/blob/d57c49551fbc177bf07451f50461bf4319764692/src/plugins/ds_json.c
[...]

PR #3256, opened 2024-03-07
Add datastore/notification plugin setup information to docs (see #3255)

nullromo:
- Mention how to install and set up datastore/notification plugins.
- Correct internal LYB datastore to internal JSON datastore, since the latter has replaced the former.

### DIFF
--- a/doc/sr_plugins.dox
+++ b/doc/sr_plugins.dox
@@ -4,7 +4,9 @@
 Sysrepo itself supports 2 kinds of plugins not to be confused with `sysrepo-plugind` plugins. **Datastore plugins**
 define a set of callbacks that implement all the operations performed on a datastore allowing to implement a completely
 custom datastore. Every datastore for every YANG module can use a different datastore implementation. **Notification
-plugins** is the exact same concept used for implementing storage of notifications stored for replay.
+plugins** is the exact same concept used for implementing storage of notifications stored for replay. These plugins can
+be installed with `sysrepoctl --plugin-install`. Once installed, `sysrepoctl --module-plugin` can be used to specify
+which datastores and YANG modules will use each plugin.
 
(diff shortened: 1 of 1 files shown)


doc/sr_plugins.dox
```diff
@@ -4,7 +4,9 @@
 Sysrepo itself supports 2 kinds of plugins not to be confused with `sysrepo-plugind` plugins. **Datastore plugins**
 define a set of callbacks that implement all the operations performed on a datastore allowing to implement a completely
 custom datastore. Every datastore for every YANG module can use a different datastore implementation. **Notification
-plugins** is the exact same concept used for implementing storage of notifications stored for replay.
+plugins** is the exact same concept used for implementing storage of notifications stored for replay. These plugins can
```
michalvasko, line 7: In general, `sysrepoctl` and `sysrepocfg` tools are not directly mentioned in the docs because they are just useful utilities that call *sysrepo* API functions and have their own `help`.

doc/sr_plugins.dox
```diff
@@ -21,7 +23,7 @@ on large data. In this case the whole data must be parsed, the change performed,
 back. Instead, another datastore may be able to store the change directly without parsing and printing all the other
 stored data. As for the space required, if implementing a datastore for a specific YANG module, the data structure can
 fully depend on the schema nodes defined in the module and hence avoid storing any redundant information. Finally,
-note that the internal LYB datastore is also implemented as a datastore plugin so it can be used as an example
+note that the internal JSON datastore is also implemented as a datastore plugin so it can be used as an example
```
michalvasko, line 26: This should be updated but there were 3 mentions of `LYB` in this text, not just one.

michalvasko (commented):
For the reasons in the comments and because we accept PRs into `devel` branch only, I have just updated all the `LYB` mentions and will close this PR. Thanks anyway.

PR #3257, opened 2024-03-08
M flag fix

nullromo:
I think initializing the `iitem->module_ds.plugin_name` array to the default values is the right approach, and this code works correctly.

`sysrepoctl --install <module> -m "running:"` will work as expected, with the running datastore disabled for the module.

However, I wasn't able to compile when trying to access `sr_module_ds_default` from `common.h` directly. My knowledge of cmake is still limited, so I was unable to make this work. `sr_module_ds_default_2` is certainly not a clean solution, but it does work. Any suggestion to pull the default values into `sysrepoctl.c` in a compile-able way?


nullromo:
Link issue #3255

src/executables/sysrepoctl.c
```diff
@@ -184,6 +194,8 @@ new_iitem(const char *optarg, sr_install_mod_t **iitems, uint32_t *iitem_count)
     ++(*iitem_count);
 
     iitem->schema_path = optarg;
+    /*memcpy(iitem->module_ds.plugin_name, sr_module_ds_default.plugin_name, sizeof(sr_module_ds_t));*/
```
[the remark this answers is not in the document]
nullromo: > I wasn't able to compile when trying to access sr_module_ds_default from common.h directly.

I didn't clarify properly. Compiling is OK, but linking is the issue. I get the following error:

```
/usr/bin/ld: CMakeFiles/sysrepoctl.dir/src/executables/sysrepoctl.c.o: warning: relocation against `sr_module_ds_default' in read-only section `.text.startup'
/usr/bin/ld: CMakeFiles/sysrepoctl.dir/src/executables/sysrepoctl.c.o: in function `main':
sysrepoctl.c:(.text.startup+0x682): undefined reference to `sr_module_ds_default'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
collect2: error: ld returned 1 exit status
```

I figured out that using the `API` macro will make `sr_module_ds_default` visible, but I'm not sure this is what you want. I don't know if `sr_module_ds_default` _should_ be visible to the external API or not.

Also, adding `common.c` to `SYSREPOCTL_SRC` causes a lot of other errors.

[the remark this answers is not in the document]
michalvasko: > Also, adding common.c to SYSREPOCTL_SRC causes a lot of other errors.

You are right, not sure what I did before that it worked. Okay, so I guess an API function could be added that will return `const` pointer to the default datastore structure and it can then be used with ease in `sysrepoctl`.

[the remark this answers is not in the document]
nullromo: It looks better now with a proper API function.

src/executables/sysrepoctl.1
```diff
@@ -93,7 +93,8 @@ are being changed. \fIMOD-DATASTORE\fP can be:
 .TP
 .BR "\-m\fR,\fP \-\^\-module-plugin \fIMOD-DATASTORE\fP:\fIPLUGIN-NAME\fP"
 Set specific module datastore plugin for a module datastore, can be specified multiple
-times for different module datastores. Accepted by \fBinstall\fP op.
+times for different module datastores. If \fIPLUGIN-NAME\fP is an empty string, it is set to NULL. Accepted by
```
michalvasko, line 96: ... which is only allowed for running datastore. Also mention that whatever datastores are not set will use the default plugin.

src/sysrepo.c
```diff
@@ -1574,6 +1574,11 @@ sr_install_module(sr_conn_ctx_t *conn, const char *schema_path, const char *sear
     return sr_install_module2(conn, schema_path, search_dirs, features, NULL, NULL, NULL, 0, NULL, NULL, 0);
 }
 
+API const sr_module_ds_t *
+sr_get_module_ds_default() {
```
michalvasko, line 1578: Please use `void` as the parameter, also in the prototype.

michalvasko (changes requested):

michalvasko (approved):

src/sysrepo.c
```diff
@@ -1574,6 +1574,11 @@ sr_install_module(sr_conn_ctx_t *conn, const char *schema_path, const char *sear
     return sr_install_module2(conn, schema_path, search_dirs, features, NULL, NULL, NULL, 0, NULL, NULL, 0);
 }
 
+API const sr_module_ds_t *
+sr_get_module_ds_default(void) {
```
michalvasko, line 1578: Opening bracket needs to be on a new line.

nullromo: Whoops, sorry about that. Tried to match the existing style wherever possible

michalvasko:
Okay, thanks, seems good now.

PR #3258, opened 2024-03-11
bugfix shm_sub event type finished instead of none

irfanHaslanded:
change_sub event listeners should use `SR_SUB_EV_FINISHED` instead of `SR_SUB_EV_NONE` to signal that they are done. Only the originators of the event should set it to `SR_SUB_EV_NONE`.

`sr_shmsub_notify_new_wrlock` waits for `SR_SUB_EV_NONE`, and it can assume that the previous event is finished but it's originator may not have seen it yet.

To prevent this, previously `orig_cid` was used, but this is not reliable. If a timeout happens while the listeners have the read lock, the originator cannot clear `orig_cid`. The shm becomes unusable after this as `new_wrlock` will wait forever for `orig_cid == 0`.

Since `sub_shm->event` is an atomic type, we can always set it even though we don't have a WRITE lock, and we already do this.

RPCs will continue to use `SR_SUB_EV_NONE` for `SR_SUB_EV_NOTIF` as the originators don't always wait, and setting event to `SR_SUB_EV_FINISHED` is not needed.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -340,7 +340,7 @@ sr_shmsub_notify_new_wrlock(sr_sub_shm_t *sub_shm, const char *shm_name, sr_sub_
         return err_info;
     }
 
-    if (sub_shm->orig_cid || (ATOMIC_LOAD_RELAXED(sub_shm->event) && (ATOMIC_LOAD_RELAXED(sub_shm->event) != lock_event))) {
+    if (ATOMIC_LOAD_RELAXED(sub_shm->event) && (ATOMIC_LOAD_RELAXED(sub_shm->event) != lock_event)) {
         /* instead of wating, try to recover the event immediately */
         sr_shmsub_recover(sub_shm);
     }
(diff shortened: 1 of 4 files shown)


src/shm_sub.c
```diff
@@ -586,6 +589,7 @@ _sr_shmsub_notify_wait_wr(sr_sub_shm_t *sub_shm, sr_sub_event_t event, uint32_t
  * @param[in] sub_shm Subscription SHM.
  * @param[in] expected_ev  Expected event. Can be:
  *              ::SR_SUB_EV_NONE - just wait until the event is processed, SHM will not be accessed,
+ *              ::SR_SUB_EV_FINISHED - just wait until the event is processed, SHM will not be accessed,
```
michalvasko, line 592: These descriptions are exactly the same and the difference should be explained somewhere, this seems like the best place.

michalvasko (approved):

michalvasko:
Okay, thanks, now it should be possible to understand it all from the code.

PR #3259, opened 2024-03-11
shm_sub update assert for orig_cid

irfanHaslanded:
When an event exists, orig_cid must exist. Upon obtaining a new wr_lock we can clear any previous orig_cid as it is no longer relevant.

This can happen if a previous notifier bailed before waiting for subscribers to complete. It would clear the event but may not have been able to clear orig_cid.


michalvasko (approved):
I can't say I know exactly what use-case are you handling but it seems sensible so okay.

PR #3278, opened 2024-04-26
xpath BUGFIX printing of subtree filter

lePici:
Metadata value was incorrectly printed during subtree filter to XPath conversion. Also, the XPath expression containing the lyds_tree metadata is not printed because it has no use.


michalvasko (approved):
+1 for a test

PR #3287, opened 2024-05-02
build BUGFIX check if sysrepo-plugind is enabled

jeremie6wind:
Add condition to avoid building sysrepo-plugind when ENABLE_SYSREPO_PLUGIND is not on.


michalvasko:
Right, thanks.

Issue #3273, opened 2024-04-17
factory-reset behavior


tbsuht:
Is there currently a way or could an addition be added (runtime or compile option) to customize the factory-default behavior?

> There is an internal subscription to the /ietf-factory-default:factory-reset RPC which performs the copying of factory-default data into all the other datastores.

For our use-case we don't want to copy the factory-default data into all other datastores but rather only `startup`.

jktjkt:
Since that behavior is [described by the RFC](https://datatracker.ietf.org/doc/html/rfc8808#name-factory-reset-rpc), I wonder what exactly your use case is. Or is sysrepo maybe silently discarding, e.g., the pushed operational data from the `operational` DS?

michalvasko:
It was implemented according to the definition of the RPC and the behavior should remain the default one. But I see no problem with adding an optional leaf-list (for example) of only those datastores that should be reset by modifying `sysrepo-factory-default@2023-02-23.yang` and then implementing the functionality. You are welcome to provide a PR.

PR #3289, opened 2024-05-06
factory-reset: allow to specify which datastores are considered

tbsuht:
Implements the possible improvement described in: https://github.com/sysrepo/sysrepo/issues/3273


modules/sysrepo-factory-default@2024-05-02.yang
```diff
@@ -18,19 +18,31 @@ module sysrepo-factory-default {
...
+             automatically populates the leaf-lists with all the modules/datastores being reset.";
+
         leaf-list module {
             type string;
             description
                 "Module that will be reset to its default factory state.";
         }
+
+        leaf-list datastore {
+            type string;
```
michalvasko, line 43: Please use the dedicated type for this, `identityref` with base `ietf-datastores:conventional`.

modules/sysrepo-factory-default@2024-05-02.yang
```diff
@@ -18,19 +18,31 @@ module sysrepo-factory-default {
...
+            "Adds leaf-list of modules and datastores that are reset to the default factory state. If there are none, sysrepo
+             automatically populates the leaf-lists with all the modules/datastores being reset.";
+
         leaf-list module {
             type string;
             description
                 "Module that will be reset to its default factory state.";
         }
+
+        leaf-list datastore {
```
michalvasko, line 42: With this change, it would be proper to improve organization of the nodes and `module` should be enclosed in a container `modules` and `datastore` in `datastores`. Will also make it easier to process these nodes.

modules/sysrepo-factory-default@2024-05-02.yang
```diff
@@ -18,19 +18,31 @@ module sysrepo-factory-default {
...
         description
             "Initial revision.";
     }
 
     augment /fd:factory-reset/fd:input {
         description
-            "Adds leaf-list of modules that are reset to the default factory state. If there are none, sysrepo
-             automatically populates the leaf-list with all the modules being reset.";
+            "Adds leaf-list of modules and datastores that are reset to the default factory state. If there are none, sysrepo
+             automatically populates the leaf-lists with all the modules/datastores being reset.";
```
michalvasko, line 34: Would be good to specify what "all" datastores mean (`running`, `startup`, and `candidate`).

michalvasko (changes requested):
Please rework the changes based on my feedback and also change the target branch to `devel`, we do not accept PRs into `master`.

src/shm_sub.c
```diff
@@ -2242,6 +2242,20 @@ sr_shmsub_oper_get_notify(struct sr_mod_info_mod_s *mod, const char *xpath, cons
     return err_info;
 }
 
+static int
+str2ds(const char *str, sr_datastore_t *ds)
```
michalvasko, line 2246: You can use `sr_ident2mod_ds()` now instead.

src/shm_sub.c
```diff
@@ -2292,6 +2310,20 @@ sr_shmsub_rpc_internal_call_callback(sr_conn_ctx_t *conn, const struct lyd_node
...
     }
+    ly_set_free(set, NULL);
+
+    /* get datastores which should be used for reset */
+    if ((err_info = sr_lyd_find_xpath(input, "datastores/datastore", &set))) {
+        goto cleanup;
+    }
+    for (i = 0; i < set->count; ++i) {
+        if (!str2ds(lyd_get_value(set->dnodes[i]), &ds)) {
+            datastores |= (1 << ds);
```
michalvasko, line 2321: Please no hacks like these, some more readable code instead. For example, `int reset_ds[2]` with either `0` or `1` on `SR_DS_*` index if the specific datastore should be reset or not.

src/shm_sub.c
```diff
@@ -2270,16 +2285,19 @@ sr_shmsub_rpc_internal_call_callback(sr_conn_ctx_t *conn, const struct lyd_node
         return err_info;
     }
 
-    /* collect all required modules */
-    LY_LIST_FOR(lyd_child(input), child) {
+    /* get modules which should be resetted */
+    if ((err_info = sr_lyd_find_xpath(input, "modules/module", &set))) {
```
michalvasko, line 2289: The sets are not really required but I can accept it.

src/sysrepo.c
```diff
@@ -6083,38 +6084,64 @@ sr_rpc_internal_input_update(sr_conn_ctx_t *conn, struct lyd_node *input_op)
...
+        if ((err_info = sr_lyd_new_term(node, ly_srfd_mod, "datastore", "ietf-datastores:startup"))) {
+            goto cleanup;
+        }
+        if ((err_info = sr_lyd_new_term(node, ly_srfd_mod, "datastore", "ietf-datastores:running"))) {
+            goto cleanup;
+        }
+    }
+
+cleanup:
+    ly_set_free(set, NULL);
```
michalvasko, line 6144: Is this really necessary? The modules always needed to be passed to the callback but the datastores should not, unless I am mistaken.

michalvasko (changes requested):

src/sysrepo.c
tbsuht: nope, you are right

src/shm_sub.c
```diff
@@ -2256,9 +2256,10 @@ sr_shmsub_rpc_internal_call_callback(sr_conn_ctx_t *conn, const struct lyd_node
     sr_error_info_t *err_info = NULL, *cb_err_info = NULL;
     struct sr_mod_info_s mod_info;
     struct lyd_node *data[2] = {NULL};
-    const struct lyd_node *child;
     const struct lys_module *ly_mod;
+    struct ly_set *set = NULL;
     sr_datastore_t ds;
+    int reset_ds[8] = {0};
```
tbsuht, line 2262: 8 => because currently `sr_ident2mod_ds` is used as idx. see below

src/shm_sub.c
```diff
@@ -2292,6 +2296,21 @@ sr_shmsub_rpc_internal_call_callback(sr_conn_ctx_t *conn, const struct lyd_node
...
+    if ((err_info = sr_lyd_find_xpath(input, "datastores/datastore", &set))) {
+        goto cleanup;
+    }
+
+    if (set->count == 0) {
+        reset_ds[SR_DS_STARTUP] = 1;
+        reset_ds[SR_DS_RUNNING] = 1;
+    } else {
+        for (i = 0; i < set->count; ++i) {
+            reset_ds[sr_ident2mod_ds(lyd_get_value(set->dnodes[i]))] = 1;
```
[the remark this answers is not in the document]
tbsuht: running, candidate, startup? `ietf-conventional` also allows "intended" but there is no enum for that?

src/shm_sub.c
tbsuht: not sure whether I will find the time this week to rework that, so if you are fine with the current approach, we can leave it as it is

michalvasko: I am in no hurry and `lyd_find_xpath()` really seems unnecessary in this case.

tbsuht: Updated :)

I guess using the set in `sysrepo.c` is fine? There its quite convenient to check for the absence of nodes.

src/shm_sub.c
```diff
@@ -2270,26 +2271,52 @@ sr_shmsub_rpc_internal_call_callback(sr_conn_ctx_t *conn, const struct lyd_node
...
-        ly_mod = ly_ctx_get_module_implemented(conn->ly_ctx, lyd_get_value(child));
-        if (!ly_mod) {
-            sr_errinfo_new(&err_info, SR_ERR_NOT_FOUND, "Module \"%s\" was not found in sysrepo.", lyd_get_value(child));
-            goto cleanup;
-        } else if (!strcmp(ly_mod->name, "sysrepo")) {
-            sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, "Internal module \"%s\" cannot be reset to factory-default.",
-                    lyd_get_value(child));
-            goto cleanup;
-        }
+        if (!strcmp(LYD_NAME(child), "modules")) {
```
michalvasko, line 2275: Just a minor improvement, you can avoid one loop using `lyd_find_path()` (which is not using a set) to find first `modules` and then `datastores` nodes, then you can iterate over all their children.

michalvasko (changes requested):

src/shm_sub.c
tbsuht: updated.

I've added a TODO where I'm not sure why "child->next" is required instead of "child" ... do you know what I'm doing wrong?

michalvasko: Why is `child->next` needed? I do not think it is.

tbsuht: there are two modules nodes found using `sr_lyd_find_path` and only the one used with `->next` has the children

src/sysrepo.c
```diff
@@ -6083,38 +6083,47 @@ sr_rpc_internal_input_update(sr_conn_ctx_t *conn, struct lyd_node *input_op)
...
-        } else if (!sr_module_has_data(ly_mod, 0)) {
-            /* no configuration data */
-            continue;
+    /* check for explicitly defined modules */
+    if ((err_info = sr_lyd_find_xpath(input_op, "modules/module", &set))) {
+        goto cleanup;
+    }
+    if (set->size == 0) {
+        if ((err_info = sr_lyd_new_inner(input_op, ly_srfd_mod, "modules", &node))) {
+            goto cleanup;
```
michalvasko, line 6102: This will be the problem when 2 `modules` containers are created. Just because no `module` exists, it does not mean `modules` does not exist either. Actually, that NP container should always exist because the RPC is validated.

tbsuht: thanks, updated

michalvasko (approved):

michalvasko:
Thanks, seems fine now except some formatting errors. I wanted to refactor it a bit anyway, so I can fix the formatting, too, if you are not able to.

PR #3298, opened 2024-05-20
build: fix wsl check

safesintesi:
At least on Win11 with WSL2, the following command:
```bash
cat /proc/sys/kernel/osrelease
```
returns `<kernel-version>-microsoft-standard-WSL2`.
This is never detected since cmake string FIND operation is case-sensitive.
To avoid future mismatches due to changes by Microsoft, I also directly added a TOLOWER operation so the find is no longer case-sensitive.


michalvasko:
Yes, the fix makes sense and can be merged. But did you actually try *sysrepo* on the current WSL? It was added quite some time ago when many UNIX features were presented as working when they were not and I got tired of wasting my time on such problems. I would expect the situation is much better now.

safesintesi:
Currently we are using sysrepo only on WSL2 with ubuntu 22.04 and everything seems to work.

PR #3300, opened 2024-05-21
shm_sub BUGFIX skip process events when suspended

irfanHaslanded:
When a specific subscription in a subscription context is suspended, it may still contain other active subscriptions.

When the handler thread is woken up to process some active subscription(s), we check for all available events to process.

However, we must not process events for suspended subscriptions. If we do, we would be stealing the event from the intended subscribers.

It can also cause asserts to fired when more subs than subscriber_count have processed the event.

To prevent this, we should also store the suspended state of a subscription within the subscription context and check it before processing any events.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -2972,7 +2972,7 @@ sr_shmsub_notif_notify(sr_conn_ctx_t *conn, const struct lyd_node *notif, struct
     }
 
     /* notify all subscribers using event pipe */
-    for (i = 0; i < notif_sub_count; ) {
+    for (i = 0; i < notif_sub_count; i++) {
         if (ATOMIC_LOAD_RELAXED(notif_subs[i].suspended)) {
             /* skip suspended subscribers */
             continue;
@@ -2981,8 +2981,6 @@ sr_shmsub_notif_notify(sr_conn_ctx_t *conn, const struct lyd_node *notif, struct
         if ((err_info = sr_shmsub_notify_evpipe(notif_subs[i].evpipe_num))) {
             goto cleanup_ext_sub_unlock;
         }
-
-        ++i;
     }
 
     /* EXT READ UNLOCK */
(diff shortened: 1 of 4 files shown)


irfanHaslanded:
> Thanks, it looks fine and I assume you have tested it. But if you can, I would appreciate adding a new test for this which I think should not be that difficult to write if you avoid subscription threads and just force a suspended subscription to process events.

While that is true, having a subscription thread seemed more real-world and exposed another bug (infinite loop in notification sub).
So, I chose this approach.

src/shm_sub.c
```diff
@@ -2637,9 +2637,10 @@ sr_shmsub_rpc_notify(sr_conn_ctx_t *conn, sr_rwlock_t *sub_lock, off_t *subs, ui
         goto cleanup;
     }
 
-    if (!strcmp(path, SR_RPC_FACTORY_RESET_PATH) && (cur_priority == SR_RPC_FACTORY_RESET_INT_PRIO)) {
-        assert(subscriber_count == 1);
+    /* There can be only 1 RPC subscription at each prioirty level */
+    assert(subscriber_count == 1);
```
michalvasko, line 2641: Why did you change the `assert`? I would prefer for the code to remain unified and work correctly if we decide to support more subscriptions on RPCs for a single priority.

tests/test_multi_connection.c
```diff
@@ -190,12 +192,245 @@ test_new(void **state)
...
+
+        /* RPC/action subscriptions */
+        ret = sr_rpc_subscribe(st->sess1, rpc_xpath[i], suspend_rpc_cb, &pvt_data[i], 0, 0, &subscr[j]);
+        assert_int_equal(ret, SR_ERR_OK);
+        test_sub_suspend_helper(subscr[j], i);
+    }
+
+    /* all subscriptions are set up, wait for threads to be spawned, and exercise them all */
+    usleep(100000);
+    sched_yield();
```
michalvasko, line 359: Please do not use such constructions, they are never too reliable. Instead I usually create 2 threads and synchronize them using a barrier.

michalvasko (changes requested):

src/shm_sub.c
irfanHaslanded: Oh okay, I will remove this part of the code. I assumed it is never possible to have more than one subscription at the same priority for RPCs.

michalvasko: Your assumption is correct but I prefer for this code to be the same for all subscriptions, if we decide to support it later and simpler maintenance and bug fixing.

tests/test_multi_connection.c
irfanHaslanded: Yes, you are right. I had put it in before when I had some threads, but now this is unnecessary and can be removed.
The subscriptions are already in ext SHM and publishers will have to wait for them to be invoked anyway.

tests/test_multi_connection.c
```diff
@@ -190,12 +192,241 @@ test_new(void **state)
...
+        return;
+    }
+    sub_id = sr_subscription_get_last_sub_id(subscr);
+    ret = sr_subscription_suspend(subscr, sub_id);
+    assert_int_equal(ret, 0);
+}
+
+struct cb_data {
+    int fail;
+};
```
michalvasko, line 211: We usually use the `state` with its `cb_called` counter and check that on several places throughout the test as needed. But if you do not want to rewrite it, this is okay, too.

irfanHaslanded: I would like to leave it as it is, as this gives the `backtrace` when called unexpectedly, rather than just a counter, and makes it easier to debug a race for me.

michalvasko (approved):

michalvasko:
Thanks, merging.

PR #3303, opened 2024-05-23
srmon BUGFIX needs rpc lock and ext SHM lock

irfanHaslanded:
While reading the rpc sub ext SHM data, we need to acquire both rpc lock and ext SHM read lock. Without it, we risk reading a wrong memory location.

This is already done for change_subs, oper get subs and notif subs by a1710c897b9be6c54524530b0944effa88f5ec40

However, locks are still missing for RPC subs and added now.


michalvasko (approved):

michalvasko:
Right, missed these subscriptions, thanks.

PR #3318, opened 2024-06-11
shm_sub REFACTOR recover, new lock and notify wait

irfanHaslanded:
The logic in the following function has been revised multiple times, and some of the checks are no longer relevant.

This is mainly due to the use of `SR_SUB_EV_FINISHED` by subscribers and `SR_SUB_EV_NONE` by originators.

The following are relevant for synchronization:

1. sub_shm->orig_cid: This is used to track the originator of an event, and must always be set unless holding the mutex.
This is cleared only after originator has finished (success or failure) the event, and does not want to access the SHM anymore. It's purpose is entirely for sub_shm recovery (clear the event).
  a. if set, and orig_cid is dead
  b. if unset and event exists.

2. sub_shm->event: Originators set it to one of SR_IS_NOTIFY_EVENT and listeners set it to one of SR_IS_LISTEN_EVENT.
Only originators clear the event once they are done accessing the SHM or when they recover an abandoned event.

3. sub_shm->request_id: Identifies the event uniquely for the subscription. It is checked by subscribers to prevent processing the same event more than once. It is incremented by the originator to notify a new event only after acquiring `sr_shmsub_notify_new_wrlock`.

Special cases:

1. Notification subscription It is the only one which does not use a orig_cid while an event is ongoing.

2. Aborted events In case of aborted events, the originator does not clear the error status of the SHM, but clears the orig_cid.

3. Originator loses WRITE lock When originator loses WRITE lock, because some of the subscribers are holding READ locks, abort events are not sent. The event is cleared regardless of SUCCESS, FAILURE or FINISHED.
orig_cid is also cleared since it is no longer waiting for the event.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -309,18 +309,25 @@ sr_shmsub_recover(sr_sub_shm_t *sub_shm)
 {
     sr_sub_event_t ev = ATOMIC_LOAD_RELAXED(sub_shm->event);
 
-    if (sub_shm->orig_cid && !sr_conn_is_alive(sub_shm->orig_cid)) {
-        SR_LOG_WRN("EV ORIGIN: SHM event \"%s\" of CID %" PRIu32 " ID %" PRIu32 " recovered.",
-                sr_ev2str(ev), sub_shm->orig_cid,
-                (uint32_t)ATOMIC_LOAD_RELAXED(sub_shm->request_id));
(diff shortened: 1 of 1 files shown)


michalvasko (approved):

michalvasko:
Thanks, seems fine.

PR #3328, opened 2024-06-27
sysrepocfg log api errors

irfanHaslanded:
sysrepocfg should report back to users any errors contained in the session.

Without this, it is difficult to understand what may have gone wrong.


michalvasko (approved):

michalvasko:
Makes sense, thanks.

PR #3332, opened 2024-07-02
do not recover dead subs during notify

irfanHaslanded:
During event notification, we only have the ext SHM subscription locked
in READ mode. If we discover that a subscription is dead, we attempted
to recover the SHM by unlocking and acquiring a WRITE lock and
performing cleanup and then reverting to a READ lock.

This was error-prone especially because WRITE locks can become contented
and some return codes had to be ignored.
Further, it needed substantial amount of code to work.

It is better to ignore these dead subscriptions.

Before adding a new subscription of the same type to the same module, we check for any dead subscriptions and recover them. This is easy because we already hold write lock.


michalvasko (commented):
I can't say I like this too much. I agree the current code is not particularly good but this does not seem like an improvement.

irfanHaslanded:
> I can't say I like this too much. I agree the current code is not particularly good but this does not seem like an improvement.

When a subscriber dies, and subsequently a originator has failed to acquire WRITE locks, it can cause attempting to unlock locks which were not locked, triggering asserts in unlock code.

 Additionally, "normal" operation of the system is affected until the subscription is recovered due to delays involved.

So, while the approach is unconventional, I would think that this is an improvement in that we no longer unlock read -> write-lock -> downgrade to read lock, and still keep the SHM somewhat clean.

michalvasko:
One issue is that suspended subscriptions of dead connections would never be recovered.

But I was referring more to the code readability and complexity. I may try to come up with my own improvement to the situation later.

irfanHaslanded:
You are right about the dead suspended subscriptions never getting cleaned. I think I have fixed that now.

If there are specific portions of the changes that need to be refactored to be simpler/readable, please let me know.

I am open to any suggestions to attempt a different simpler/cleaner solution if you have ideas.

michalvasko:
I am not sure it is safe to change the `suspended` value without holding WRITE lock on the subscriptions (because you are removing a subscription, as far as originators are concerned). But, most likely, they can only timeout on some event (but it would be best to check that). Assuming this is fine, why not just normally suspend the subscriptions and then actually remove them when the WRITE lock is held. Yes, there would be some redundant subscription traversal but I strongly believe it is negligible (and can be optimized later, if it turns out it is a problem). This seems like a much cleaner solution with minimal additional code.

irfanHaslanded:
> I am not sure it is safe to change the `suspended` value without holding WRITE lock on the subscriptions (because you are removing a subscription, as far as originators are concerned). But, most likely, they can only timeout on some event (but it would be best to check that).

The subscription is already dead, so marking it as `suspended` is safe. Other originators beyond this point would see it as `suspended` instead of seeing it as `dead`. IMHO, no additional timeouts can happen.

> Assuming this is fine, why not just normally suspend the subscriptions and then actually remove them when the WRITE lock is held. Yes, there would be some redundant subscription traversal but I strongly believe it is negligible (and can be optimized later, if it turns out it is a problem).

You are right. The main advantage of marking as `scrapped` vs `suspended` is not having to search the entire subscription array every time we add a new subscription for dead subscriptions among suspended subs, in case a module has a large number of subscriptions. As long as there aren't a lot of suspended subscribers, this could be a negligible amount of workload.

Other reason was to have a distinction between suspended and scrapped subscriptions for debugging.

> This seems like a much cleaner solution with minimal additional code.

I will make changes to not track "scraps" in shm subscription segments and make it cleaner.

michalvasko:
I still see some `scraps` code in the PR, is it really finished?

irfanHaslanded:
I see 2 problems to get rid of tracking scraps completely:
1. We will need to do more expensive `sr_conn_is_alive` checks instead of a single atomic int comparison until full recovery.
2. Due to `sr_shmsub_oper_poll_get_sub_change_notify_evpipe`, we will end up waking up threads unnecessarily.

If these are acceptable, then we need not do anything during notification when a dead `CID` is discovered.

Before creating new subscriptions, we will then always check all existing subscriptions for `sr_conn_is_alive`, and perform recovery as needed.

michalvasko:
I see no problems with some inefficiencies. If you want to avoid them, do not let subscriptions die, simple as that.

src/shm_ext.h
```diff
@@ -272,18 +267,25 @@ sr_error_info_t *sr_shmext_rpc_sub_del(sr_conn_ctx_t *conn, off_t *subs, uint32_
...
  * @param[in] recovery Whether to print subscription recovery warning.
  * @return err_info, NULL on success.
  */
-sr_error_info_t *sr_shmext_rpc_sub_stop(sr_conn_ctx_t *conn, sr_rwlock_t *sub_lock, off_t *subs, uint32_t *sub_count,
-        const char *path, uint32_t del_idx, int del_evpipe, sr_lock_mode_t has_locks, int recovery);
+sr_error_info_t *sr_shmext_rpc_sub_stop(sr_conn_ctx_t *conn, off_t *subs, uint32_t *sub_count,
+        const char *path, uint32_t del_idx, int del_evpipe, int recovery);
+
+/**
+ * @brief Remove SCRAPPED main SHM module RPC/action subscription
```
michalvasko, line 282: s/SCRAPPED/dead/

src/sysrepo.c
```diff
@@ -5984,12 +5984,16 @@ _sr_rpc_subscribe(sr_session_ctx_t *session, const char *xpath, sr_rpc_cb callba
 
     /* add RPC/action subscription into ext SHM and create separate specific SHM segment */
     if (is_ext) {
-        if ((err_info = sr_shmext_rpc_sub_add(conn, &shm_mod->rpc_ext_lock, &shm_mod->rpc_ext_subs,
+        /* Remove any SCRAPPED subscriptions */
```
michalvasko, line 5987: s/SCRAPPED/dead/

src/sysrepo.c
```diff
@@ -5984,12 +5984,16 @@ _sr_rpc_subscribe(sr_session_ctx_t *session, const char *xpath, sr_rpc_cb callba
...
+        sr_shmext_rpc_sub_remove_dead(conn, &shm_mod->rpc_ext_subs, &shm_mod->rpc_ext_sub_count);
+        if ((err_info = sr_shmext_rpc_sub_add(conn, &shm_mod->rpc_ext_subs,
                 &shm_mod->rpc_ext_sub_count, path, sub_id, xpath, priority, 0, (*subscription)->evpipe_num, conn->cid))) {
             goto cleanup_unlock2;
         }
     } else {
-        if ((err_info = sr_shmext_rpc_sub_add(conn, &shm_rpc->lock, &shm_rpc->subs, &shm_rpc->sub_count, path, sub_id,
+        /* Remove any SCRAPPED subscriptions */
```
michalvasko, line 5994: s/SCRAPPED/dead/

michalvasko (changes requested):

src/shm_ext.c
```diff
@@ -498,6 +498,27 @@ sr_shmext_print(sr_mod_shm_t *mod_shm, sr_shm_t *shm_ext)
     free(items);
 }
 
+static void
+sr_shmext_change_sub_remove_dead(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, sr_datastore_t ds)
```
michalvasko, line 502: Please include doxygen for all static functions.

michalvasko (changes requested):
Sorry, have missed before, should be the last minor change needed. The tests are appreciated and complex but are also quite messy.

src/shm_ext.c
irfanHaslanded: Sorry, missed that. added now

michalvasko (approved):

PR #3333, opened 2024-07-02
shm_sub BUGFIX clear left over event and orig_cid

irfanHaslanded:
`orig_cid` should always be cleared when the originator stops waiting for the subscribers. If not, subsequent events cannot be generated on the subscription SHM, until the previous originator process is restarted.

Normally, the wait logic takes care of this. However, in case a subscription evpipe could not be notified we bail on waiting and return with error. However, in this case `orig_cid` and `event` are left behind never to be cleared.


michalvasko (approved):

michalvasko:
Fine, thanks.

PR #3353, opened 2024-07-15
sysrepocfg: add -s/--no-subs export argument

irfanHaslanded:
This sets the `SR_OPER_NO_SUBS` option when getting data from the operational datastore, thus retrieving only "push" data.

This is useful when pull operational data is not needed and expensive callbacks can be avoided.

Also when a component pull callback is failing
but a full dump of the operational store is still desired to assist with debugging.


michalvasko (approved):

michalvasko:
Fine, merging.

PR #3356, opened 2024-07-16
sysrepocfg: build-time select default LYD_FORMAT

irfanHaslanded:
It is useful to be able to choose the default file format for sysrepocfg at build time.

For instance, if most users prefer `JSON` over `XML`, it can be tedious to specify `-f json` for each `sysrepocfg` invocation.

With this change, it will be possible to build sysrepo using `-DSYSREPOCFG_DEFAULT_LYD_FORMAT=LYD_JSON` and use sysrepocfg.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,6 +228,11 @@ if(NOT SRPD_PLUGINS_PATH)
 endif()
 message(STATUS "SRPD plugins path:  ${SRPD_PLUGINS_PATH}")
 
+if(NOT SYSREPOCFG_DEFAULT_LYD_FORMAT)
+    set(SYSREPOCFG_DEFAULT_LYD_FORMAT LYD_XML)
+endif()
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -228,6 +228,11 @@ if(NOT SRPD_PLUGINS_PATH)
 endif()
 message(STATUS "SRPD plugins path:  ${SRPD_PLUGINS_PATH}")
 
+if(NOT SYSREPOCFG_DEFAULT_LYD_FORMAT)
+    set(SYSREPOCFG_DEFAULT_LYD_FORMAT LYD_XML)
+endif()
+message(STATUS "sysrepocfg default format: ${SYSREPOCFG_DEFAULT_LYD_FORMAT}")
```
michalvasko, line 234: Not sure if this message is necessary, but fine, if you just add `sysrepocfg default data format: ...`.

michalvasko (changes requested):

michalvasko (approved):

PR #3360, opened 2024-07-18
tests BUGFIX test_mult_update sync threads

irfanHaslanded:
One of the threads can starve the other unless we make sure that after each iteration, the other thread is also given a chance to run. Also, for the purposes of the test, it is preferable if both threads are running in parallel, instead of one thread being starved.

Without this, valgrind test_apply_changes sometimes fail, where only one thread runs for an extended period of time.

One example of such a failure is the run https://github.com/sysrepo/sysrepo/actions/runs/9971257900/job/27551955360

### DIFF
--- a/tests/test_apply_changes.c
+++ b/tests/test_apply_changes.c
@@ -7080,9 +7080,8 @@ test_write_starve(void **state)
 static void *
 apply_when1_thread(void *arg)
 {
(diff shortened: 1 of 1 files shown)


michalvasko (approved):
Thanks, I saw the test failed but had no spare time to analyze why.

PR #3365, opened 2024-07-24
Don't set group id if SR_ENV_RUN_TESTS is set

irfanHaslanded:
Various sysrepo initializing functions such as sr_open, sr_mkpath, call chown or chmod which the process using sysrepo may not always have the permissions to do.

For example, when a debian package is being built, we will need to run the unit-tests of the application using sysrepo as root or have permissions to change file ownership and permissions.

For the principle of least privilige, we shouldn't need to elevate the process running unit-tests.

In order to achieve this, we check for an environment variable called `SR_ENV_RUN_TESTS`. If this is set, sysrepo will not call chmod, chown system calls. In addition to SR_ENV_RUN_TESTS, SR_SHM_PREFIX_ENV is also expected to be set to a location where the process has permissions to read and write, such as a temporary dir in /tmp.


src/common.c
```diff
@@ -3498,13 +3498,16 @@ sr_mkpath(char *path, mode_t mode)
     int r;
     gid_t gid;
 
-    assert(path[0] == '/');
+    if (sr_is_prod_env()) {
+        /* Test environments may use relative paths */
+        assert(path[0] == '/');
+    }
```
michalvasko, line 3504: Seems that relative paths will work fine so this assert can simply be removed.

src/plugins/common_json.c
```diff
@@ -528,13 +528,16 @@ srpjson_mkpath(const char *plg_name, char *path, mode_t mode)
     const char *group = SR_GROUP;
     gid_t gid;
 
-    assert(path[0] == '/');
+    if (sr_is_prod_env()) {
+        /* Test environments may use relative paths */
+        assert(path[0] == '/');
+    }
```
michalvasko, line 534: Again, can be removed.

michalvasko (changes requested):
Seems the solution is fairly clean so I have nothing against it. But I would expect this variable then to be set for the tests and when building the package. It should also probably be documented (mentioned) somewhere. Lastly, there are some formatting errors.

jktjkt:
> For example, when a debian package is being built, we will need to run the unit-tests of the application using sysrepo as root or have permissions to change file ownership and permissions.

Unit tests are already run with env vars which "redirect" the repository location into a build directory (and use prefixes for SHM paths). This is a prerequisite for parallel testing, btw, so I'm surprised that you're hitting a problem. Is there any particular failure that you're getting during the package/CI/whatever builds?

> For the principle of least privilige, we shouldn't need to elevate the process running unit-tests.

On the other hand, test code should trigger code paths which are as close to production deployments as possible. This patch changes that approach quite substantively.

irfanHaslanded:
@jktjkt The unit-tests you describe run without `-DSYSREPO_GROUP=sysrepo` and as a result probably contain `build/config.h:#define SR_GROUP ""`.
As a result `chown` is never called due to `strlen(SR_GROUP) && (chown(path, -1, gid) == -1)`.

However this is not always the case, for instance
`distro/pkg/deb/rules` defines `-DSYSREPO_GROUP=sysrepo` and if we wanted to run unit-tests as part of the debian package build, we would see something like

```c
41: 715.695335 [367841][367841][ERR]: chown() failed (Operation not permitted).
41: 715.698912 [367841][367841][ERR][TestLog]: "sr_connect(0, &conn) == SR_ERR_OK" not true
#0  _test_log_msg (level=SR_LL_ERR, message=0x5555555755a0 "chown() failed (Operation not permitted).", prefix=0x55555556b8c0 "") at ./tests/tcommon.c:27
#1  0x00007ffff7f372d9 in sr_errinfo_new (err_info=err_info@entry=0x7fffffffd9d0, err_code=err_code@entry=SR_ERR_SYS,
    msg_format=msg_format@entry=0x7ffff7f79d5c "%s() failed (%s).") at ./src/log.c:255
#2  0x00007ffff7f2a1bb in sr_mkpath (path=0x555555574d20 "/build/sysrepo-2Fkr4k/sysrepo-2.3.10/obj-x86_64-linux-gnu/test_apply_changes", mode=mode@entry=511)
    at ./src/common.c:3863
#3  0x00007ffff7f4d4eb in sr_shmmain_check_dirs () at ./src/shm_main.c:88
#4  0x00007ffff7f14d7c in sr_connect (opts=opts@entry=0, conn_p=conn_p@entry=0x555555574c90) at ./src/sysrepo.c:224
#5  0x0000555555557abf in setup (state=0x7fffffffdc48) at ./tests/test_apply_changes.c:66
#6  0x00007ffff7fb79f3 in ?? () from /usr/lib/x86_64-linux-gnu/libcmocka.so.0
#7  0x00007ffff7fb7e0c in _cmocka_run_group_tests () from /usr/lib/x86_64-linux-gnu/libcmocka.so.0
#8  0x000055555555754b in main () at ./tests/test_apply_changes.c:7295
```

without this patch.

irfanHaslanded:
>On the other hand, test code should trigger code paths which are as close to production deployments as possible.

Yes, I agree. As explained above, current test code skips `chown`, as will code with this patch. The only difference is, currently tests cannot be run during package build time, unless builds are run as root. With this patch, it should be possible.

michalvasko:
So what about the package scripts, do you want to update those? Or are you using your own custom scripts that set this env variable?

irfanHaslanded:
> So what about the package scripts, do you want to update those? Or are you using your own custom scripts that set this env variable?

I have updated the debian packaging under `distro/pkg/deb/` although I am not sure what needs to be done for rpm.

CMakeLists.txt
```diff
@@ -31,6 +31,9 @@ string(TOUPPER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_UPPER)
 if("${BUILD_TYPE_UPPER}" STREQUAL "RELEASE")
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build Type" FORCE)
     set(CMAKE_C_FLAGS "-DNDEBUG -O2 ${CMAKE_C_FLAGS}")
+elseif("${BUILD_TYPE_UPPER}" STREQUAL "PACKAGE")
+    set(CMAKE_BUILD_TYPE "Package" CACHE STRING "Build Type" FORCE)
+    set(CMAKE_C_FLAGS "-DNDEBUG -g -O2 ${CMAKE_C_FLAGS}")
```
michalvasko, line 36: You can put these flags for `RELWITHDEBINFO` target, should probably have been there. Then use this build type, no need to add another.

michalvasko (changes requested):

michalvasko:
Thanks, that is fine and it would be a good idea to build the package in the CI just like for *libyang* but I can add that later.

michalvasko (approved):

michalvasko:
I had to make some changes (mainly 9d7220f0) for the packages to build, feel free to update them in a PR, the CI will now check them for you.

PR #3371, opened 2024-07-29
test headers bugfix check_includes.sh

irfanHaslanded:
The script must be executed in bash and not sh as it uses bash brace expansion.

Also cleaning up the script.


michalvasko (approved):

michalvasko:
Thanks, merged.

PR #3372, opened 2024-07-29
ly_wrap: discard ignored libyang errors

deastoe:
If specific libyang errors are ignored, discard them. Otherwise they are effectively leaked until/unless a unignored error is encountered.


michalvasko (changes requested):
Thanks, I have nothing against the change, but I would say you are too defensive, these functions are only called internally by *sysrepo* and it should never call them with invalid arguments (meaning if `ctx` is required, it will always be provided). Also, there should be no previous errors, whatever are generated need to be processed right after the *libyang* function fails, so you can ignore those (and they are those "leaks" you are talking about).

michalvasko (approved):

deastoe:
Yes, good point, I've simplified the diffs.

PR #3383, opened 2024-08-06
subs - lower periodic logs from info to debug level

irfanHaslanded:
various sub_shm functions produce logs for "EV ORIGIN" and "EV LISTEN" and most of these are of a periodic nature.
For long running systems this can pose a problem by filling up the system journal logs and cause useful data be overwritten.

Informational level logs are usually understood to be for reporting the state of the system rather than reporting on every event execution in the normal running of the system.

Debug level is more suitable for this purpose which can then be filtered by applications which perform actions frequently.


michalvasko (approved):

michalvasko:
Alright, no problem.

PR #3396, opened 2024-08-16
Simplify GNU C / LLVM clang check

plajjan:
The CMAKE_C_COMPILER_ID won't work out of the box when this compat.h.in is translated via Zigs ConfigHeader, whereas this simpler style does and as I believe they are equivalent / have the same purpose, I opted for the simpler variant.

### DIFF
--- a/compat/compat.h.in
+++ b/compat/compat.h.in
@@ -46,7 +46,7 @@
 #  endif
 #endif
 
-#if (@CMAKE_C_COMPILER_ID@ == GNU) || (@CMAKE_C_COMPILER_ID@ == Clang)
+#if defined (__GNUC__) || defined (__llvm__)
 # define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))
 # define _PACKED __attribute__((__packed__))
 #else


michalvasko (approved):

michalvasko:
Okay, no problem, except we accept PRs only into the `devel` branch.

jktjkt:
I do not know what Zig is, but note that this header file is *never* used directly. Instead, CMake processes its content (which involves [replacing some strings](https://cmake.org/cmake/help/latest/command/configure_file.html#transformations)) and produces a *generated* `.h` file that's in turn used by the C compiler during the build. Why are you pointing this Zig thing on a file that's supposed to be processed by CMake first?

plajjan:
@jktjkt Zig is a programming language and toolchain, see https://ziglang.org

As far as the language itself goes, it's C-level but more modern, less footguns etc and is ABI compatible with C, so complete interop. Briefly, Zig aims to be a better systems level language than C (and in a different style than C++ and Rust). Part of that is also being better at building software, replacing things like make, cmake, ninja, or autotools. To this end, Zig has a build system. Among other features, it can read and generate config headers in the style of cmake or autotools (or completely manual). While it can parse and work with the template itself, it doesn't generate the variable values and so variables populated by cmake is a small hurdle.

In this case, the simpler option is IMHO both easier to read and has less reliance on CMake itself which is hardly a bad thing.

I think Zigs build system is really quite nice. It has amazing caching, it builds concurrently per default. Nice DAG. Unlike CMake, which is some script style syntax, the Zig build system is a Zig program in itself - you write the build "script" (program!) in the Zig language. I think it's nice to have a proper programming language for this. It has fantastic cross-compilation support and all around does so many things so well.

Zig also has a package manager, so the build system can rely on dependencies on the Internet. Here's an example of how I use zig to build sysrepo.

2024-08-19-224758_1619x947_scrot

The build.zig configs aren't complete here, otherwise we could easily do cross-compilation to macos aarch64 (apple M1 silicon) or whatnot. Really cool stuff IMHO :)

PR #3424, opened 2024-10-01
Allow runtime configurable shm dir

irfanHaslanded:
using `SYSREPO_SHM_DIR` env variable, it is now possible to relocate where sysrepo stores its shared memory data.

This will allow processes without access to default system locations such as `/dev/shm` to run unit-tests, without needing to rebuild sysrepo.


michalvasko (changes requested):
You have decided to use a static variable to cache the directory meaning it cannot be changed during the lifetime of a process. Not sure if that can cause any problems but similar restriction is not used for any other env directories so I would prefer you do not use it even in this case.

irfanHaslanded:
> You have decided to use a static variable to cache the directory meaning it cannot be changed during the lifetime of a process. Not sure if that can cause any problems but similar restriction is not used for any other env directories so I would prefer you do not use it even in this case.

I was thinking that since changing the vars during the lifetime of a process will most certainly cause problems as we don't handle any "migration" on discovering that the value is now changed, it should be static.
IMHO, the other places should also be using static?

I can make this as the rest of the code if you still prefer it to be read each time.

michalvasko:
Yeah, you are right that it would cause issues. So fine, you can modify the other evn directories this way. Just a note regarding formatting, please leave an empty line before `return` (not sure why `uncrustify` did not catch this).

irfanHaslanded:
> Yeah, you are right that it would cause issues. So fine, you can modify the other evn directories this way.

```C
       As typically implemented, getenv() returns a pointer to a string within the environment list.  The caller must take care not to modify this  string,  since  that  would
       change the environment of the process.

       The  implementation of getenv() is not required to be reentrant.  The string pointed to by the return value of getenv() may be statically allocated, and can be modified
       by a subsequent call to getenv(), putenv(3), setenv(3), or unsetenv(3).
```

My assumption was wrong. Looks like using `static` won't ensure the static pointer value is what was first read. So, the applications should simply never change these ENV variables in the lifetime of the process, as `getenv` is not even reentrant.

So, the only benefit of using a `static char *` would be to cache the address of the value and avoid repeated look ups.

What do you think should be done?

michalvasko:
> My assumption was wrong. Looks like using static won't ensure the static pointer value is what was first read.

Ah, you are right, I have forgotten about that as well. So you have to make a copy and you could use a fixed-sized string (but use `PATH_MAX`, which should be 256). That seems to me is the best solution. Maybe also add a note into README with this env vars that they cannot be changed during the lifetime of a process and if they are, the changes will be ignored.

src/sysrepo.c
```diff
@@ -1220,14 +1220,21 @@ sr_session_get_connection(sr_session_ctx_t *session)
 API const char *
 sr_get_repo_path(void)
 {
+    static char sr_repo_path[256] = "";
```
michalvasko, line 1223: `SR_PATH_MAX` not used.

src/plugins/common_json.c
```diff
@@ -124,18 +124,27 @@ srpjson_file_exists(const char *plg_name, const char *path)
...
+        *prefix = sr_shm_prefix_val;
         return err_info;
     }
 
-    return NULL;
+    /* first time */
+    tmp = getenv(SR_SHM_PREFIX_ENV);
+    if (tmp == NULL) {
+        tmp = SR_SHM_PREFIX_DEFAULT;
+    } else if (strlen(tmp) >= SR_PATH_MAX) {
```
michalvasko, line 140: I would like such a check for all the other ENV paths. Silently not using the value is really difficult to debug.

michalvasko (changes requested):

michalvasko (approved):

michalvasko:
Fine now, thanks.

PR #3431, opened 2024-10-14
reproducible builds and bugfix strncpy length

irfanHaslanded:
- **debian rules set CMAKE_BUILD_WITH_INSTALL_RPATH ON**
    Make the build reproducible by fixing the remapping compiled
    references to __FILE__ and by avoiding RPATH insertions.

- **tests - set LD_LIBRARY_PATH to locate libsysrepo7.so**
    With rpath enabled in cmake, the unit-tests may not discover where the
    shared object of libsysrepo is located after the build.

- **bugfix strncpy null term of maxlen SR_PATH_MAX**
    `strncpy` does not result in a destination string but a character
    sequence. If 'dst' is not long enough to contain the null byte,
    'dst' will not be null-terminated.

    `stlcpy` does this but is not portable.

    `strncpy` expects us to pass the size of the destination buffer as the
    limiting size for copying.

    It also expects that the source string fits into the destination
    buffer including the null-termination byte, if the result is to be a
    string and not a character sequence.

    The strlen() function calculates the length of the string pointed
    to by s, excluding the terminating null byte ('\0').

    So, the correct way to use `strncpy()` is:
```c
    char dst[DST_SZ];
    char src[SRC_SZ];
    ...
    if (strnlen(src, SRC_SZ) < DST_SZ) {
        strncpy(dst, src, DST_SZ);
    }
```

    See `String vs character sequence` in manpage of `string_copying`.
    strlcpy or strscpy maybe alternatives, but with other challenges such as
    portability or availability.

- **bugfix Wunused-result in test_process.c**

### DIFF
--- a/distro/pkg/deb/rules
+++ b/distro/pkg/deb/rules
@@ -7,16 +7,20 @@ include /usr/share/dpkg/default.mk
 # determine number of processors to enable parallel tests
 NPROCS:=$(shell grep -c ^processor /proc/cpuinfo)
 
+# Use the build directory to set LD_LIBRARY_PATH for tests
+SR_BUILD_DIR:=$(shell pwd)/obj-$(DEB_HOST_GNU_TYPE)
+
 %:
 	dh $@ -j
 
 override_dh_auto_configure:
 	dh_auto_configure -- \
 		-DCMAKE_BUILD_TYPE:String="RelWithDebInfo" \
+		-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
 		-DSYSREPO_UMASK=007 \
 		-DSYSREPO_GROUP=sysrepo \
 		-DNACM_SRMON_DATA_PERM=660
 
 override_dh_auto_test:
-	make -C obj-$(DEB_HOST_GNU_TYPE) test ARGS=' -j $(NPROCS) -V'
+	LD_LIBRARY_PATH="$(SR_BUILD_DIR)" make -C obj-$(DEB_HOST_GNU_TYPE) test ARGS=' -j $(NPROCS) -V'
 	make -C obj-$(DEB_HOST_GNU_TYPE) test_clean
--- a/src/common.c
+++ b/src/common.c
@@ -215,7 +215,7 @@ sr_ds_handle_init(struct sr_ds_handle_s **ds_handles, uint32_t *ds_handle_count)
         }
 
         if ((len = strlen(tmp)) < SR_PATH_MAX) {
-            strncpy(plugins_dir, tmp, SR_PATH_MAX);
+            snprintf(plugins_dir, SR_PATH_MAX, "%s", tmp);
         } else {
             sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, "SR_PLUGINS_PATH (%s) cannot be longer than %u.", tmp, SR_PATH_MAX);
             goto cleanup;
@@ -398,7 +398,7 @@ sr_ntf_handle_init(struct sr_ntf_handle_s **ntf_handles, uint32_t *ntf_handle_co
         }
 
         if ((len = strlen(tmp)) < SR_PATH_MAX) {
-            strncpy(plugins_dir, tmp, SR_PATH_MAX);
+            snprintf(plugins_dir, SR_PATH_MAX, "%s", tmp);
         } else {
             sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, "SR_PLUGINS_PATH (%s) cannot be longer than %u.", tmp, SR_PATH_MAX);
             goto cleanup;
@@ -1070,7 +1070,7 @@ sr_shm_dir_get(void)
         }
     }
 
-    strncpy(sr_shm_dir_str, tmp, SR_PATH_MAX);
+    snprintf(sr_shm_dir_str, SR_PATH_MAX, "%s", tmp);
 
     return sr_shm_dir_str;
 }
@@ -1108,7 +1108,7 @@ sr_shm_prefix(const char **prefix)
     if (err_info) {
         *prefix = NULL;
     } else {
-        strncpy(sr_shm_prefix_val, tmp, SR_PATH_MAX);
+        snprintf(sr_shm_prefix_val, SR_PATH_MAX, "%s", tmp);
         *prefix = sr_shm_prefix_val;
     }
     return err_info;
--- a/src/common.c
+++ b/src/common.c
@@ -210,15 +210,15 @@ sr_ds_handle_init(struct sr_ds_handle_s **ds_handles, uint32_t *ds_handle_count)
     if (!plugins_dir[0]) {
         /* get plugins dir from environment variable, or use default one */
         tmp = getenv("SR_PLUGINS_PATH");
-        if (tmp) {
-            if ((len = strlen(tmp)) < SR_PATH_MAX) {
-                strncpy(plugins_dir, tmp, len);
-            } else {
-                sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, "SR_PLUGINS_PATH (%s) cannot be longer than %u.", tmp, SR_PATH_MAX);
-                goto cleanup;
-            }
+        if (!tmp) {
+            tmp = SR_PLG_PATH;
+        }
+
+        if ((len = strlen(tmp)) < SR_PATH_MAX) {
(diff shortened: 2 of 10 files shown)


src/common.c
```diff
@@ -212,13 +212,13 @@ sr_ds_handle_init(struct sr_ds_handle_s **ds_handles, uint32_t *ds_handle_count)
...
             if ((len = strlen(tmp)) < SR_PATH_MAX) {
-                strncpy(plugins_dir, tmp, len);
+                strncpy(plugins_dir, tmp, SR_PATH_MAX);
             } else {
                 sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, "SR_PLUGINS_PATH (%s) cannot be longer than %u.", tmp, SR_PATH_MAX);
                 goto cleanup;
             }
         } else {
-            strncpy(plugins_dir, SR_PLG_PATH, SR_PATH_MAX - 1);
+            strncpy(plugins_dir, SR_PLG_PATH, SR_PATH_MAX);
```
michalvasko, line 221: The reason for adding `-1` is that if the string is `SR_PATH_MAX` chars long, it would create a non-terminated string (detected by `coverity`). It cannot occur in the first case since the string length is checked, but it can in this one, in theory.

distro/pkg/deb/rules
```diff
@@ -13,6 +13,7 @@ NPROCS:=$(shell grep -c ^processor /proc/cpuinfo)
 override_dh_auto_configure:
 	dh_auto_configure -- \
 		-DCMAKE_BUILD_TYPE:String="RelWithDebInfo" \
+		-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
```
michalvasko, line 16: Could you please add some details to justify this change? Having to customize `LD_LIBRARY_PATH` for the tests is not ideal.

src/common.c
irfanHaslanded: You are right. However, that would lead to silently truncating the value, which may lead to unpredictable behavior and may not be easy to debug. So, I have added a length check for the default value as well.

distro/pkg/deb/rules
irfanHaslanded: We want to make the builds reproducible. 
For reference:
https://wiki.debian.org/ReproducibleBuilds/Howto

https://reproducible-builds.org/docs/deterministic-build-systems/

> **CMake notes**
The default configuration of CMake makes the build directory part of the build environment. Here are some known issues and recommendations:
CMake sets a RPATH for binaries that link to a library in the the same project. Even when this is stripped at installation time, the build-id section will be different. Possible workarounds:
Users can set [CMAKE_SKIP_RPATH=ON](https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_RPATH.html) to disable the use of RPATH. Disadvantage: programs from the build directory cannot be run without setting LD_LIBRARY_PATH.
Projects can set [CMAKE_BUILD_WITH_INSTALL_RPATH=ON](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_WITH_INSTALL_RPATH.html) to ensure a deterministic RPATH. Disadvantage: programs from the build directory cannot be run without setting LD_LIBRARY_PATH.

src/common.c
```diff
@@ -210,15 +210,15 @@ sr_ds_handle_init(struct sr_ds_handle_s **ds_handles, uint32_t *ds_handle_count)
...
-            } else {
-                sr_errinfo_new(&err_info, SR_ERR_INVAL_ARG, "SR_PLUGINS_PATH (%s) cannot be longer than %u.", tmp, SR_PATH_MAX);
-                goto cleanup;
-            }
+        if (!tmp) {
+            tmp = SR_PLG_PATH;
+        }
+
+        if ((len = strlen(tmp)) < SR_PATH_MAX) {
+            strncpy(plugins_dir, tmp, SR_PATH_MAX);
```
michalvasko, line 218: Why do you insist on using `SR_PATH_MAX` for the length when such a length is invalid? I insist on using a valid length instead, otherwise this line is confusing for no reason at all. Please use either `len` or `SR_PATH_MAX - 1` is fine as well, where `len` is not used.

distro/pkg/deb/rules
michalvasko: Fine, if you think there is any advantage to this. But it should be possible to set that `LD_LIBRARY_PATH` in this script, no? My guess would be that you can just set it in `override_dh_auto_test` when running the tests. But if not, there must be some other suitable package build step.

michalvasko (changes requested):

src/common.c
irfanHaslanded: I was trying to address a warning you can see here [build-log-link](https://productionresultssa5.blob.core.windows.net/actions-results/79812071-a81b-4be6-9587-95fb984355c3/workflow-job-run-dfb3d9d1-7d78-56db-b388-e25a8a802d50/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-10-15T08%3A27%3A45Z&sig=%2FOyBzxeGFCRZfMIO7buTXuOnAnAKI5%2BKGD4rFch42IM%3D&ske=2024-10-15T17%3A37%3A54Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-10-15T05%3A37%3A54Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-10-15T08%3A17%3A40Z&sv=2024-08-04).

```C
2024-10-11T13:40:16.5830026Z In function ‘strncpy’,
2024-10-11T13:40:16.5831926Z     inlined from ‘sr_ds_handle_init’ at /home/runner/work/sysrepo/sysrepo/pkg/build/pkgs/ubuntu-22.04/sysrepo_2.11.10-1/sysrepo-2.11.10/src/common.c:215:17:
2024-10-11T13:40:16.5835253Z /usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:10: warning: ‘__strncpy_chk’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
2024-10-11T13:40:16.5837328Z    95 |   return __builtin___strncpy_chk (__dest, __src, __len,
2024-10-11T13:40:16.5838277Z       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2024-10-11T13:40:16.5839258Z    96 |                                   __glibc_objsize (__dest));
2024-10-11T13:40:16.5840023Z       |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
2024-10-11T13:40:16.5841948Z /home/runner/work/sysrepo/sysrepo/pkg/build/pkgs/ubuntu-22.04/sysrepo_2.11.10-1/sysrepo-2.11.10/src/common.c: In function ‘sr_ds_handle_init’:
2024-10-11T13:40:16.5844951Z /home/runner/work/sysrepo/sysrepo/pkg/build/pkgs/ubuntu-22.04/sysrepo_2.11.10-1/sysrepo-2.11.10/src/common.c:214:24: note: length computed here
2024-10-11T13:40:16.5846622Z   214 |             if ((len = strlen(tmp)) < SR_PATH_MAX) {
```

Using `len` will not copy the null termination character from `tmp`.
When using `SR_PATH_MAX-1`, if `strlen(tmp) == SR_PATH_MAX-1`, then again `strncpy` will not copy the null termination character.

However in this case, since `plugins_dir` is `static`, the last byte is already 0 and does not need to be copied, So I guess it really is a false-positive.

If it is confusing to use `SR_PATH_MAX`, I could use `len+1`, would that be better?

michalvasko: Fair enough, such a trivial task and so difficult to write it correctly. What about just using `strcpy()`, do you think the compiler/static analysis would mind?

irfanHaslanded: `strcpy` also causes problems for analyzers such as 
> This 'call to strcpy' with input from may overflow the destination. CodeQL
The program performs a buffer copy or write operation with no upper limit on the size of the copy. An unexpectedly long input that reaches this code will cause the buffer to overflow. In addition to causing program instability, techniques exist which may allow an attacker to use this vulnerability to execute arbitrary code.

michalvasko: Yeah, but I would expect them to be able to figure out the length is checked before. Whatever, I would then prefer using `snprintf()` to `strncpy()` because you can use `SR_PATH_MAX` as the destination buffer length.

michalvasko (approved):

PR #3433, opened 2024-10-15
more descriptive debugs for no datastore changes

irfanHaslanded:
When there are no datastore changes in an sr_apply_changes call, it is useful to know what caused the changes to vanish.

For example, was it validation, or an update from a subscriber.

Also reduce it to debug as it can be noisy at INFO level.


src/sysrepo.c
```diff
@@ -4110,6 +4110,7 @@ sr_apply_changes(sr_session_ctx_t *session, uint32_t timeout_ms)
         sr_errinfo_merge(&err_info, cb_err_info);
         sr_errinfo_new(&err_info, SR_ERR_CALLBACK_FAILED, "User callback failed.");
     }
+    SR_LOG_DBG("%s for %s %s.", __func__, sr_ds2str(session->ds), err_info ? "failed" : "success");
```
michalvasko, line 4113: Just to keep it unified and not use any function names, please make the message something akin to
```
Applying "<DS>" datastore changes failed/suceeded.
```

michalvasko (changes requested):
No problem, I requested just a minor change.

michalvasko (approved):

PR #3434, opened 2024-10-15
Add PTHREAD_MUTEX_ERRORCHECK attr

irfanHaslanded:
from the manpage of pthread_mutexattr_settype

This type of mutex provides error checking.
A thread attempting to relock this mutex without first unlocking it shall return with an error. A thread attempting to unlock a mutex which another thread has locked shall return with an error. A thread attempting to unlock an unlocked mutex shall return with an error.

It is good to have pthread mutex error checking enabled. It may have a performance impact, but it is not enabled in release builds.


michalvasko (approved):

michalvasko:
Fine even though I have never needed to set this attribute and manage to debug all the problems without it.

PR #3435, opened 2024-10-15
test_apply_changes increase timeout to 100ms

irfanHaslanded:
10ms is too low for the test to correctly work sometimes when the CPU
is too busy on the CI pipeline runner, causing intermittent failures.


michalvasko:
You motivated me to get rid of all those loops with active waiting. Please rebase and add whatever changes you think are still necessary.

michalvasko (approved):

PR #3445, opened 2024-10-30
debugs sr_rwlock_t tracing

irfanHaslanded:
It is nice to have an ID associated with each sr_rwlock_t to make debugging complex locking issues easier, and for visualising the lock order.

It is also useful to have traceability to account for which locks are in contention.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -61,6 +61,45 @@
 #include "shm_sub.h"
 #include "sysrepo.h"
 
+#define SR_RWLOCK_ID(rwlock_ptr) (~(1u<<31) & (rwlock_ptr)->id)
+#define _SR_LOG_RWLOCK(rwlock_ptr, mode, cid, func, str)                   \
+    SR_LOG_DBG("%s: %s rwlock %" PRIu32 " in mode %d by CID %" PRIu32, \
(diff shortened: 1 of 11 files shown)


michalvasko (changes requested):
My first thought going through this is that it seems like an ad-hoc solution, to help with debugging locks. But, I agree with the idea of having separate categories of debugging messages that could be enabled independently (currently, debug output is quite unusable because there are lots of SHM messages, which slow down all the operations).

However, I would prefer to have it implemented similarly to *libyang* or even *netopeer2*, have an API function that would allow enabling only selected debug categories avoiding environment variables.

Support for a log file is also basic and it should be properly integrated into the existing mechanisms of `stderr` and `syslog` logging.

Finally, the rwlock identification and logging logic seems quite complex and a bit messy, so a major refactoring is preferred. Also, I try to never add too much data/code that is used only for debugging and ideally find the simplest mechanism that includes all the required information (despite not being immediately obvious) but does not add complexity to the project itself.

So, I would start by making a separate first commit of introducing debug categories with the only category `SHM` and then build on that.

irfanHaslanded:
I will create separate pull-requests to make incremental changes.

PR #3450, opened 2024-11-12
shm_sub ensure recovery on previous incomplete cleanup

irfanHaslanded:
A process holding a subscription could die for any number of reasons and not always have the opportunity to clean up completely.

Sometimes, we may end up with a incompletely cleaned subscription/SHM where the shm file is deleted but the subscription is not fully removed yet.

In this case, we should ensure the subscription SHM is recovered before adding a new subscription.

However, by checking the return value of `unlink` and not making allowance for the shm file to be missing, we may be forever stuck, being unable to add a new subscription.

For example, the below failure occurs when the process is restarted:

```
[app::sysrepo] Recovering module "ietf-interfaces" operational get subscription of CID 11.
[app::sysrepo] Failed to unlink "/dev/shm/srsub_ietf-interfaces.oper.f05e29e6" SHM (No such file or directory).
[app::session] sr_oper_get_subscribe() failed: 3
```

So, we should make an exception for `errno == ENOENT` while checking unlink return code.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -187,8 +187,8 @@ sr_shmsub_unlink(const char *name, const char *suffix1, int64_t suffix2)
         goto cleanup;
     }
 
-    /* unlink */
-    if (unlink(path) == -1) {
+    /* unlink and ignore non-fatal missing file as it may have been removed earlier or never created (on a crash) */
+    if ((unlink(path) == -1) && (errno != ENOENT)) {
         sr_errinfo_new(&err_info, SR_ERR_SYS, "Failed to unlink \"%s\" SHM (%s).", path, strerror(errno));
         goto cleanup;
     }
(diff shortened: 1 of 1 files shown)


michalvasko:
It seems this really can happen but only for [`oper pull` subscriptions](https://github.com/sysrepo/sysrepo/blob/master/src/shm_ext.c#L952), the other ones are removed from SHM first and only them the files are deleted. Wouldn't it make more sense to change these subscriptions similarly so that this can never occur (file being removed before the subscription)?

src/shm_sub.c
```diff
@@ -189,8 +189,11 @@ sr_shmsub_unlink(const char *name, const char *suffix1, int64_t suffix2)
 
     /* unlink */
     if (unlink(path) == -1) {
-        sr_errinfo_new(&err_info, SR_ERR_SYS, "Failed to unlink \"%s\" SHM (%s).", path, strerror(errno));
-        goto cleanup;
+        if (errno != ENOENT) {
```
michalvasko, line 192: Please put the additional condition together with the current one, no need for a nested indent. Also, include the reason for the error being ignored.

michalvasko (changes requested):

src/shm_sub.c
```diff
@@ -187,8 +187,8 @@ sr_shmsub_unlink(const char *name, const char *suffix1, int64_t suffix2)
         goto cleanup;
     }
 
-    /* unlink */
-    if (unlink(path) == -1) {
+    /* unlink and ignore non-fatal missing file as it may have been removed earlier or never created */
```
michalvasko, line 190: Please also add the reason for that, just `(on a crash)` at the end, for example.

michalvasko (changes requested):

michalvasko (approved):

PR #3458, opened 2024-11-26
Add sr_xpath_parent to retrieve the parent path

nbiskup-eridan:
Added a new function sr_xpath_parent to allow the retrieval of the parent path from an xpath expression.

### DIFF
--- a/src/utils/xpath.c
+++ b/src/utils/xpath.c
@@ -582,6 +582,52 @@ sr_xpath_node_name_eq(const char *xpath, const char *node_name)
     }
 }
 
+API char *
+sr_xpath_parent(const char *xpath)
+{
+    const char *last_slash = NULL, *quot = NULL;
+    size_t parent_length = 0;
+    char *parent_xpath = NULL;
+
+    if ((NULL == xpath) || ('\0' == *xpath)) {
+        return NULL;
+    }
+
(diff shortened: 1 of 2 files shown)


michalvasko:
Is it really necessary, to add the function into *sysrepo*? Why don't you keep it only in your project?

nbiskup-eridan:
> Is it really necessary, to add the function into _sysrepo_? Why don't you keep it only in your project?

I believe these functions should be part of sysrepo or libyang, as these libraries are designed to work with YANG models and xpath expressions. Additionally, such functionalities are already implemented within the libraries but are not publicly accessible. 

I understand that these libraries are not intended to be xpath handlers or generators, so if you feel this functionality doesn't belong here, we can close this PR, and I will keep it as a patch in my project.

michalvasko:
I understand but you said it, the XPath parsing functionality is mostly kept internal. The reason being that there are lots of possible use-cases (yours is just one) that would result in a host of new public functions. Also, they would have t be generic enough, which may result in poor performance (as seen with *sysrepo* xpath functions). For these reasons, we have opted not to include them and let every project write their own because they are fairly simple to write and tailored to the specific use-case, they can be efficient. So thanks for the effort but I am closing this PR.

PR #3459, opened 2024-11-27
reuse module sub_shm diff for multiple events

irfanHaslanded:
Each commit can have multiple events published to subscribers:
1. UPDATE
2. CHANGE
3. DONE
4. ABORT (in case of failure)

These subscribers can also have different priorities.

A diff is shared with each subscriber as part of the notification
process. This diff maybe updated during the UPDATE phase, but is not
changed after that. (ABORT events use a reversed diff).

The shared diff is prepared by printing the libyang data tree for the
module in `LYD_LYB` format for each subscription priority, and for each
event including CHANGE and DONE events, first into memory, and then by
copying it into the shared memory (sub_data_shm).
At scale, this diff preparation can be expensive.

Since this diff does not change during CHANGE and DONE events, it can be
reused to save significant amount of time.

Since multiple locks are held during `sr_changes_notify_store` or
`sr_shmsub_oper_poll_listen_process_module_events`, the time saved is
even more valuable.

A new per-module flag `reuse_diff` is added to the mod_info structure to
allow reusing the diff for multiple notifications. This flag is set
after the diff is prepared by the first CHANGE event for a module and
is reused for all subsequent CHANGE and DONE events which will be
published afterwards.

Since ABORT events use a reversed diff, the `reuse_diff` flag is cleared
for all modules before looking up subscriptions to notify ABORT events.
The reversed diff once written in, can also be reused for all ABORT
events of that module at different priorities.

Since the flag resides in the `struct sr_mod_info_s`, it is local to the
publisher and for this commit, and does not affect other publishers or
future events as they will prepare a different mod_info structure.
The only exception to this is `sr_shmsub_rpc_internal_call_callback`
which reuses the mod_info structure, by manually resetting fields.

- handling SR_SUBSCR_CHANGE_ALL_MODULES subscription flag

Additional care must be taken due to the SR_SUBSCR_CHANGE_ALL_MODULES
flag. This allows for generation of two kinds of diffs:
1. per-module diff
2. all-modules diff

For example, It would be wrong to pass a per-module diff to a
subscription which has requested an `all-modules` diff.
To handle this case, only `per-module` diff is reused. `all-module` diff
will need to be copied into the sub_data_shm each time. However, the
libyang data tree is walked for printing only twice (once for the first
CHANGE event, and then again for the first DONE event).

This should be okay as SR_SUBSCR_CHANGE_ALL_MODULES is not very
performant anyway and making optimizations for its preparation will
probably not make much of a difference.

Note:
If "all-modules" subscriptions co-exists with "normal" subscriptions,
at the same priority, the full diff to be published.
This behaviour is not changed by this commit.
If the next priority subscriptions don't request "all-modules", the
per-module diff will need to be regenerated before notification.

### DIFF
--- a/src/modinfo.h
+++ b/src/modinfo.h
@@ -58,6 +58,7 @@ struct sr_mod_info_s {
         uint32_t xpath_count;   /**< Count of XPaths. */
         uint32_t state;         /**< Module state (flags). */
         uint32_t request_id;    /**< Request ID of the published event. */
+        uint32_t reuse_diff;    /**< Whether a reusable diff has been written into the shm for this request_id. */
     } *mods;                    /**< Relevant modules. */
     uint32_t mod_count;         /**< Modules count. */
 };
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -1209,6 +1209,7 @@ sr_shmsub_change_notify_diff_has_changes(struct sr_mod_info_mod_s *mod, const st
  * @param[in] diff Full diff.
  * @param[in] ly_mod Module of the subscription.
  * @param[in] sub_opts Subscription options.
+ * @param[in,out] reuse_diff Whether diff is reusable, if requested.
  * @param[in,out] full_diff_lyb Printed full diff so that it can be reused.
  * @param[in,out] full_diff_lyb_len Length of @p full_diff_lyb.
  * @param[out] diff_lyb Printed diff to use.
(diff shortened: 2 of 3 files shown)


michalvasko (changes requested):
It took me a few minutes to understand you want to share the diff across more events of a single commit. Why not but I think it makes sense only for diff written by CHANGE to be reused by DONE (UPDATE likely changes the diff and ABORT uses its own reversed diff). But that should be okay except, to make the flag more clear and robust, I would use a variable called `shm_data_diff_req_id`, which would be the `request_id` of the last CHANGE event that has written its diff. That should make it safe for the DONE event with the same `request_id` to reuse the diff. And this value does not have to be cleared, it is obsoleted automatically.

src/modinfo.h
```diff
@@ -58,6 +58,7 @@ struct sr_mod_info_s {
         uint32_t xpath_count;   /**< Count of XPaths. */
         uint32_t state;         /**< Module state (flags). */
         uint32_t request_id;    /**< Request ID of the published event. */
+        uint32_t reuse_diff;    /**< Whether a reusable diff has been written into the shm for this request_id. */
```
irfanHaslanded, line 61: Is this the flag you want me to change to `shm_data_diff_req_id`?

michalvasko: Yes, that is what I was referring to.

irfanHaslanded: I don't think it will be automatically outdated for `sr_shmsub_rpc_internal_call_callback`, which is the only function for which resetting this flag is required.

michalvasko: Yes, I mistakenly thought the flag is global.

michalvasko:
> This flag is set
after the diff is prepared by the first CHANGE event for a module and
is reused for all subsequent CHANGE and DONE events which will be
published afterwards.

One needs to be careful with the printed diff. In case there are more subscriptions with different priorities and different flags (some requesting full diff, some not), it may not be possible to share the diff.

src/modinfo.h
irfanHaslanded: It was not clear from the previous commit message, but I have added some more explanation to the commit now. Hopefully it is adequate now.

irfanHaslanded:
> > This flag is set
> > after the diff is prepared by the first CHANGE event for a module and
> > is reused for all subsequent CHANGE and DONE events which will be
> > published afterwards.
> 
> One needs to be careful with the printed diff. In case there are more subscriptions with different priorities and different flags (some requesting full diff, some not), it may not be possible to share the diff.

You are right. I will need to fix this, and add a test for `SR_SUBSCR_CHANGE_ALL_MODULES` option.

src/shm_sub.c
```diff
@@ -1606,18 +1606,30 @@ sr_shmsub_change_notify_change(struct sr_mod_info_s *mod_info, const char *orig_
...
+                /* prepare the diff to write into subscription SHM */
+                if ((err_info = sr_shmsub_change_notify_get_diff(mod_info->diff, nsub->mod->ly_mod, opts, &full_diff_lyb,
+                        &full_diff_lyb_len, &diff_lyb, &diff_lyb_len, &free_diff))) {
+                    goto cleanup;
+                }
+
+                /* set mod->reuse_diff only if we prepared a per-module diff */
+                nsub->mod->reuse_diff = !(opts & SR_SUBSCR_CHANGE_ALL_MODULES);
+            }
+
```
michalvasko, line 1632: Not sure why you moved the code generating the `diff` after sub SHM open but I suppose it makes no difference. But please move all this new logic into `sr_shmsub_change_notify_get_diff`, no reason not to and it will not make these notify functions longer, they already are long enough.

michalvasko (changes requested):

src/shm_sub.c
irfanHaslanded: I moved the code generating diff below sub shm open, because earlier I was also storing the diff_lyb_len in the shm, to pass onto `sr_shmsub_notify_write_event` to prevent any accidental remaps.

However, I realized that this was not necessary, because xpath is always NULL for change_sub events. So as long as the diff_lyb_len remains 0, we should not have a problem.

But it may still make sense to write the diff_lyb_len into the SHM for a paranoid/sanity check?

michalvasko (approved):

PR #3464, opened 2024-12-02
sr_shmsub_change_notify_evpipe must track the number of subscribers notified

irfanHaslanded:
- sr_shmsub_change_notify_evpipe track num_notified

    `sr_shmsub_change_notify_evpipe` must track the number of subscribers
    notified which can be lesser than the expected `subscriber_count`
    written into the SHM, if any subscribers exit without cleaning up.

    In this case, we must update the expected `subscriber_count`.

    However, if no subscribers were notified, we must cancel the written
    event and clear up the sub SHM.

- bugfix goto in sr_shmsub_change_notify_update

    A write lock is held at the point, and the goto should be
    cleanup_wrunlock and not cleanup.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -1118,28 +1118,29 @@ cleanup:
 /**
  * @brief Write into change subscribers event pipe to notify them there is a new event.
  *
- * @param[in] conn Connection to use.
+ * @param[in] mod_info
  * @param[in] mod Mod info module to use.
- * @param[in] ds Datastore.
- * @param[in] diff Event diff.
  * @param[in] ev Change event.
  * @param[in] priority Priority of the subscribers with new event.
  * @return err_info, NULL on success.
  */
 static sr_error_info_t *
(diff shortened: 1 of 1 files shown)


src/shm_sub.c
```diff
@@ -1544,6 +1556,17 @@ sr_shmsub_change_notify_nsubs_set_mod_prio(struct sr_shmsub_many_info_change_s *
     *max_mpriority = cur_mprio - 1;
 }
 
+static void
+sr_shmsub_change_cancel_written_event(struct sr_shmsub_many_info_change_s *nsub, sr_cid_t cid)
```
michalvasko, line 1560: I don't like that this function is so short and includes an unlock so I would prefer to just use the code instead of the function unless you have a better idea.

michalvasko (changes requested):
Thanks, logically it seems fine, just some minor improvements.

michalvasko (approved):

PR #3472, opened 2024-12-06
check sub is_alive before copying running to oper

irfanHaslanded:
While fetching operational data, data from the running datastore is copied automatically if there is a non passive subscription to the running datastore for that module.

But it is also possible that this subscription is no longer alive, in which case we should not be reporting this data in the operational datastore.


michalvasko (approved):

michalvasko:
Fine, thanks.

PR #3483, opened 2024-12-18
ds_mongo support multiple discard-items

irfanHaslanded:
sysrepo discard-items are stored with the key /sysrepo:discard-items. This key (_id) should be unique.

But unlike other nodes, multiple sysrepo discard-items nodes can be present.

So, we should also add an index to make these nodes unique. This index is added with a prefix 'O' for easy detection.

While loading data back, look for the prefix 'O' to detect opaque keys and discard the prefix and index to get the original path back.


michalvasko (changes requested):
Thanks, seems okay but please include a similar change for the redis plugin, should be almost the same.

irfanHaslanded:
Since the tests for `redis` pass, I assumed this was not required for the `redis` plugin.

michalvasko:
I see, I suppose it is fine then.

michalvasko (approved):

PR #3484, opened 2024-12-18
notify subscribers of oper discard-items and remove redundant push data

irfanHaslanded:
Remove any "pushed" operational data matching the xpath of discard-items nodes from the mod_info->data.

This will allow creation of a diff reporting data being discarded.

Without this, discarded data are still present in the SHM, which is also not ideal.

For example:
cat /dev/shm/sr_ietf-interfaces.operational.1-1 | jq {
  "ietf-interfaces:interfaces-state": {
    "@": {
      "ietf-origin:origin": "ietf-origin:unknown"
    },
    "interface": [
      {
        "name": "eth1",
        "type": "iana-if-type:ethernetCsmacd"
      },
    ]
  },
  "sysrepo:discard-items": "/ietf-interfaces:interfaces-state/interface[name='eth1']"
}

Here, we see `eth1` data still present along with the discard-items.

This slows down reading from the operational datastore, which most likely happens far more frequently than writing to it.

With this change, operational data is simplified to cat /dev/shm/sr_ietf-interfaces.operational.1-1 | jq {
  "sysrepo:discard-items": "/ietf-interfaces:interfaces-state/interface[name='eth1']"
}

This will still not inform subscribers of changes to enabled data in the operational datastore.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -632,6 +632,114 @@ cleanup:
     return err_info;
 }
 
+/**
+ * @brief Gets the XPath value of a "/sysrepo:discard-items" node, if the XPath refers to the given module_name.
+ *
+ * First checks that the node is opaque, and is a sysrepo "discard-items" node with a XPath.
+ * Then if the referred module in that XPath is the given module_name, it returns the XPath.
+ * In all other cases returns NULL.
+ *
+ * @param[in] node mod_info->diff node to check
+ * @param[in] module_name Name of the module the XPath must refer to
+ * @return xpath, on success, NULL otherwise
+ */
+static const char *
+sr_modinfo_mod_discard_xpath_get(struct lyd_node *node, const char *module_name)
+{
+    const struct lys_module *ly_mod;
+    const char *xpath = NULL;
+
+    if (node->schema) {
+        return NULL;
+    }
+
+    ly_mod = lyd_owner_module(node);
+    if (!ly_mod || strcmp(ly_mod->name, "sysrepo") || strcmp(LYD_NAME(node), "discard-items")) {
+        /* other opaque nodes */
+        return NULL;
+    }
+
+    xpath = lyd_get_value(node);
+    if (!xpath || !xpath[0]) {
+        /* invalid XPath */
+        return NULL;
+    }
+
+    if (!sr_xpath_refs_mod(xpath, module_name)) {
+        /* not interested in other modules at this time */
+        return NULL;
+    }
+
+    return xpath;
+}
+
+/**
+ * @brief Process the mod_info data and diff according to the "/sysrepo:discard-items" nodes present
+ *
+ * For the specified module:
+ * 1. Remove any nodes in mod_info->data matching the XPath in discard-items
+ *    This makes reading operational data more efficient, as the redundant nodes are removed before writing.
+ * 2. Add a "delete" operation to the nodes from mod_info->diff so that subscribers are notified of the discarded data.
+ *
+ * @param[in] mod_info Mod info to use.
+ * @param[in] module_name Name of the module to process the discard-items nodes for
+ * @return err_info, NULL on success.
+ * */
+static sr_error_info_t *
+sr_modinfo_process_mod_discards(struct sr_mod_info_s *mod_info, const char *module_name)
+{
+    sr_error_info_t *err_info = NULL;
+    struct lyd_node *next = NULL, *node = NULL;
+    struct ly_set *set = NULL;
+    uint32_t j;
+    const char *xpath = NULL;
+    struct lyd_meta *meta = NULL;
+    enum edit_op op;
+
+    LY_LIST_FOR_SAFE(mod_info->diff, next, node) {
+        xpath = sr_modinfo_mod_discard_xpath_get(node, module_name);
+        if (!xpath) {
+            /* Not a '/sysrepo:discard-items' node or xpath doesn't refer to module_name */
+            continue;
+        }
+
+        if ((op = sr_edit_diff_find_oper(node, 0, NULL) == EDIT_DELETE)) {
+            /* discard-items is being deleted in this edit and not being added */
+            continue;
+        }
+
+        /* Step 1. from mod_info->data, remove all the discarded nodes */
+        if ((err_info = sr_lyd_xpath_complement(&mod_info->data, xpath))) {
+            return err_info;
+        }
+
+        /* Step 2. into mod_info->diff, add operation delete */
+
+        /* Find the nodes matching the xpath from the diff */
+        if ((err_info = sr_lyd_find_xpath(mod_info->diff, xpath, &set))) {
+            return err_info;
+        }
+
+        /* get rid of all redundant results that are descendants of another result */
+        if ((err_info = sr_xpath_set_filter_subtrees(set))) {
+            return err_info;
+        }
+
+        /* Set meta to operation="delete" */
+        for (j = 0; j < set->count; ++j) {
+            meta = lyd_find_meta(set->dnodes[j]->meta, NULL, "yang:operation");
+            if (meta) {
+                lyd_change_meta(meta, "delete");
+            } else if ((err_info = sr_diff_set_oper(set->dnodes[j], "delete"))) {
+                return err_info;
+            }
+        }
+        ly_set_free(set, NULL);
+    }
+
+    return err_info;
+}
+
 sr_error_info_t *
 sr_modinfo_oper_edit_apply(struct sr_mod_info_s *mod_info, const struct lyd_node *data, int create_diff)
 {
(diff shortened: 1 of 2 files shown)


michalvasko:
I am not sure you understand what `discard-items` is supposed to do, have you read the [docs](https://netopeer.liberouter.org/doc/sysrepo/devel/html/push_oper_data.html)?

> Without this, discarded data are still present in the SHM, which is also not ideal.

The discarded data **are supposed to be present** in the SHM. The `discard-items` nodes essentially modify the data for the session, not the data of other sessions.

irfanHaslanded:
You are right, this is still not clear to me in spite of reading the docs multiple times.

> Additionally, using [sr_discard_items()](https://netopeer.liberouter.org/doc/sysrepo/devel/html/group__edit__data__api.html#ga3c93aa25830a13af79add17e3269f9d4) an XPath can be set that will select all the nodes to be removed from previous data, which may be enabled running data or any push oper data with a lower order (higher priority).

>  an XPath can be set that will select all the nodes to be removed from previous data

1. I don't think it is clear _"when"_ this removal should happen, and more specifically _"why"_ should the removal not happen while writing to the operational datastore and only happen when operational data is read.

>  The discard-items nodes essentially modify the data for the session, not the data of other sessions.

2. This is completely the opposite of what I understood from reading the docs.

>  which may be enabled running data or any push oper data with a lower order (higher priority).

and 

> When building the operational datastore, all the stored push data are applied in a specific order. That means data for each session are loaded, previous items discarded (if any set), and then merged into the operational datastore. Using a combination of discarding items and merging data, it should be possible to fully customize the final operational datastore.

By _"previous items discarded"_, don't you mean data loaded until now from all sources (other push sessions and enabled data)?

jktjkt:
The `operational` DS still works with *edits*, not with *content*. This is important because these edits are applied on top of some "previous data", and this "previous data" might change independently of your stored edit. For example, the very first layer of these data is the content of `running`. If you push a change which "discards" some previous data, you're storing an "intention to remove", which will make sure that this particular subtree will *never* make it past the current edit -- no matter whether the data are actually present in the previous content, or not.

The word "discard" is now, unfortunately, overridden for two completely different meanings in the public API. The `sr_discard_items` is used for the purpose of ensuring that the data from previous content will not be present after the current edit is applied. The `sr_discard_changes` or `sr_discard_changes_xpath` manipulates the edit of the current session.

>> The discard-items nodes essentially modify the data for the session, not the data of other sessions.
>
> This is completely the opposite of what I understood from reading the docs.

That explanation might be a bit misleading. I guess that what @michalvasko was saying is that these data *cannot* be physicaly removed from the SHM, in-place. The reason for that is that if the current session is closed, its edits must be undone. If the edit physically removed the data from SHM, these data would be gone now, and could not be restored. So the operation very much "modifies the final visible content of the operational DS", but it does that in a way which does *not* modify the stored data in-place.

michalvasko:
> I don't think it is clear "when" this removal should happen, and more specifically "why" should the removal not happen while writing to the operational datastore and only happen when operational data is read.

I think the "when" is explained in detail in the first sentence you quoted. But if you have ideas for improving the docs, please share them so it can be improved. My rationale for not reporting any changes to subscribers for `discard-items` nodes is that only the changes of the data pushed by the session are reported. Since the data of other sessions/enabled running data are discarded, these are not reported.

The removed data **at the moment** the `discard-items` are pushed could be reported but the downside is that the actually discarded data can change in time without the push oper data of the session changing. But since there is no perfect solution, I have no problem with this change.

> This is completely the opposite of what I understood from reading the docs.

We are disagreeing on irrelevant terms. What I simply meant is that the `discard-items` node is stored in the data of the session but yes, it is essentially applied on data of other sessions/enabled running data.

> By "previous items discarded", don't you mean data loaded until now from all sources (other push sessions and enabled data)?

Yes, that is what it means.

irfanHaslanded:
Thanks @michalvasko  and @jktjkt for the detailed explanation.

>That explanation might be a bit misleading. I guess that what @michalvasko was saying is that these data cannot be physicaly removed from the SHM, in-place. The reason for that is that if the current session is closed, its edits must be undone. If the edit physically removed the data from SHM, these data would be gone now, and could not be restored. So the operation very much "modifies the final visible content of the operational DS", but it does that in a way which does not modify the stored data in-place.

I will think more on this and how it affects us.

irfanHaslanded:
@michalvasko @jktjkt 

> The word "discard" is now, unfortunately, overridden for two completely different meanings in the public API. The sr_discard_items is used for the purpose of ensuring that the data from previous content will not be present after the current edit is applied. The sr_discard_changes or sr_discard_changes_xpath manipulates the edit of the current session.

To make the intention and result of calling the functions very clear, is it possible to rename them to accurately reflect what they do

`sr_discard_changes()` -> `sr_erase_staged()` : Erase indicates that it is irreversible. Staged indicating that the changes are not yet committed.
`sr_discard_changes_xpath()` -> `sr_erase_staged_xpath()` 
`sr_discard_oper_changes()` -> `sr_erase_oper_pushed()`

And keep only `sr_discard_items` with discard in name.

michalvasko:
I do not think the naming confusion is such a major concern. Also, as discussed with @jktjkt I would prefer to avoid renaming functions right now and increasing the major SO version. Please focus on the functionality.

jktjkt:
In the C++ bindings, we used the phrase "drop foreign operational content" for `sr_discard_items`. Maybe an even better word might be "block" because that (hopefully) shows that the data are "still somewhere", but they are "not propagated any further". I'm not a native speaker, though. The important port is IMHO not whether we call this "erase", "drop" or "discard", but to ensure that we do not use the very same phrase to refer to multiple different things, which is why I dislike @irfanHaslanded 's proposal with the word "erase" used for everything.

IMO, given the huge difference in semantics, calling this API break an API break and bumping the SOVERSION is the obvious thing to do, but it seems that the maintainer disagrees.

irfanHaslanded:
Sure, I hope renaming will be considered at the time of next major version bump, whenever that happens.

> The removed data at the moment the discard-items are pushed could be reported but the downside is that the actually discarded data can change in time without the push oper data of the session changing. But since there is no perfect solution, I have no problem with this change.

@michalvasko So, Is the functionality change this PR brings okay?

michalvasko:
The idea is fine, like I said, but I do not really like the implementation. There seems to be a lot of (new) code and not called at the place I would expect it to so I will implement it myself.

michalvasko:
Actually, this is a much bigger change than assumed. Currently, the changes in push oper data of a session are reported only compared to the previous push oper data of the session (why `discard-items` is not reported at all). With this change the diff will have to be created by comparing the whole operational DS with the previous push oper data of the session and with the new push oper data.

In other words, while now only the push oper data of the session need to be loaded, it will be required to build the whole operational DS twice and then compare them.

Like I said, I have no problem with this but the performance consequences are obvious, so are you fine with this @irfanHaslanded?

irfanHaslanded:
Yes, I am fine with the performance consequence, given that not reporting changed operational datastore to subscribers (because of discarded operational data) is not an option for us.

jktjkt:
Same here -- as a subscriber, what I'm interested in is the "final state" of the operational DS. From a subscriber's PoV, it is not important *what* caused a change, but it's definitely a bug if I'm not notified about "a visible change" in the DS content.

michalvasko:
That is all fine in theory but in practice it is more complex than that. What about oper data get subscriptions? What would you expect?
```
enabled running data + stored oper push data + session OLD oper push data + oper get subscriptions data
   ^
  diff (reported changes)
   v
enabled running data + stored oper push data + session NEW oper push data + oper get subscriptions data
```
or without `oper get subscriptions data`? I would prefer without them because with them seems like an overkill, not just regarding the performance. But, if you insist on "being notified about every visible change" accurately (at least at that particular moment), they need to be included, too.

michalvasko:
Actually, the better choices would probably be
```
enabled running data + lower-priority stored oper push data + session OLD oper push data + higher-priority stored oper push data + oper get subscriptions data
   ^
  diff (reported changes)
   v
enabled running data + lower-priority stored oper push data + session NEW oper push data + higher-priority stored oper push data + oper get subscriptions data
```
vs
```
enabled running data + lower-priority stored oper push data + session OLD oper push data
   ^
  diff (reported changes)
   v
enabled running data + lower-priority stored oper push data + session NEW oper push data
```
The difference is that the former reports fully accurate changes in the oper DS data while the latter reports `discard-changes` accurately and just changes in the push oper data of the session (which may not be accurate or even be none in the final oper DS).

Thoughts? @jktjkt @irfanHaslanded

irfanHaslanded:
I don't think changes in "oper get" data can ever be properly notified, as this usually contains frequently changing data (for example interface statistics) and the system design expects this. So any notification is likely to be outdated.

So, IMHO, we should be fine not notifying "oper get" data changes, even if they are blocked due to some pushed data / configuration.

irfanHaslanded:
Will considering "enabled running data" make locking more complex (meaning reading from operational datastore/ writing to operational datastore) needing locks on Running data and change_sub?

michalvasko:
Naturally, the relevant `running` data needs to be normally loaded as the first step, from which only the enabled data are then selected.

irfanHaslanded:
I think we should skip `"enabled running data"`, due to the fact that we will need to hold a lot more locks and for a lot more time, which I am afraid will cause deadlocks and increased lock contention and IMHO degraded overall performance.

Prior to v3.x, we did not get notified of changes to this either, So, it should be okay, I think.

michalvasko:
Okay, I suppose, even though I think you are too focused on your specific use-case and ignore any other possible requirements. The degraded performance was my initial argument which you dismissed back then.

michalvasko:
Took quite some effort but should be implemented and working.

PR #3485, opened 2024-12-18
tests dont check for same priority callbacks order

irfanHaslanded:
If there are multiple subscribers at the same priority, they can be called in any order. So, don't test for the order of invocation.


michalvasko (approved):

michalvasko:
I was just trying to reproduce the failing of this test so I guess this fixes it, thanks.

PR #3490, opened 2025-01-07
build BUGFIX link libatomic when found

gotthardp:
The `libatomic` is not linked when found by the first `check_library_exists` call.
The bug was introduced by commit bf79330.


michalvasko (approved):

michalvasko:
Right, thanks.

PR #3494, opened 2025-01-13
shmext REFACTOR conn_remap_unlock remove last hole

irfanHaslanded:
The logic in sr_shmext_conn_remap_unlock can be cleaner.

A call to sr_file_get_size() can be avoided since we already know it as `conn->ext_shm.size`.

Also cleanup the somewhat complex cast + arithmetic in a if condition.

### DIFF
--- a/src/shm_ext.c
+++ b/src/shm_ext.c
@@ -131,8 +131,8 @@ sr_shmext_conn_remap_unlock(sr_conn_ctx_t *conn, sr_lock_mode_t mode, int ext_lo
 {
     sr_error_info_t *err_info = NULL;
     sr_ext_hole_t *iter, *last = NULL;
-    uint32_t last_size;
-    size_t shm_file_size = 0;
+    uint32_t new_size;
+    char *last_hole_end;
 
     /* make ext SHM smaller if there is a memory hole at its end */
     if (((mode == SR_LOCK_WRITE) || (mode == SR_LOCK_WRITE_URGE)) && ext_lock) {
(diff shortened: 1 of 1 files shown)


michalvasko (approved):

src/shm_ext.c
```diff
@@ -131,26 +131,27 @@ sr_shmext_conn_remap_unlock(sr_conn_ctx_t *conn, sr_lock_mode_t mode, int ext_lo
...
     if (((mode == SR_LOCK_WRITE) || (mode == SR_LOCK_WRITE_URGE)) && ext_lock) {
         while ((iter = sr_ext_hole_next(last, SR_CONN_EXT_SHM(conn)))) {
             last = iter;
         }
 
-        if (last && ((uint32_t)((char *)last - conn->ext_shm.addr) + last->size == conn->ext_shm.size)) {
+        last_hole_end = last ? last + last->size : NULL;
+
+        if (last_hole_end == conn->ext_shm.addr + conn->ext_shm.size) {
             if ((err_info = sr_file_get_size(conn->ext_shm.fd, &shm_file_size))) {
```
irfanHaslanded, line 146: @michalvasko For some reason `conn->ext_shm.size` and `sr_file_get_size(conn->ext_shm.fd, &shm_file_size)` get out of sync, causing the failure in test_process, although we have a write lock

michalvasko: Interesting that it manifests only in the external DB tests. I have no ideas so you will have to investigate this yourself, or I can later.

irfanHaslanded: I am checking it.

irfanHaslanded: I believe this was the reason:

```c
last_hole_end = last ? last + last->size : NULL;
```
is really not the same as 
```c
last_hole_end = last ? (char *)last + last->size : NULL;
```
as `*last` is `sr_ext_hole_t` size and not a `char` size.

michalvasko: Sorry, I meant
```C
last_hole_end = last ? ((char *)last + last->size) : NULL;
```
to have the whole operand of the ternary operator in parentheses.

irfanHaslanded: Ah, okay. Done

michalvasko (approved):

PR #3495, opened 2025-01-13
modinfo UPDATE sess push oper data module cache

irfanHaslanded:
Maintain a list of modules whose push oper data were modified by the session.

Acquiring the global ext_lock is significantly more expensive.

Previously this was implemented for connections in 3b0165d656709b5f0cd3e8debc2e55c76c73663a

But in sysrepo v3.x, oper push data is tied to the session and the logic needs to be re-introduced.


michalvasko (approved):

michalvasko:
Fine, if you think this necessary.

PR #3497, opened 2025-01-13
sr_module_oper_data_load - shorten read ext_lock

irfanHaslanded:
We don't need to hold onto `ext_lock` for the entire duration of `sr_module_oper_data_load`.

We only need it to get the `sr_mod_oper_push_t` array for the module.

So quickly make a copy of it and unlock, instead of holding onto ext SHM lock while we load the operational data.

Modules with no push data can be skipped without acquiring ext_lock.


michalvasko:
Okay, but I implemented it myself, did not like some code in your PR.

irfanHaslanded:
> Okay, but I implemented it myself, did not like some code in your PR.

Thanks! Yours is much better, and it also breaks the recursive `sr_module_oper_data_load` for dead CID handling.

PR #3498, opened 2025-01-13
upgrade ext_lock from mutex to rwlock

irfanHaslanded:
`ext_lock` is currently a *process-shared* mutex for accessing holes and
truncating the ext SHM.

Only enlarging the ext SHM requires a `WRITE` lock, whereas all other
operations requires a `READ` lock on the system-wide `ext_lock`.

Prior to sysrepo v3, this meant that the `ext_lock` was mainly
locked for subscriptions.
However, with sysrepo v3.x, reading operational data also requires
locking ext SHM.

So, a `ext_lock` *mutex* severely degrades the overall performance when
there are multiple processes using sysrepo.

This is easily fixed by upgrading the ext_lock to a full `rwlock`, which
will allow parallel reads and only blocks writers, lowering contention.

But, this creates a problem with creating and merging ext SHM holes
during deletion of entries from the ext SHM. This is because deletes
happen while holding a `READ` lock.
This is solved by creating a `ext_hole_lock` that is held only when
creating these ext holes.

It is not required to hold the `ext_hole_lock` if `ext_lock` is held in
`WRITE` mode. When `ext_lock` is held in `READ` mode, `ext_hole_lock` must be
acquired.

### DIFF
--- a/src/shm_ext.c
+++ b/src/shm_ext.c
@@ -697,8 +697,8 @@ sr_shmext_change_sub_free(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, sr_datastore_t
 
     shm_sub = &((sr_mod_change_sub_t *)(conn->ext_shm.addr + shm_mod->change_sub[ds].subs))[del_idx];
 
-    /* EXT_HOLE WRITE LOCK */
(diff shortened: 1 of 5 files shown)


irfanHaslanded:
Will use a different approach

PR #3499, opened 2025-01-14
shm_ext remove sr_shmext_recover_sub_all

irfanHaslanded:
This function and all its dependent code are not needed anymore.

Before live context update was supported, this code was necessary to recover any dead connections.

`sr_shmext_recover_sub_all()` is only called from `sr_connect()` if the main_shm was created.

If main_shm is created, both `sr_shmmod_open()` and `sr_shmext_open()` will create empty sr_mod and sr_ext SHM files.

So, `sr_shmext_recover_sub_all()` really has nothing left to do. since `SR_CONN_MOD_SHM(conn)->mod_count` will be zero.

`sr_shmext_check_sub_all()` already handles the updated context.


michalvasko (approved):

michalvasko:
Seems reasonable, thanks.

PR #3500, opened 2025-01-15
ext_lock is not needed for oper push READ or DEL

irfanHaslanded:
The ***system-wide*** `ext_lock` is needed only when
1. ext SHM needs to be remapped or truncated
2. ext SHM holes are added/deleted/merged

Reading oper push data and deleting the oper push data from the mod shm does not cause either of the above two operations.

So, ***system-wide*** `ext_lock` is not needed.

The process specific (among different threads of the same process) lock `conn->ext_remap_lock` already ensures that reading the `ext_shm` is safe and the current process will not remap to learn new changes from other processes.

The appropriate module locks that are already held, ensure that the ext SHM locations do not change while being read.

This means that
1. `sr_module_oper_data_load`
2. `sr_shmext_oper_push_get`
3. `sr_shmmod_del_module_oper_data`

do not need to acquire the ***system-wide*** `ext_lock`.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -1264,7 +1264,7 @@ sr_module_oper_data_load(struct sr_mod_info_mod_s *mod, sr_conn_ctx_t *conn, uin
     int last_sid = 0, dead_cid = 0;
 
     /* EXT READ LOCK */
-    if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_READ, 1, __func__))) {
+    if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_READ, 0, __func__))) {
         goto cleanup;
     }
 
@@ -1288,7 +1288,7 @@ sr_module_oper_data_load(struct sr_mod_info_mod_s *mod, sr_conn_ctx_t *conn, uin
     }
 
     /* EXT READ UNLOCK */
-    sr_shmext_conn_remap_unlock(conn, SR_LOCK_READ, 1, __func__);
+    sr_shmext_conn_remap_unlock(conn, SR_LOCK_READ, 0, __func__);
 
     if (err_info) {
         goto cleanup;
--- a/src/shm_ext.c
+++ b/src/shm_ext.c
@@ -2740,7 +2740,7 @@ sr_shmext_oper_push_get(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, const char *mod_
     }
 
     /* EXT READ LOCK */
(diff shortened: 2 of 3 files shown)


michalvasko:
Actually, I do not think there is a use-case where the EXT SHM lock is not needed, now that I think about it. If we are talking about 2 processes (so we can ignore the REMAP lock), what prevents one process to read ext SHM at the same time the other process is changing it?

I think the idea is that it does not matter as long as nothing that is being read is changed. And that should be true, in general, because when accessing change subscriptions, their lock is being held and so on for most of the structures in ext SHM. But the new push oper data arrays have no dedicated lock and looking at the structures, I think they are the only ones. So, there are already a few ideas for improvements (such as mentioning that all ext data need to be protected by a dedicated lock in the `ext_lock` description). And your PR is probably fine **if** that dedicated lock is added as well.

irfanHaslanded:
> the new push oper data arrays have no dedicated lock 

But they are protected by the module WRITE lock when pushing operational data, and the module READ lock when reading the operational data.
Am I mistaken?

irfanHaslanded:
The actual use-case for ext_lock is, just as it was originally documented:

```c
    pthread_mutex_t ext_lock;   /**< Process-shared lock for accessing holes and truncating ext SHM. */
```
We need it every time holes are created (ext data is deleted) or merged (adjacent holes) or used (data is added into a hole).

We don't need to hold it in any other case.

michalvasko:
> But they are protected by the module WRITE lock when pushing operational data, and the module READ lock when reading the operational data.

Yes, it seems that way and it makes sense but I have forgotten all the stuff, there were too many things I needed to change for these push oper data... But I have updated the docs to include the information I forgot this time so I can look at these PRs and yes, ext SHM locks are not needed then.

michalvasko (approved):

irfanHaslanded:
Thanks!

I will go through the docs once I am finished with these changes, and check if everything is clear.

It is indeed very easy to get lost among all the different types of locks.

irfanHaslanded:
For further dropping `ext_lock`, 
`sr_shmext_print()` will assert some false positives, as it does not acquire the *dedicated locks*.
For example, a new change_sub could have been added causing the ext SHM to be remapped to a larger size `S2` by process P2.

If P1 had just removed a rpc sub, it would assert here:
```c
    /* check that no item exists after the mapped segment */
    assert((unsigned)cur_off == shm_ext->size);
```
the cur_off will be the S2, which would be picked up by the traversal logic in `sr_shmext_print()`.

This will need to be fixed TBD how.

In fact, copying data for printing from each of the ext SHM's items is done without a lock and is a problem for `sr_shmext_print()`.

PR #3501, opened 2025-01-15
shm session_oper_order needs context lock not ext

irfanHaslanded:
`sr_shmmod_session_oper_order()` *get/set* operations need the context lock using `sr_lycc_lock()` and not the *ext remap lock*.

The *context lock* is needed because shm_mod->mod_count should not change while we are traversing the modules.
That is modules should not installed/uninstalled.

The *ext remap lock* is meant for accessing the ext SHM, which doesn't happen here, but only later in
`sr_shmext_oper_push_get()` and `sr_shmext_oper_push_update()`

These two functions acquire the appropriate *ext remap lock* correctly.

### DIFF
--- a/src/shm_mod.c
+++ b/src/shm_mod.c
@@ -1882,12 +1882,6 @@ sr_shmmod_session_oper_order(sr_session_ctx_t *session, const struct lys_module 
         }
     } else {
         mod_shm = SR_CONN_MOD_SHM(session->conn);
-
-        /* EXT READ LOCK */
-        if ((err_info = sr_shmext_conn_remap_lock(session->conn, SR_LOCK_READ, 1, __func__))) {
-            goto cleanup;
-        }
-
         /* update all the modules with any push oper data */
         for (i = 0; i < mod_shm->mod_count; ++i) {
             shm_mod = SR_SHM_MOD_IDX(mod_shm, i);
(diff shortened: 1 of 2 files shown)


irfanHaslanded:
ABI check Failure is due to

<body>
<h2>Problems with Symbols, High Severity <span class="failed"> 2 </span></h2><hr>
<span class="h_name">sysrepo.h</span>, <span class="lib_name">libsysrepo.so.7.29.4</span><br>
<span class="section" onclick="javascript:showContent(this, 'c_1')">
<span class="ext">[−]</span> sr_get_oper_changes_order <span class="sym_p"><span>( sr_session_ctx_t* <span class="color_p">session</span></span>, <span>char const* <span class="color_p">module_name</span></span>, <span>uint32_t* <span class="color_p">order</span></span> )</span> <span class="failed"> 2 </span></span>
<br>
<div id="c_1" style="display: block; visibility: visible;">


  | Change | Effect
-- | -- | --
The parameter module_name became passed in rbp register instead of r12. | Applications will read the wrong memory block instead of the parameter value.
The parameter order became passed in r12 register instead of rbp. | Applications will read the wrong memory block instead of the parameter value.


<br>
</div>

</body>

michalvasko:
Yeah, I do not understand what these new errors are supposed to mean, why would the registers used be changed, but only the new Ubuntu 24.04 uses versions of the tools that report this.

michalvasko (approved):

PR #3504, opened 2025-01-16
shm_ext locking improvements

irfanHaslanded:
## shm_ext - take ext_lock only when needed

Many shm_ext functions don't require *system-wide* `ext_lock` to be
acquired before changing the data in the ext SHM.

This lock is only required when manipulating ext SHM holes, or
changing the size of the ext SHM (truncate/enlarge).

So, in many cases, it suffices to acquire `ext_lock` just before
calling `sr_shmrealloc_del`, and releasing it immediately after.

This further shortens the window where `ext_lock` is held.

## Lock order `conn->ext_remap_lock` then `SR_CONN_MAIN_SHM(conn)->ext_lock`

`sr_shmext_conn_remap_lock()` acquires `ext_lock` first and then
the connection specific `ext_remap_lock`.

However, The other functions which directly acquire `ext_lock` hold the
`ext_remap_lock` already. This is causes lock-order-inversion!

So, `sr_shmext_conn_remap_lock()` is changed to have the correct order.

This also seems logical:
1. First, Prevent sessions of our connection.
2. Next, Prevent other connections if required.

## sr_shmext_print()

`sr_shmext_print()` may see that there is ext data stored beyond the
currently mapped ext SHM size.
This is because no remap was done after the `ext_lock` was acquired.

This is fixed by only validating if the currently mapped size is the same as the actual ext SHM file size.
```C
validate = (shm_file_size == shm_ext->size);
```

It is important for `sr_shmext_print()` to skip reading *new* data
from *yet-to-be-mapped* offsets.

## shm_ext - add locks for reading suspended state

No lock was obtained while reading the suspended state of subscriptions.
The dedicated READ lock must be obtained to prevent data races.
For instance, another process may remove the data concurrently.

- do not need WRITE remap lock to remove subscriptions even for
`sr_shmext_FOO_sub_check()`. a READ ext remap lock is enough.

- for sub_check() functions, also remember when all locks are unlocked
and we fail to obtain a WRITE lock. We should not try to unlock.
A correct `goto cleanup` is used, but add this to be clear.

### fix-up comments for the ext lock being acquired and released.
explicitly use the term `CONN EXT REMAP` for the connection specific
`rwlock` that guards various sessions of the same connection.

The term `EXT LOCK` only refers to the *process-shared* (across multiple
connections) `mutex` from now.

When `ext_remap_lock` is acquired in WRITE mode along with ext_lock, it is
referred to as `REMAP(WRITE) AND EXT LOCK`.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -5569,7 +5569,7 @@ sr_generate_notif_has_subs_or_replay(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, int
     sr_mod_notif_sub_t *notif_subs;
     uint32_t notif_sub_count;
 
-    /* EXT READ LOCK */
+    /* CONN EXT REMAP READ LOCK */
     if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_READ, 0, __func__))) {
         goto cleanup;
     }
@@ -5577,7 +5577,7 @@ sr_generate_notif_has_subs_or_replay(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, int
     /* get subscriber count */
     err_info = sr_notif_find_subscriber(conn, "sysrepo-notifications", &notif_subs, &notif_sub_count, NULL);
 
-    /* EXT READ UNLOCK */
+    /* CONN EXT REMAP READ UNLOCK */
     sr_shmext_conn_remap_unlock(conn, SR_LOCK_READ, 0, __func__);
 
     if (err_info) {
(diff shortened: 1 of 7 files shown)


irfanHaslanded:
> > 1. First, Prevent sessions of our connection.
> > 2. Next, Prevent other connections if required.
> 
> I do not agree with this. The question to focus on is what happens if one lock is held and the event the other lock protects from happens before it is acquired. If you have EXT lock and not the REMAP lock, the mapping may be changed but you should not be accessing it before the REMAP lock anyway so it is perfectly fine. If you have REMAP lock before the EXT lock, it is possible another process changes EXT SHM and only then releases the EXT lock. What you end up with is a modified SHM mapping but being unable to remap it, which, I think, could result in then accessing the modified structure, which could be stored in space after your mapping. So you will end up accessing invalid memory.

Looks like there are some cases missed. I'll think on this some more. Thanks for the pointer.

PR #3505, opened 2025-01-16
modinfo - fast check session has push oper for mod

irfanHaslanded:
Since we cache the names of module for which the session has pushed operational data, it is faster and lock-free to check the cache rather than to read ext SHM.

So, call `sr_modinfo_sess_has_push_oper_data()` instead of `sr_shmext_oper_push_get()`

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -2645,6 +2645,28 @@ cleanup:
 }
 
 /**
+ * @brief Check whether the session has pushed any operational data to this module
+ *
+ * @param[in] sess Session to check.
+ * @param[in] mod_name module name to search.
+ * @return 1 if there is data, else 0
+ */
+static int
+sr_modinfo_sess_has_push_oper_data(sr_session_ctx_t *sess, const char *mod_name)
+{
+    uint32_t i;
+
+    for (i = 0; i < sess->oper_push_mod_count; ++i) {
+        if (!strcmp(sess->oper_push_mods[i].name, mod_name)) {
+            /* module found, return has_data */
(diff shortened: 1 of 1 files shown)


src/modinfo.c
```diff
@@ -2644,6 +2644,28 @@ sr_modinfo_module_data_load_srmon(struct sr_mod_info_s *mod_info, struct sr_mod_
...
+{
+    uint32_t i;
+
+    for (i = 0; i < sess->oper_push_mod_count; ++i) {
+        if (!strcmp(sess->oper_push_mods[i].name, mod_name)) {
+            /* module found, return has_data */
+            return sess->oper_push_mods[i].has_data;
+        }
+    }
+    return 0;
```
michalvasko, line 2666: Empty line before `return`.

src/modinfo.c
```diff
@@ -2644,6 +2644,28 @@ sr_modinfo_module_data_load_srmon(struct sr_mod_info_s *mod_info, struct sr_mod_
...
 }
 
+/**
+ * @brief Check whether the session has pushed any operational data to this module
+ *
+ * @param[in] sess Session to check.
+ * @param[in] mod_name module name to search.
+ *
+ * @return 1 if there is data, else 0
+ * */
```
michalvasko, line 2654: Please keep the doxygen unified, no empty line before `return` and just `*/` at the comment end.

michalvasko (changes requested):
Good optimization, just some minor formatting issues.

michalvasko (approved):

PR #3506, opened 2025-01-16
sr_modinfo_data_store avoid ext wrlock if possible

irfanHaslanded:
A `WRITE ext_lock` is needed only when creating the oper push data for a module by a session the first time, or when changing the order.

Just updating the `has_data` flag does not need a `WRITE ext_lock`. A `READ` lock is sufficient for this.

Most sessions which have once pushed data to a module, will always have some data until the session is stopped.
So, even the `has_data` flag rarely ever changes.

Optimize this by enhancing `sr_modinfo_push_oper_mod_update_sess()` to learn whether the ext SHM data for this oper_push needs to be created or if the has_data flag needs to be changed.

Only the necessary ext SHM operation if any is performed.

A dedicated `sr_shmext_oper_push_change_has_data()` is added for just updating `has_data` flag.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -3845,22 +3845,31 @@ cleanup:
  * @param[in] sess Session to update.
  * @param[in] mod_name Module name.
  * @param[in] has_data Flag to store.
+ * @param[out] created Learn whether oper push data for module was just created (first time) .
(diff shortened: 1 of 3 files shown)


michalvasko (approved):
Seems fine.

PR #3507, opened 2025-01-17
oper push - delete ext shm data at sr_session_stop

irfanHaslanded:
If the session has no oper push data but it has pushed some data in the past, the Ext SHM has entries for this, which must be deleted on stopping the session.

- also don't log "No operational datastore changes to apply" message when stopping the session.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -3914,6 +3914,8 @@ sr_modinfo_data_store(struct sr_mod_info_s *mod_info, sr_session_ctx_t *session,
     sr_datastore_t store_ds;
     uint32_t i, sid;
     int create = 0, change = 0;
+    const char *mod_name = NULL;
+    sr_mod_t *shm_mod;
 
     assert(!mod_info->data_cached);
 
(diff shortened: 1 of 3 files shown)


irfanHaslanded:
@michalvasko I believe `sr_remove_modules` also must call `sr_shmext_oper_push_del` to delete all module oper push info.

michalvasko (approved):

michalvasko:
`sr_remove_modules` removes the SHM module altogether so there is no need to delete its oper push data. But now I remember it was one reason why I have not added the push oper data cache into sessions originally, because on `sr_remove_modules` the removed modules should ideally be removed from all the sessions, which seems like a difficult task. On the other hand, I am not sure it can cause any issues, if removed modules stay in the sessions that changed them before.

irfanHaslanded:
> `sr_remove_modules` removes the SHM module altogether so there is no need to delete its oper push data.
We still would need to delete that data from the ext SHM, right?
Would this not cause any problems elsewhere?

I agree, It should be fine to stay in the session cache, without causing many problems.
Sessions could be updated when they call `sr_lycc_lock()` if really needed.

michalvasko:
Okay, I have actually looked into the code and yes, I think the push oper data arrays of removed SHM modules need to be deleted explicitly.

PR #3509, opened 2025-01-18
shm_ext BUGFIX ext_lock

irfanHaslanded:
### **sr_shmext_check_sub_all BUGFIX hold ext_lock**

The different sub-functions called by `sr_shmext_check_sub_all()`
1. `sr_shmext_change_sub_check()`
2. `sr_shmext_oper_sub_check()`
3. `sr_shmext_notif_sub_check()`
4. `sr_shmext_rpc_sub_check()`

must hold `ext_lock` as they may potentially access *Ext SHM* holes and add
a hole to the *Ext SHM*.

They don't need the `connection remap lock` in `WRITE` mode, as *Ext SHM*
will not be remapped as a result of their operations. A `READ` lock is
enough.

### **shm_ext BUGFIX `sr_shmext__sub_suspended` missing locks**

All the `sr_shmext_FOO_sub_suspended()` functions acquire a `WRITE` module
subscription lock to change the subscription state, but no lock for
reading the subscription suspended state.

They need to acquire a `READ` module subscriptions lock to read safely.
Although the data field itself is atomic, without the `READ` lock, there
is no guarantee that the location in *Ext SHM* being read from refers to
this subscription. For example, It could be moved due to addition of
another subscription to the same module.

### **shm_ext BUGFIX `sr_shmext_rpc_sub_remove_dead` iteration**

If `sr_shmext_rpc_sub_stop()` fails, we must advance the iteration loop
in order to not get stuck trying to remove the same subscription
repeatedly. It could be a persistent problem.

### **`sr_shmext_FOO_sub_remove_dead()` no WRITE lock needed**

The need not happen under `WRITE` *connection remap ext* lock.
A `READ` *connection ext remap* lock is sufficient.
Removing data from the *Ext SHM* does not require remapping the *Ext SHM*.
So, no `WRITE` lock is needed. Since holes may be accessed and created, `ext_lock` is still needed.


src/shm_ext.c
```diff
@@ -547,13 +557,13 @@ sr_shmext_change_sub_add(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, sr_datastore_t
     sr_mod_change_sub_t *shm_sub;
     uint32_t i;
 
+    sr_shmext_change_sub_remove_dead(conn, shm_mod, ds);
+
```
michalvasko, line 561: I am not sure such constructs really make sense, I always thought not. Meaning to hold only READ lock first because it is enough and later unlocking it and getting WRITE lock. Results in a more confusing code, more locks and may ultimately cause the operation to take longer because another session may get the WRITE lock and this one will have to wait for it. So when the operation is fast (traversing an array of module subscription for a DS), I would just stick with the WRITE lock to keep it organized, I do not believe this change made any relevant optimization. Similarly for further similar changes.

src/shm_ext.c
```diff
@@ -1786,10 +1814,11 @@ sr_shmext_change_sub_check(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, sr_datastore_
                 }
                 has_locks = SR_LOCK_WRITE;
 
-                /* EXT WRITE LOCK */
-                if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_WRITE, 0, __func__))) {
+                /* EXT READ LOCK */
+                if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_READ, 1, __func__))) {
                     goto changesub_unlock;
                 }
+                ext_lock = 1;
```
michalvasko, line 1821: I think this is not correct (but was not even before). On the `goto`, you will be unlocking REMAP lock even though it is not locked. I suppose if you change line 1855 to `if (has_locks && ext_lock) {`, it would work correctly. Similarly for the following changes.

michalvasko (changes requested):

src/shm_ext.c
irfanHaslanded: I had 3 reasons for doing this:

1. All `sr_shmext___sub_del` and `sr_shmext__sub_stop` functions will be uniform in having READ remap + ext_lock.
2. Align with `sr_shmext_rpc_sub_remove_dead` which already takes the READ remap + ext_lock.
3. The operation here is not just traversing an array of module subscriptions, but doing a `sr_conn_is_alive()` check which does:
  a. `asprintf()` to get the malloc the connection file path string
  b. `sr_open()`
  c. `fcntl()`

I can leave this as it is, if you still prefer the previous code.

michalvasko: Yeah, you are right there is more to do so okay, I suppose.

src/shm_ext.c
michalvasko: Right, I completely missed that there are 2 labels... Okay then.

michalvasko (approved):

PR #3516, opened 2025-01-23
Test

ReaseySereiNeath:


michalvasko:
This is a really bad PR adding all the build files and having no description at all but I guess you were trying to increase the default oper get callback timeout. But I do not see a reason for that, it should be possible to customize it by user in all API calls.

ReaseySereiNeath:
> This is a really bad PR adding all the build files and having no description at all but I guess you were trying to increase the default oper get callback timeout. But I do not see a reason for that, it should be possible to customize it by user in all API calls.

I have problem with the timeout callback. I don't know what the reason that cause this issuse.
`[ERR]: SR: Callback event "rpc" with ID 6 processing timed out.
[WRN]: SR: Event "rpc" with ID 6 priority 0 failed (Timeout expired).
[ERR]: SR: User callback failed.
[ERR]: NP: Failed to send an RPC (User callback failed).`
If I make a lot of request from Netopeer2 server, It will show this error, do you have any suggestion to show the pin point beside changing the source code?

michalvasko:
Run *netopeer2-server* with the parameter `-t 30`, it is the number of seconds used as the default timeout for all *sysrepo* API calls.

ReaseySereiNeath:
> Run _netopeer2-server_ with the parameter `-t 30`, it is the number of seconds used as the default timeout for all _sysrepo_ API calls.

Thank you so much!!

PR #3519, opened 2025-01-27
support delete of multi-key lists without key

irfanHaslanded:
`sr_delete_item()` should be able to delete pushed operational data even without having all the keys specified.

In this case, it should simply delete all entries that would normally match.

### DIFF
--- a/src/ly_wrap.c
+++ b/src/ly_wrap.c
@@ -398,7 +398,7 @@ sr_lys_find_xpath(const struct ly_ctx *ctx, const char *xpath, uint32_t options,
         *valid = 1;
     }
 
-    if (lys_find_xpath(ctx, NULL, xpath, options, set)) {
+    if (lys_find_xpath(ctx, NULL, xpath, options, set) || !(*set)->count) {
         if (valid) {
             *valid = 0;
         } else {
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -3525,7 +3525,7 @@ sr_delete_item(sr_session_ctx_t *session, const char *path, const sr_edit_option
     const char *op;
     const struct lysc_node *snode;
     struct lyd_node *node;
-    struct ly_set *set;
+    struct ly_set *set = NULL;
     uint32_t temp_lo = 0;
     uint32_t i;
     int rc;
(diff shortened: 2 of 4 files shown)


src/sysrepo.c
```diff
@@ -3561,40 +3561,34 @@ sr_delete_item(sr_session_ctx_t *session, const char *path, const sr_edit_option
     }
 
     if (session->ds == SR_DS_OPERATIONAL) {
-        if ((err_info = sr_lys_find_path(session->conn->ly_ctx, path, NULL, &snode))) {
-            /* invalid path */
+        set = NULL;
```
michalvasko, line 3564: Seems redundant.

src/sysrepo.c
```diff
@@ -3561,40 +3561,34 @@ sr_delete_item(sr_session_ctx_t *session, const char *path, const sr_edit_option
...
-                    (err_info = sr_lyd_find_path(session->dt[session->ds].edit->tree, path, 0, &node))) {
-                goto cleanup;
-            }
-            if (node) {
-                sr_lyd_free_tree_safe(node, &session->dt[session->ds].edit->tree);
-            } else if (opts & SR_EDIT_STRICT) {
-                /* not found */
-                sr_errinfo_new(&err_info, SR_ERR_NOT_FOUND, "Node \"%s\" not found in session push oper data.", path);
+        if (!set || !set->count) {
+            /* due to a libyang bug, LYS_FIND_NO_MATCH_ERROR doesn't seem to work correctly */
```
michalvasko, line 3570: If there is a bug in *libyang*, it can be fixed. Please provide the use-case and I can fix it.

src/ly_wrap.c
```diff
@@ -398,7 +398,7 @@ sr_lys_find_xpath(const struct ly_ctx *ctx, const char *xpath, uint32_t options,
         *valid = 1;
     }
 
-    if (lys_find_xpath(ctx, NULL, xpath, options, set)) {
+    if (lys_find_xpath(ctx, NULL, xpath, options, set) || !(*set) || !(*set)->count) {
```
michalvasko, line 401: The rule should be that either `lys_find_xpath()` returns an error or `set` will be non-`NULL` (but may still be empty). So the `*set` check (and in your other changes) should be redundant.

michalvasko (changes requested):

src/sysrepo.c
irfanHaslanded: I think the commit 39a3459903a6252ef541259b70681a3aac601c20 I just added should reproduce the `libyang` bug and cause only a warning to be printed, in spite of 
```c
3565            if ((err_info = sr_lys_find_xpath(session->conn->ly_ctx, path, LYS_FIND_NO_MATCH_ERROR, NULL, &set))) {
(gdb)
767.909305 [3724081][140737347787520][WRN]: Schema node "list-enha" for parent "/mod:container" not found; in expr "/mod:container/list-enha" with context node "/".
3569            ly_set_free(set, NULL);
(gdb)
3570            set = NULL;

```

michalvasko: Right, in should be fixed in the latest *libyang*.

michalvasko:
I guess this is the simplest solution so fine, merging, thanks.

PR #3520, opened 2025-01-28
sr_module_oper_data_load remap lock if data exists

irfanHaslanded:
In `sr_module_oper_data_load()`, we can skip acquiring the ext remap lock unless push operational data exists.

Also, in `sr_shmext_print()` check for `shm_mod->oper_push_data_count`, rather than `shm_mod->oper_push_data`, which is the ext_shm offset. While it is okay to check the offset, it's not consistent with all the other similar checks.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -1263,6 +1263,7 @@ sr_module_oper_data_load(struct sr_mod_info_mod_s *mod, sr_conn_ctx_t *conn, uin
     uint32_t i, j, merge_opts, oper_push_count = 0;
     int last_sid = 0, dead_cid = 0;
 
+    /* oper_push_data_count is protected by operational DS module data locks */
     if (mod->shm_mod->oper_push_data_count) {
         /* EXT READ LOCK */
         if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_READ, 0, __func__))) {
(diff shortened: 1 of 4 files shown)


michalvasko:
The reason the lock is taken before reading `oper_push_data_count` is because that is the principle used, this variable is accessed from many threads and processes and is protected by this lock. It is true that in practice it is extremely unlikely it can cause any real issues but, theoretically, the value may be changed in the SHM but a process would use its locally cached value, without acquiring the lock. So I am against this change and since it is only a READ lock, it should not affect the performance in any significant way.

irfanHaslanded:
The *process-shared* shm module locks acquired by `sr_shmmod_lock` protect `shm_mod->oper_push_data_count`.

The connection ext remap lock being acquired here is a connection specific lock, and does not affect other processes, and cannot protect `oper_push_data_count` anyway.

As far as I understand, it only protects the Ext SHM from being remapped by all sessions of the current connection.

```C
    /* EXT READ LOCK */
    if ((err_info = sr_shmext_conn_remap_lock(conn, SR_LOCK_READ, 0, __func__))) {
        goto cleanup;
    }
```

Although it is only a READ lock, during system startup time, what I think is happening is that subscriptions and push oper data are being added to the `Ext SHM`, and hence growing continuously as processes set up subscriptions.

During this time, calling `sr_shmext_conn_remap_lock` means that there is a good chance of `shm_file_size != conn->ext_shm.size` being true, causing [shm_ext.c#L72](https://github.com/sysrepo/sysrepo/blob/40f2c8404514c47f6ad95305ae383de05973c040/src/shm_ext.c#L72 ) to [shm_ext.c#L111](https://github.com/sysrepo/sysrepo/blob/40f2c8404514c47f6ad95305ae383de05973c040/src/shm_ext.c#L111), to be executed more often than necessary.

Since we have around 150 yang modules, this tends to add up.

michalvasko:
Right, I have somehow forgotten about the SHM mod locks protecting this, so please add a corresponding comment above access to this count. It can then be merged.

PR #3529, opened 2025-02-04
oper ds plugins - skip load and create diff on del

irfanHaslanded:
None of the operational datastore plugins work with a diff but actual data to be stored.

So, it is unnecessary to load the data and create a diff for storage.

A common scenario would be when a process with large operational data restarts abruptly, and when it comes back up, it has to clean up the previous data. This can be very CPU intensive and is easily avoided.

### DIFF
--- a/src/plugins/ds_json.c
+++ b/src/plugins/ds_json.c
@@ -968,4 +968,5 @@ const struct srplg_ds_s srpds_json = {
     .access_check_cb = srpds_json_access_check,
     .last_modif_cb = srpds_json_last_modif,
     .data_version_cb = NULL,
+    .oper_store_require_diff = 0,
 };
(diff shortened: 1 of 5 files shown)


michalvasko:
I do not really like this change. Yes, it optimizes the single scenario of a process with lots of push oper data that are being deleted but none other. And the DB plugins should soon use this diff instead of just replacing all the data every time so is this change really justified? It does not seem to me that way.

irfanHaslanded:
Even if the datastore plugins eventually use the `ds_diff`, A diff is still not "required" for removal of all stored data of a session, and this alternative will still be much faster.

IMHO, This is necessary to allow the system to recover as quickly as possible, from a process failure.
It even affects performing a routine system upgrade where the downtime is increased needlessly, where all processes are competing for CPU during `sr_session_stop()`.

Moreover, `sr_shmmod_del_module_oper_data()` is called with module `WRITE` locks held, and causes system-wide lock contention during this time, so this is even more useful, to reduce impact on other processes.

michalvasko:
Fine, I suppose.

PR #3535, opened 2025-02-07
sessions cache operational push data

irfanHaslanded:
To edit a sessions operational push data there are 2 main steps
1. fetch previous oper push data
2. apply the changes by calling sr_apply_changes()

 ## Fetching operational push data to prepare a new edit
Most processes push operational data more than once in their lifetime.
For editing operational data, the previous push data is first fetched,
and a top level replace node is added.

This fetch involves the following operations:
1. reading the data from a datstore plugin (default JSON).
2. parsing the data into a libyang data tree `mod_info->data`

 ## sr_apply_oper_changes()
`sr_apply_oper_changes()` also fetches the previously stored oper push
data to generate the diff. This happens in 2 phases
1. `sr_modinfo_consolidate()` -> `sr_modinfo_data_load()`
2. `sr_modinfo_get_oper_data()` for diff generation.

After data is stored by calling `sr_modinfo_data_store()`, the `mod_info->data` is free'd by calling `lyd_free_siblings()` from `sr_modinfo_erase()`.

Both `lyd_parse()` and `lyd_free_siblings()` don't scale very well. This can be avoided by caching the previously pushed operational data.

This especially helps, when large operational data has been pushed already and a small edit needs to be performed (i.e. setting one leaf).

The `oper_push_data` cache must be flushed anytime context changes.

If additional memory footprint is a concern, the connection flag `SR_CONN_NO_OPER_PUSH_CACHE` can be passed to `sr_connect()` to disable caching for all sessions of the connection.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -1243,27 +1243,55 @@ sr_oper_data_merge_cb(struct lyd_node *trg_node, const struct lyd_node *src_node
 }
 
 /**
- * @brief Use (and spend) the oper push data cache if available for the module.
+ * @brief Check whether the session has pushed any operational data to this module.
+ *        If data exists, get it from cache if available.
  *
- * @param[in] sess Session to use.
- * @param[in] module_name Module name to get the cache for.
- * @return The oper push data cache of the module, if available.
+ * @param[in] sess Session to check.
+ * @param[in] mod_name module name to search.
+ * @param[in] dup Flag to duplicate data and not consume it.
+ * @param[out] has_data Flag whether session has push operational data for this module.
+ * @param[out] mod_data Cached module data if available.
+ * @return err_info if an error occurs during duplication, NULL otherwise.
  */
-static struct lyd_node *
-sr_module_oper_data_use_cache(sr_session_ctx_t *sess, const char *module_name)
+static sr_error_info_t *
+sr_modinfo_module_data_cache_get(sr_session_ctx_t *sess, const char *mod_name, int dup, int *has_data, struct lyd_node **mod_data)
 {
+    sr_error_info_t *err_info = NULL;
     uint32_t i;
-    struct lyd_node *data = NULL;
 
-    for (i = 0; i < sess->oper_push_mod_count; i++) {
-        if (!strcmp(module_name, sess->oper_push_mods[i].name)) {
-            data = sess->oper_push_mods[i].cache;
-            sess->oper_push_mods[i].cache = NULL;
+    *mod_data = NULL;
+    *has_data = 0;
+
+    for (i = 0; i < sess->oper_push_mod_count; ++i) {
+        if (!strcmp(sess->oper_push_mods[i].name, mod_name)) {
             break;
         }
     }
 
-    return data;
+    /* module not found */
+    if (i == sess->oper_push_mod_count) {
+        return NULL;
+    }
+
+    *has_data = sess->oper_push_mods[i].has_data;
+
+    /* no cache available for use */
+    if (!sess->oper_push_mods[i].cache) {
+        return NULL;
+    }
+
+    if (dup) {
+        /* duplicate the cached data for use */
+        if ((err_info = sr_lyd_dup(sess->oper_push_mods[i].cache, NULL, LYD_DUP_RECURSIVE | LYD_DUP_WITH_FLAGS, 1, mod_data))) {
+            return err_info;
+        }
+    } else {
+        /* consume the cache */
+        *mod_data = sess->oper_push_mods[i].cache;
+        sess->oper_push_mods[i].cache = NULL;
+    }
+
+    return NULL;
 }
 
 /**
@@ -1286,7 +1314,7 @@ sr_module_oper_data_load(struct sr_mod_info_mod_s *mod, sr_conn_ctx_t *conn, sr_
     const char *xpath;
     struct ly_set *set;
     uint32_t i, j, merge_opts, oper_push_count = 0;
-    int last_sid = 0, dead_cid = 0;
+    int last_sid = 0, dead_cid = 0, has_data = 0;
     uint32_t sid = sess ? sess->sid : 0;
 
     /* oper_push_data_count is protected by operational DS module data locks */
@@ -1342,7 +1370,9 @@ sr_module_oper_data_load(struct sr_mod_info_mod_s *mod, sr_conn_ctx_t *conn, sr_
         if (!last_sid || !mod_oper_data) {
             if (oper_push_dup[i].sid == sid) {
                 /* load push oper data from the session cache (consume it) */
-                mod_data = sr_module_oper_data_use_cache(sess, mod->ly_mod->name);
+                if ((err_info = sr_modinfo_module_data_cache_get(sess, mod->ly_mod->name, 0, &has_data, &mod_data))) {
+                    goto cleanup;
+                }
             }
             /* load push oper data for the session from the datastore plugin if cache was empty */
             if (!mod_data && (err_info = sr_module_file_data_append(mod->ly_mod, mod->ds_handle, SR_DS_OPERATIONAL,
@@ -2675,33 +2705,6 @@ cleanup:
 }
 
 /**
- * @brief Check whether the session has pushed any operational data to this module.
- *        If data exists, load it from cache if available.
- *
- * @param[in] sess Session to check.
- * @param[in] mod_name module name to search.
- * @param[out] mod_data Cached module data if available. Must be duplicated before use. Can be NULL if flushed or spent already.
- * @return 1 if there is data, else 0
- */
-static int
-sr_modinfo_module_data_cache_load(sr_session_ctx_t *sess, const char *mod_name, struct lyd_node **mod_data)
-{
-    uint32_t i;
-
-    for (i = 0; i < sess->oper_push_mod_count; ++i) {
-        if (!strcmp(sess->oper_push_mods[i].name, mod_name)) {
-            if (!sess->oper_push_mods[i].has_data) {
-                return 0;
-            }
-            *mod_data = sess->oper_push_mods[i].cache;
-            return 1;
-        }
-    }
-
-    return 0;
-}
-
-/**
  * @brief Load module data of a specific module.
  *
  * @param[in] mod_info Mod info to use.
@@ -2718,10 +2721,10 @@ sr_modinfo_module_data_load(struct sr_mod_info_s *mod_info, struct sr_mod_info_m
 {
     sr_error_info_t *err_info = NULL;
     sr_conn_ctx_t *conn = mod_info->conn;
-    struct lyd_node *mod_data = NULL, *dup = NULL;
+    struct lyd_node *mod_data = NULL;
     const char **xpaths;
     uint32_t xpath_count;
-    int modified;
+    int modified, has_data;
     char *orig_name = NULL;
     void *orig_data = NULL;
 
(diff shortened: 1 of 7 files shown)


michalvasko:
You are significantly raising the complexity with such optimizations and I will be the one stuck maintaining it. The way it was designed, you should be able to cache the push oper data yourself and just replace any previous data every time, is that not possible?

irfanHaslanded:
> You are significantly raising the complexity with such optimizations and I will be the one stuck maintaining it. The way it was designed, you should be able to cache the push oper data yourself and just replace any previous data every time, is that not possible?

I will think more on this. You are right about the additional complexity added from all these optimizations.

irfanHaslanded:
>  you should be able to cache the push oper data yourself and just replace any previous data every time

You are right about the first preparation, and it could be avoided by the application.
If the application caches the last pushed data directly, then it can avoid calling `sr_get_oper_changes()` to first prepare the changeset, by directly calling `sr_edit_batch()`.

But this still leaves the following calls from `sr_apply_oper_changes()`:
1. `sr_modinfo_get_oper_data()` 
2. `sr_modinfo_consolidate()` which calls `sr_modinfo_data_load()`
both of which need the datastore plugins to load data by parsing which is quite expensive.

These cannot be optimized by the application directly.

src/modinfo.c
```diff
@@ -1314,8 +1316,12 @@ sr_module_oper_data_load(struct sr_mod_info_mod_s *mod, sr_conn_ctx_t *conn, uin
             last_sid = 1;
         }
         if (!last_sid || !mod_oper_data) {
-            /* load push oper data for the session */
-            if ((err_info = sr_module_file_data_append(mod->ly_mod, mod->ds_handle, SR_DS_OPERATIONAL,
+            if ((oper_push_dup[i].sid == sid) && sess->oper_push_cache) {
+                /* load push oper data from the session cache (consume it) */
+                mod_data = sr_module_data_unlink(&sess->oper_push_cache, mod->ly_mod, 1);
+            }
+            /* load push oper data for the session from the datastore plugin if no data loaded above */
```
michalvasko, line 1323: I think the situation with no push oper data of this session would not work correctly. Could perhaps be fixed by moving `oper_push_cache` in the session to `oper_push_mods`. These can be separated by module unlike `running` data because they are never used as one large data tree but always only erged or something similar to the `operational` DS data.

src/modinfo.c
```diff
@@ -3030,6 +3036,16 @@ sr_modinfo_data_load(struct sr_mod_info_s *mod_info, int read_only, sr_session_c
         }
     }
 
+    /* use cached oper push data if available and usable */
+    if ((mod_info->ds == SR_DS_OPERATIONAL) && (mod_info->ds2 == SR_DS_OPERATIONAL) && sess &&
+            sess->oper_push_cache && !mod_info->data) {
+        /* need to duplicate and not consume it, as sr_apply_oper_changes() will consume oper_push_cache */
+        if ((err_info = sr_lyd_dup(sess->oper_push_cache, NULL, LYD_DUP_RECURSIVE | LYD_DUP_WITH_FLAGS, 1, &mod_info->data))) {
```
michalvasko, line 3043: If you really have use-cases with so much push oper data, this seems suboptimal because lots of data may be unnecessary and duplicated redundantly. With the change mentioned above, it should be better to duplicate the data per-module. You can then also check `state` of each module (if needed, not even sure) to handle empty data trees (instead of just checking `modinfo->data`).

src/sysrepo.c
```diff
@@ -4266,10 +4266,21 @@ sr_apply_oper_changes(struct sr_mod_info_s *mod_info, sr_session_ctx_t *session,
...
+    /* remove any previous oper_push_cache */
+    lyd_free_siblings(session->oper_push_cache);
+    session->oper_push_cache = NULL;
+
+    /* store mod_info->data as the new oper_push_cache only if there are no errors */
+    if (!err_info && !*err_info2 && !(session->conn->opts & SR_CONN_NO_OPER_PUSH_CACHE)) {
+        session->oper_push_cache = mod_info->data;
+        mod_info->data = NULL;
+    }
+
```
michalvasko, line 4278: Why is this code in the cleanup? I think freeing the cache could occur only on success (so can be moved before the label, on error the old cache can remain) and similarly updating the cache.

src/sysrepo_types.h
```diff
@@ -105,7 +105,8 @@ typedef enum {
     SR_CONN_DEFAULT = 0x0,              /**< No special behaviour. */
     SR_CONN_CACHE_RUNNING = 0x1,        /**< Always cache running datastore data which makes mainly repeated retrieval
                                              of data much faster. Affects all sessions created on this connection. */
-    SR_CONN_CTX_SET_PRIV_PARSED = 0x2   /**< Use LY_CTX_SET_PRIV_PARSED option for the connection libyang context. */
+    SR_CONN_CTX_SET_PRIV_PARSED = 0x2,  /**< Use LY_CTX_SET_PRIV_PARSED option for the connection libyang context. */
+    SR_CONN_NO_OPER_PUSH_CACHE = 0x4    /**< Disable operational push data cache for sessions of this connection. */
```
michalvasko, line 109: Does this flag have any use? If so, shouldn't it work similarly to `SR_CONN_CACHE_RUNNING` and enable the cache instead of disabling?

michalvasko (changes requested):

src/sysrepo_types.h
irfanHaslanded: I think it's better to just remove this flag.

src/modinfo.c
```diff
@@ -1242,16 +1242,41 @@ sr_oper_data_merge_cb(struct lyd_node *trg_node, const struct lyd_node *src_node
     return LY_SUCCESS;
 }
 
+/**
+ * @brief Use (and spend) the oper push data cache if available for the module.
+ *
+ * @param[in] sess Session to use.
+ * @param[in] module_name Module name to get the cache for.
+ * @return The oper push data cache of the module, if available.
```
michalvasko, line 1250: I would say you still cannot distinguish no cached data and empty cached data. Best to just use 2 out parameters `data` and `found`, for example. And use `void` return value since the function cannot fail.

src/modinfo.c
```diff
@@ -2696,13 +2731,20 @@ sr_modinfo_module_data_load(struct sr_mod_info_s *mod_info, struct sr_mod_info_m
...
-        if ((err_info = sr_module_file_data_append(mod->ly_mod, mod->ds_handle, mod_info->ds2,
+        if (mod_data) {
+            /* duplicate the cached data for use */
+            if ((err_info = sr_lyd_dup(mod_data, NULL, LYD_DUP_RECURSIVE | LYD_DUP_WITH_FLAGS, 1, &dup))) {
+                return err_info;
+            }
+            lyd_insert_sibling(mod_info->data, dup, &mod_info->data);
+        }
+        /* load module data from the ds plugin because it is not in the cache */
+        else if ((err_info = sr_module_file_data_append(mod->ly_mod, mod->ds_handle, mod_info->ds2,
```
michalvasko, line 2747: Please keep it on a single line, `} else if (...`. You can put the comment right below `if`.

michalvasko (changes requested):
As a minor optimization, the members of `oper_push_mods` can have a bit different meaning. `has_data` can probably be removed and one can assume that if a cache entry for a module exist, its `data` is valid. And if `!data`, it means `!has_data`, obviously. But the DS plugin callback does not need to be called.

src/modinfo.c
irfanHaslanded: I don't think this is necessary. 
```C
         if (!oper_push_dup[i].has_data) {
              /* no push oper data */
              continue;
          }
```
is checked in `sr_module_oper_data_load()` just before calling `sr_module_oper_data_use_cache()`.
So, there is no ambiguity.
If `sr_module_oper_data_use_cache()` returns `NULL`, we must always load the data from the DS plugin callback.

michalvasko: I see, a bit confusing. Can that actually happen, that there are no cached push oper data but there are some stored?

irfanHaslanded:
> And if `!data`, it means `!has_data`, obviously

To avoid any confusion, I have renamed `data` to `cache`.

While `cache` means `has_data`, and `!cache` does not mean `!has_data`.
This is because, cache is flushed after a context change, or the last `sr_apply_changes()` fails, after consuming the cache.
In both these cases, `has_data==1`, but `!cache`

src/modinfo.c
```diff
@@ -2645,21 +2675,26 @@ sr_modinfo_module_data_load_srmon(struct sr_mod_info_s *mod_info, struct sr_mod_
...
+ *        If data exists, load it from cache if available.
  *
  * @param[in] sess Session to check.
  * @param[in] mod_name module name to search.
+ * @param[out] mod_data Cached module data if available. Must be duplicated before use. Can be NULL if flushed or spent already.
  * @return 1 if there is data, else 0
  */
 static int
-sr_modinfo_sess_has_push_oper_data(sr_session_ctx_t *sess, const char *mod_name)
+sr_modinfo_module_data_cache_load(sr_session_ctx_t *sess, const char *mod_name, struct lyd_node **mod_data)
```
michalvasko, line 2687: I know it is a bit annoying but I would still prefer the `out` parameter to be `const` in this case for clarity. Also, seems the functionality is very similar to `sr_module_oper_data_use_cache` but because of this difference (that data needs to be duplicated) I suppose it can remain the way it is. But perhaps just using a flag to decide between using/duplicating the data would be best (and you can put the duplication in the function).

michalvasko (changes requested):
Just one last possible refactoring.

michalvasko (changes requested):
Sorry for not being able to find all the issues together, one last thing, hopefully.

src/modinfo.c
```diff
@@ -2696,18 +2734,22 @@ sr_modinfo_module_data_load(struct sr_mod_info_s *mod_info, struct sr_mod_info_m
...
+            return err_info;
         }
 
-        /* load module data */
-        if ((err_info = sr_module_file_data_append(mod->ly_mod, mod->ds_handle, mod_info->ds2,
-                conn->cid, sess->sid, mod->xpaths, mod->xpath_count, &mod_info->data))) {
-            return err_info;
+        if (has_data) {
+            if (mod_data) {
+                lyd_insert_sibling(mod_info->data, mod_data, &mod_info->data);
```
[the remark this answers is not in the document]
irfanHaslanded: Ack. Done

src/modinfo.c
```diff
@@ -3892,36 +3934,39 @@ sr_modinfo_push_oper_mod_learn_changes(sr_session_ctx_t *sess, const char *mod_n
...
+        SR_CHECK_MEM_GOTO(!sess->oper_push_mods[i].name, err_info, cleanup);
+        ++sess->oper_push_mod_count;
+    }
 
-    sess->oper_push_mods[i].name = strdup(mod_name);
-    SR_CHECK_MEM_GOTO(!sess->oper_push_mods[i].name, err_info, cleanup);
-    sess->oper_push_mods[i].has_data = has_data;
-    ++sess->oper_push_mod_count;
+    sess->oper_push_mods[i].has_data = !!data;
+    sess->oper_push_mods[i].cache = data;
```
michalvasko, line 3969: Is it guaranteed that `cache` is `NULL`? Well, obviously not when adding a new module because `realloc` is used but must the data be spent otherwise? If so, I would add a comment and assert at 3953.

src/modinfo.c
```diff
@@ -1242,16 +1242,69 @@ sr_oper_data_merge_cb(struct lyd_node *trg_node, const struct lyd_node *src_node
...
+ *
+ * @param[in] sess Session to check.
+ * @param[in] mod_name module name to search.
+ * @param[in] dup Flag to duplicate data and not consume it.
+ * @param[out] has_data Flag whether session has push operational data for this module.
+ * @param[out] mod_data Cached module data if available.
+ * @return err_info if an error occurs during duplication, NULL otherwise.
+ */
+static sr_error_info_t *
+sr_modinfo_module_data_cache_load(sr_session_ctx_t *sess, const char *mod_name, int dup, int *has_data, struct lyd_node **mod_data)
```
michalvasko, line 1257: Just a minor improvement, I would prefer some other word than `load`, which is generally used when actually parsing data using DS plugins. Maybe just `get` would be clearer.

michalvasko (changes requested):
Still found something fairly serious but I am pretty sure when we solve this, it can be merged.

src/modinfo.c
irfanHaslanded: Yes, it should be guaranteed to be `NULL`, after the first `sr_modinfo_get_oper_data()` call in `sr_apply_oper_changes()`.

```C
    /* get the operational DS data with the old push oper data */
    if ((err_info = sr_modinfo_get_oper_data(mod_info, session, NULL))) {
        goto cleanup;
    }
```

Since operational datastore does not support factory reset RPC, or `sr_replace_config()`, `sr_modinfo_data_store()` should never be called for a module with its cache still present.

michalvasko (approved):

PR #3547, opened 2025-02-13
oper edit BUGFIX load push oper data only once

irfanHaslanded:
For edit operations, using the editing API `sr_set_item_str()`, `sr_delete_item()`, `sr_discard_items()` etc. on the operational datastore, session pushed data needs to be loaded the first time an edit is being created.
The same edit is used for all subsequent set/delete calls.

However, tracking the presence of an edit with
`!session->dt[session->ds].edit` is not an accurate check for this.

For example, if all pushed operational data are deleted as a result of `sr_delete_item()` calls, the edit can become `NULL`.

Subsequent edit operations will mistake this for a clean session with no changes. Calling `sr_apply_changes()` at this point, will have no effect, whereas it should have been functionally equal to calling `sr_discard_oper_changes()`.
If new set operations are called, they will be placed on top of the current stored push data, with the staged `delete` lost.

To avoid any confusion, a new flag `oper_edit_prepared` is created in the session context, to ensure that oper_changes are fetched only once per staged edit.

`oper_edit_prepared` is set to zero after `sr_discard_changes_xpath()` or `sr_apply_changes()` complete successfully, as the staged changes are applied, and there is no longer a prepared edit.

- test oper delete all push data using `sr_delete_item()`

### DIFF
--- a/src/common_types.h
+++ b/src/common_types.h
@@ -206,6 +206,7 @@ struct sr_session_ctx_s {
         sr_data_t *edit;            /**< Prepared edit data tree. */
         struct lyd_node *diff;      /**< Diff data tree, used for module change iterator. */
     } dt[SR_DS_COUNT];              /**< Session-exclusive prepared changes. */
+    int oper_edit_fetched;         /**< If current operational ds edit has been fetched, even if session->dt[SR_DS_OPERATIONAL]->edit == NULL. */
 
     struct sr_sess_notif_buf {
         int thread_running;         /**< Flag whether the notification buffering thread of this session is running. */
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -3477,11 +3477,16 @@ sr_set_item_str(sr_session_ctx_t *session, const char *path, const char *value, 
         }
     }
 
-    if (!session->dt[session->ds].edit && (session->ds == SR_DS_OPERATIONAL)) {
+    if (!session->oper_edit_fetched && (session->ds == SR_DS_OPERATIONAL)) {
+        assert(!session->dt[session->ds].edit);
+
         /* prepare the current stored oper data to be modified */
         if ((rc = sr_get_oper_changes(session, NULL, &session->dt[session->ds].edit))) {
             goto cleanup;
         }
+
+        /* remember that oper changes have been fetched */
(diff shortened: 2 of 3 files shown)


irfanHaslanded:
> Actually, instead of just a flag, there could be an array of modules that have their oper data fetched. It can be used for optimizing fetching only the data of module that are going to be changed and also on `sr_apply_changes()`, to ignore processing unchanged modules.

Since sessions are inexpensive, can we live with recommending users create one session per module especially if there is any scale involved?

I am also keeping in mind that we really don't want to add complexity, if it can be avoided.

michalvasko:
In this case I do not think having a flag for all the modules or an array of selected modules makes much of a difference so like I said, I have no problem with such a change.

irfanHaslanded:
I couldn't figure out how to go about the array implementation as there are many things to consider, or maybe I am mistaking something.

1. for each edit API call, get the owner module from the path/xpath, and fetch its data and add as a sibling node if not already fetched.
2. for delete and discard, there could be multiple modules.
3. check in array if the current module list is already fetched, if not realloc that array after fetching those modules individually, by repeatedly calling `sr_get_oper_changes()` for each module.
4. `sr_discard_changes_xpath()` check which modules are removed, and unset the flags in array.

Can we merge this as a first step to bugfix `sr_delete_item()`, and then take up the array optimization if still needed?

michalvasko (approved):

michalvasko:
> for each edit API call, get the owner module from the path/xpath, and fetch its data and add as a sibling node if not already fetched.

Yes, exactly.

> for delete and discard, there could be multiple modules.

Well, yes, since both now support XPath.

> check in array if the current module list is already fetched, if not realloc that array after fetching those modules individually, by repeatedly calling sr_get_oper_changes() for each module.

Yes, so that the data of all the relevant modules are fetched.

> sr_discard_changes_xpath() check which modules are removed, and unset the flags in array.

I guess, if this is supposed to be fully supported, you should check that all the data of a module were removed and then remove that module from the array.

> Can we merge this as a first step to bugfix sr_delete_item(), and then take up the array optimization if still needed?

Fine, you asked for the module data fetching to be optimized and this seems like the simplest way of doing so.

PR #3550, opened 2025-02-17
oper ds CLARIFY ordering of session push data

irfanHaslanded:
### operational changes order
For each module, sessions can set the order explicitly by calling
`sr_set_oper_changes_order()`, or the next higher order is assigned to
it when it first pushes data to that module.

If a lower order process restarts, it will push data with
a higher order, unless it explicitly sets it's order.

 ### discard-items
Sessions can also push opaque "discard-items" nodes in addition to the
regular data nodes for the module.

These "discard-items" nodes can be used to specify XPaths that will not
be visible in the final observable "view" of the operational datastore.
Discard is *only* performed on *lower order* operational data.
They do not affect push data of sessions of higher order.

 ### `sr_set_oper_changes_order()`
1. Should be called by all push oper sessions of the module, if any
   sessions use "discard-items" matching that module.
2. As this function does not notify changes to subscribers, Order cannot
   be changed if the session has push data currently.
3. A session cannot reliably set order for all modules. This function
   must be called explicitly for each module before pushing operataionl
   data.

 ### Conclusion
For "discard-items" to work correctly on push operational data, all
the sessions should explicitly call `sr_set_oper_changes_order()` before
pushing any operational data for that module.


irfanHaslanded:
Unclear why the ABI check has failed

michalvasko:
I am thinking about removing the ABI check because it started reporting changes very often, even when obviously no ABI changes are made, no idea what is happening there. But regarding the PR, when we discussed the change it seemed like a simple change of forbidding to change the push order of sessions that have some push oper data. There are way too many changes for my liking and I do not really understand why.

irfanHaslanded:
> I am thinking about removing the ABI check because it started reporting changes very often, even when obviously no ABI changes are made, no idea what is happening there.

Sounds reasonable.

> But regarding the PR, when we discussed the change it seemed like a simple change of forbidding to change the push order of sessions that have some push oper data. There are way too many changes for my liking and I do not really understand why.

These are the changes I made
1. `sr_shmmod_session_oper_order()` is removed as `sr_shmext_oper_push_get()` and `sr_shmext_oper_push_update()` can be called directly from the set/get order API. `sr_shmmod_session_oper_order()` for get contained dead code when `ly_mod == NULL` anyway.
2. split up `sr_modinfo_push_oper_mod_update_sess()` which was doing 2 different tasks
     a. `sr_modinfo_push_oper_mod_learn_changes()`
     b. `sr_modinfo_push_oper_mod_update_sess()`
3. updated documentation for `sr_discard_items()` and `sr_set_oper_changes_order()`
4. added tests.

I can undo the parts you think are problematic.

src/sysrepo.h
```diff
@@ -972,12 +972,16 @@ int sr_oper_delete_item_str(sr_session_ctx_t *session, const char *path, const c
...
+ * the other push oper data of this session. Usable only for ::SR_DS_OPERATIONAL datastore. These changes are applied
  * only after calling ::sr_apply_changes().
  *
  * Creates an opaque node `discard-items` in the `sysrepo` YANG module namespace with @p xpath used as the value.
  * Such a node can be a part of the edit in ::sr_edit_batch() and will discard nodes like this function does.
  *
+ * This affects only operational data of lower order. For correct behaviour, `sr_set_oper_changes_order()` must be called
+ * from all sessions that push operational data to the module to explicitly set the order.
+ * `sr_discard_items` will not give consistent results with automatically generated order.
+ *
```
michalvasko, line 984: I disagree with most of this, it should be formulated as a recommendation, not a necessity. You will get "correct" behavior even without explicitly setting the order.

src/sysrepo.h
```diff
@@ -954,7 +954,7 @@ int sr_set_item_str(sr_session_ctx_t *session, const char *path, const char *val
  * changes merged into the list, use ::SR_EDIT_ISOLATE in such a case.
  *
  * For ::SR_DS_OPERATIONAL, this function deletes the selected nodes from the session push oper data. To delete the nodes
- * from the final operational datastore, use ::sr_discard_items() instead. Only ::SR_EDIT_STRICT option is allowed
+ * from the final operational datastore, use ::sr_discard_items() in addition. Only ::SR_EDIT_STRICT option is allowed
```
michalvasko, line 957: Do not understand this change.

src/sysrepo.c
```diff
@@ -4807,23 +4807,39 @@ API int
...
 {
     sr_error_info_t *err_info = NULL;
-    const struct lys_module *ly_mod = NULL;
+    sr_mod_t *shm_mod;
+    uint32_t i;
 
-    SR_CHECK_ARG_APIRET(!session || !order, NULL, err_info);
+    SR_CHECK_ARG_APIRET(!session || !module_name || !order, NULL, err_info);
+
+    if (session->oper_push_mod_count) {
```
michalvasko, line 4815: Redundant condition.

src/modinfo.c
```diff
@@ -3901,7 +3922,9 @@ sr_modinfo_push_oper_mod_update_sess(sr_session_ctx_t *sess, const char *mod_nam
     SR_CHECK_MEM_GOTO(!sess->oper_push_mods[i].name, err_info, cleanup);
     sess->oper_push_mods[i].has_data = has_data;
     ++sess->oper_push_mod_count;
+
 cleanup:
+
```
michalvasko, line 3927: Redundant.

src/sysrepo.h
```diff
@@ -1156,9 +1160,13 @@ int sr_get_oper_changes(sr_session_ctx_t *session, const char *module_name, sr_d
  * for the module.
  *
  * By default the next highest order (lowest priority) is generated for any new session push oper data of a module.
+ * It is not supported to set the order:
+ * 1. if the session already has push operational data for the specified module.
+ * 2. for all modules if the session has push operational data for any module.
```
michalvasko, line 1165: Do not understand this, you cannot change the order of all the modules anymore no matter what, based on the `module_name` description (which is fine).

michalvasko:
Such changes, when whole functions are changed/moved around are very difficult to understand in the diff so please at least separate them at least in commits next time, better yet in separate PRs so they can be merged iteratively.

src/modinfo.c
```diff
@@ -3978,10 +4001,8 @@ sr_modinfo_data_store(struct sr_mod_info_s *mod_info, sr_session_ctx_t *session,
                         goto cleanup;
                     }
                 } else {
-                    /* stored oper data, now cache the modified module in the session and determine if ext `change` or `create` is needed */
-                    if ((err_info = sr_modinfo_push_oper_mod_update_sess(session, mod->ly_mod->name, !!mod_data, &create, &change))) {
-                        goto cleanup;
-                    }
+                    /* stored oper data, determine if ext `change` or `create` is needed from the session */
```
michalvasko, line 4004: Took me a few seconds to understand, should perhaps be `/s/from/for`?

michalvasko (changes requested):

src/sysrepo.h
irfanHaslanded: Done

michalvasko (approved):

src/modinfo.c
```diff
@@ -3860,38 +3860,59 @@ sr_modinfo_generate_config_change_notif(struct sr_mod_info_s *mod_info, sr_sessi
...
  * @param[in] mod_name Module name.
  * @param[in] has_data Flag to store.
- * @param[out] created Learn whether oper push data for module was just created (first time) .
- * @param[out] changed_has_data Learn whether has_data for module changed from the previous value.
- * @return err_info, NULL on success.
+ * @return err_info, NULL on success, only fails if out of memory.
  */
-static sr_error_info_t *
-sr_modinfo_push_oper_mod_update_sess(sr_session_ctx_t *sess, const char *mod_name, int has_data, int *created, int *changed_has_data)
+sr_error_info_t *
```
michalvasko, line 3902: It seems this function can remain `static`.

michalvasko (changes requested):
Found probably just one last issue.

src/modinfo.c
irfanHaslanded: Ah, yes. let me fix that

PR #3554, opened 2025-02-26
Make MongoDB and Redis datastore plugins optional

arkamar:
This commit introduces the `ENABLE_DS_MONGO` and `ENABLE_DS_REDIS` options, allowing users to explicitly enable or disable the MongoDB and Redis datastore plugins. Without these options, the plugins had automagic properties, being included whenever their dependencies were found, which might not be intentional.

This change is especially useful for downstream packagers, as it allows them to control the dependencies explicitly and avoid unintended linkage. It reduces unnecessary dependencies in environments where these plugins are not needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
             build-type: "Debug",
             dep-build-type: "Debug",
             cc: "gcc",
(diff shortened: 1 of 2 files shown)


michalvasko (changes requested):
No problem and I think I would even go further by making the default values `OFF`. If the user is not using these plugins and they are enabled, it causes issues so I think having to explicitly enable them makes sense. Also, please change the target branch to `devel`, we do not accept PRs into `master`.

arkamar:
Yeah, I agree, `OFF` is better but I wasn't sure what is preferred default, so I made it in a way to preserve original behavior. Anyway, I changed it to `OFF`, rebased to `devel` and changed the target branch.

michalvasko (approved):

michalvasko:
Right, the CI needs to be updated, can you do it?

arkamar:
Thanks!

PR #3557, opened 2025-02-26
modinfo OPTIMIZE cond fetch implicit oper data

irfanHaslanded:
`sr_module_oper_data_update()` calls `sr_lyd_new_implicit_module()` to add the implicit NP containers for the module. The main reason to do this is to learn all the necessary oper get callbacks to invoke.

When `SR_OPER_NO_SUBS` is used, no callbacks will be invoked, so it is not necessary to call `sr_lyd_new_implicit_module()`.


irfanHaslanded:
I am not entirely sure, if the reasoning is correct.

michalvasko:
I think the reasoning is correct, that is the main reason, but it does not mean it cannot cause some issues (I am not able to tell either). I will think about it more but I am also wondering what use-case you have that spends significant time in this function (and some measurements would be nice, too).

irfanHaslanded:
```sh
    96.45%     0.00%  sysrepocfg  sysrepocfg                                                   [.] main
            |
            ---main
               |
               |--93.61%--op_export
               |          |
               |          |--79.16%--sr_get_data
               |          |          |
               |          |          |--68.14%--sr_modinfo_consolidate
               |          |          |          |
               |          |          |           --68.13%--sr_modinfo_data_load
               |          |          |                     |
               |          |          |                      --68.09%--sr_modinfo_module_data_load
               |          |          |                                |
               |          |          |                                 --67.86%--sr_module_oper_data_update
               |          |          |                                           |
               |          |          |                                           |--46.41%--sr_lyd_new_implicit_module
               |          |          |                                           |          lyd_new_implicit_module
               |          |          |                                           |          |
               |          |          |                                           |           --46.21%--lyd_new_implicit_tree
               |          |          |                                           |                     |
               |          |          |                                           |                     |--28.73%--lyd_new_implicit
               |          |          |                                           |                     |          |
               |          |          |                                           |                     |          |--13.10%--lyd_find_sibling_val
               |          |          |                                           |                     |          |          |
               |          |          |                                           |                     |          |           --11.54%--lyd_find_sibling_schema
               |          |          |                                           |                     |          |                     |
               |          |          |                                           |                     |          |                     |--6.19%--lyht_hash_multi
:
```

irfanHaslanded:
I have pasted the `perf report` output above in one such case, where there are no subscr to the modules in question, but significant amount of time is being spent.

irfanHaslanded:
It's likely that `sr_module_change_subscribe_enable()` will not work correctly when it calls
```C
    if ((err_info = sr_modinfo_consolidate(mod_info, SR_LOCK_READ, SR_MI_PERM_NO, session, 0, 0, SR_OPER_NO_SUBS))) {
        goto cleanup;
    }

    /* start with any existing config NP containers */
    if ((err_info = sr_lyd_dup_module_np_cont(mod_info->data, ly_mod, 0, &enabled_data))) {
        goto cleanup;
    }

```

michalvasko:
Thanks for the `perf` output, I was wondering whether there are not some other optimization possibilities but it seems not.

And `sr_module_change_subscribe_enable()` is a good find but could probably be fixed easily, just put `sr_lyd_new_implicit_module()` instead of `sr_lyd_dup_module_np_cont()` and it should even be faster as the containers do not have to first be created and then copied, but directly created in the enabled data.

src/modinfo.c
```diff
@@ -1520,6 +1553,11 @@ sr_module_oper_data_update(struct sr_mod_info_mod_s *mod, const char *orig_name,
             return err_info;
         }
 
+        /* check whether we need any implicit data, or it can be skippped */
+        if ((err_info = sr_module_oper_data_implicit_skip(mod, conn, get_oper_opts, &skip_implicit_data)) || skip_implicit_data) {
+            return err_info;
+        }
+
         /* add any missing NP containers in the data */
         if ((err_info = sr_lyd_new_implicit_module(data, mod->ly_mod, LYD_IMPLICIT_NO_DEFAULTS, NULL))) {
```
irfanHaslanded, line 1562: switching this to `sr_lyd_new_implicit_module()` caused `test_oper_pull` to fail, and I couldn't make sense of why.

src/sysrepo.c
```diff
@@ -5840,7 +5840,7 @@ sr_module_change_subscribe_enable(sr_session_ctx_t *session, struct sr_mod_info_
     if ((err_info = sr_modinfo_add(ly_mod, NULL, 0, 0, mod_info))) {
         goto cleanup;
     }
-    if ((err_info = sr_modinfo_consolidate(mod_info, SR_LOCK_READ, SR_MI_PERM_NO, session, 0, 0, SR_OPER_NO_SUBS))) {
+    if ((err_info = sr_modinfo_consolidate(mod_info, SR_LOCK_READ, SR_MI_PERM_NO, session, 0, 0, SR_OPER_NO_SUBS | SR_OPER_WITH_NP_CONT))) {
         goto cleanup;
     }
```
irfanHaslanded, line 5846: switching below to sr_lyd_new_implicit_module() caused test_oper_pull to fail, and I couldn't make sense of why.

michalvasko:
To be honest, I do not like the current changes at all, the function checking NP containers depends on too many conditions and it will be difficult to keep track of that and avoid unexpected behavior. But I do like the new flag, which I think I would make decide whether to return the NP containers or not. Meaning, by default the containers would always be there, which was another issue I realized now. Not returning them suddenly is a behavior change and one that goes against the principles we decided to stick to. While empty NP containers may not have much use, in *libyang* and so in *sysrepo*, they should always exist in configuration data because they behave as default values. Of course, if the user requests not to return them, then it is fine.

irfanHaslanded:
It's probably better and easier if you implement this, when you have the time.
I can close this draft PR.

michalvasko:
Seemed quite straightforward, you can check the changes.

PR #3560, opened 2025-03-05
Fix setting replay_start_time in srsn_subscribe

peckato1:
The original code always zeroed variable `replay_start` and then, the assignment to `replay_start` was never done because the condition was always false. This resulted in  `srsn_subscribe` always returning 0 in `replay_start_time`.

The second commit adds a test to check that `replay_start_time` is set to 0 when there are notifications older than `start_time` of the subscription.


michalvasko (approved):

michalvasko:
Thanks but `uncrustify` does not like some code formatting, please fix, it can then be merged.

michalvasko (approved):

PR #3568, opened 2025-03-17
ds plugins json use FILE writer instead of fd

irfanHaslanded:
A large number of write() syscalls are made while using the raw fd directly while printing the datastore.

Using FILE writers instead, which use buffering can have significant performance improvement.


michalvasko (approved):

michalvasko:
No problem, thanks.

PR #3575, opened 2025-03-25
run cache BUGFIX avoid flush when no CID

irfanHaslanded:
`sr_connect()` can fail to open the main_shm, if there are no adequate
permissions, or if there was a SHM version mismatch.

In such cases, a CID has not been assigned yet. So
`sr_conn_run_cache_flush(conn)` cannot be called as it calls
`sr_rwlock(&conn->run_cache_lock)` which fails the assertion:

`assert(mode && timeout_abs && cid);`

### DIFF
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -165,7 +165,12 @@ sr_conn_free(sr_conn_ctx_t *conn)
 
     /* unlocked data destroy */
     lyd_free_siblings(conn->ly_ext_data);
-    sr_conn_run_cache_flush(conn);
+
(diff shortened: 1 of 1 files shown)


michalvasko (changes requested):
Seems fine but I would prefer checking the CID and handling this situation specially. If there is a lock protecting something, a sane expectation is that it will be held when that something is being modified.

irfanHaslanded:
Sure, I changed it to avoid calling `sr_conn_run_cache_flush()` unless CID exists.

michalvasko (approved):
Neat now, thanks.

PR #3579, opened 2025-04-01
sysrepo UPDATE add ietf-yang-library-augmentedby support

jeremie6wind:
Parse the ietf yang module and use it as internal module. The modules and their augmentations are visible under the "/ietf-yang-library:yang-library xpath".

Link: https://datatracker.ietf.org/doc/html/draft-ietf-netconf-yang-library-augmentedby


michalvasko:
Thanks, but there is a similar problem as in *libyang*, the module RFC should be published before we implement it.

PR #3581, opened 2025-04-01
ds JSON OPTIMIZE store with shortest WRITE lock

irfanHaslanded:
modules are WRITE locked before calling `sr_modinfo_data_store()`. This is because the datastore is being replaced with new data, and readers must be blocked until the store operation is complete.

 ### JSON datastore plugin:
Storing large module data into JSON can be expensive. Normally, this requires WRITE module locks to be held as `srpds_json_store()` overwrites the current JSON module data file with new data.

This can be done differently in two steps:
1. Creating a temporary json data file
2. Replacing the current json data file with the temporary file.

Step 1. is a heavy operation, and does not require WRITE module locks. Step 2. is a extremely fast `rename()`, and requires WRITE module locks.

Doing the two-step store, allows reads to almost never have to block.

Similar logic could be implemented for other plugins if feasible.

### DIFF
--- a/src/plugins_datastore.h
+++ b/src/plugins_datastore.h
@@ -40,7 +40,7 @@ extern "C" {
 /**
  * @brief Datastore plugin API version
  */
-#define SRPLG_DS_API_VERSION 12
+#define SRPLG_DS_API_VERSION 13
 
 /**
  * @brief Setup datastore of a newly installed module.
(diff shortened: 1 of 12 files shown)


michalvasko:
Okay, I get the idea, not sure about the chosen design but that can be improved. But first, is this really desired, for readers to still be reading old data when new data are being written? I am not so sure about that and concurrency optimizations may have their limits.

irfanHaslanded:
> But first, is this really desired, for readers to still be reading old data when new data are being written?

This is my reasoning for the change:
1. Even before this change, readers were reading old data, just before new data would be written until `sr_shmmod_modinfo_rdlock_upgrade()`. This only shortens that window.
2. Readers which truly care about getting "old data" should really be watching for changes, rather than reading, as it can always give old data.

The real benefit of it is, IMHO, it avoids reader timeouts, as the WRITE lock is held for a very short window.

src/plugins_datastore.h
```diff
@@ -111,7 +111,7 @@ typedef sr_error_info_t *(*srds_conn_init)(sr_conn_ctx_t *conn, void **plg_data)
 typedef void (*srds_conn_destroy)(sr_conn_ctx_t *conn, void *plg_data);
 
 /**
- * @brief Store data for a module. Either a diff can be applied manually (if available) or full new data tree stored.
+ * @brief Store or prepare to store data for a module. Either a diff can be applied manually (if available) or full new data tree stored.
```
michalvasko, line 114: Please add a new `typedef` so that every callback has its own, just like before. And mention the main difference in the docs, in the held locks. Also, in some use-cases the WRITE lock is held already for `store_prepare()`, so either fix that or mention it clearly (so that the callback never relies on it).

michalvasko (approved):

PR #3584, opened 2025-04-04
fix: Teardown sequence must be inverse of init sequence

rbu9fe:
The sysrepo-plugind daemon loads the found plugins in a well-defined order that can be controlled by writing the list `/sysrepo-plugind:sysrepo-plugind/plugin-order/plugin`. That's a nice feature to help control dependencies between plugins. Currently the teardown sequence is the same as the initialization sequence, however, it should be _inverse_ to respect those dependencies.

### DIFF
--- a/src/executables/sysrepo-plugind.c
+++ b/src/executables/sysrepo-plugind.c
@@ -628,16 +628,17 @@ main(int argc, char **argv)
     rc = EXIT_SUCCESS;
 
 cleanup:
-    for (i = plugin_count - 1; i >= 0; --i) {
+    for (i = plugin_count; i > 0; --i) {
+        const int idx = i - 1;
         /* plugin cleanup */
-        if (plugins[i].initialized) {
-            plugins[i].cleanup_cb(sess, plugins[i].private_data);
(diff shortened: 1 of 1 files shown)


src/executables/sysrepo-plugind.c
```diff
@@ -628,7 +628,7 @@ main(int argc, char **argv)
     rc = EXIT_SUCCESS;
 
 cleanup:
-    for (i = 0; i < plugin_count; ++i) {
+    for (i = plugin_count - 1; i >= 0; --i) {
```
michalvasko, line 631: `plugin_count` may be `0`.

michalvasko (changes requested):
No problem with the changes, just handle all the cases.

michalvasko (approved):

michalvasko (changes requested):
`uncrustify` does not seem to like your changes, probably an empty line expected after the var declaration. We are not big fans of declarations at other places than the function beginning, anyway.

rbu9fe:
Next try...

PR #3586, opened 2025-04-08
ntf_json UPDATE return if notification file space is insufficient

NLako-6WIND:
When there is not enough space at a particular time stamp for registering the notification. We'll try to re-create the same file, resulting in a 'File exists' error.
If the file doesn't have enough space, check timestamps, and if they are within the exact same second, clean up and set a proper error message.


src/plugins/ntf_json.c
```diff
@@ -456,6 +456,10 @@ srpntf_json_store(const struct lys_module *mod, const struct lyd_node *notif, co
 
             /* we are done */
             goto cleanup;
+        } else if (notif_ts->tv_sec == from_ts && notif_ts->tv_sec == to_ts) {
+            srplg_log_errinfo(&err_info, srpntf_name, NULL, SR_ERR_INTERNAL,
+                "Not enough file space.");
```
michalvasko, line 461: Please explain the exact problem in the message. Such as `Notification file <from>-<to> exceeded the maximum size <max-size> KB.`.

src/plugins/ntf_json.c
```diff
@@ -456,6 +456,10 @@ srpntf_json_store(const struct lys_module *mod, const struct lyd_node *notif, co
 
             /* we are done */
             goto cleanup;
+        } else if (notif_ts->tv_sec == from_ts && notif_ts->tv_sec == to_ts) {
+            srplg_log_errinfo(&err_info, srpntf_name, NULL, SR_ERR_INTERNAL,
```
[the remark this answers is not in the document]
NLako-6WIND: Thanks for the review and comments:)

On this one -> I think SR_ERR_SYS error can be misleading. While it currently arises from a failed open() call, setting this one to SR_ERR_SYS doesn't represent a true system call failure. The error is provoked by the current limitation in sysrepo's notification file management, which could be addressed by increasing the MAX_SIZE for notification files or other methods... This is a separate issue, though, and it suggests a potential limitation in handling notifications for very large configuration changes (over 1MB).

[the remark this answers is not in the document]
michalvasko: Fair enough, then use `SR_ERR_UNSUPPORTED` or `SR_ERR_OPERATION_FAILED`. `SR_ERR_INTERNAL` should be used only for errors that should never happen (there is no known scenario), in this case there is.

src/plugins/ntf_json.c
```diff
@@ -456,6 +456,11 @@ srpntf_json_store(const struct lys_module *mod, const struct lyd_node *notif, co
 
             /* we are done */
             goto cleanup;
+        } else if ((notif_ts->tv_sec == from_ts) && (notif_ts->tv_sec == to_ts)) {
+            srplg_log_errinfo(&err_info, srpntf_name, NULL, SR_ERR_OPERATION_FAILED,
+                "Notification file %d-%d exceeded the maximum size of %d KB.",
```
michalvasko, line 461: Please print `time_t` similarly to [other such prints](https://github.com/sysrepo/sysrepo/blob/master/src/plugins/common_json.c#L737) to avoid warnings on various platforms.

michalvasko (changes requested):

michalvasko (approved):

NLako-6WIND:
Thanks, @michalvasko 

I wanted to ask if I should open an separate `#issue` for the Notification storage when the file is 'full'. This is an issues only when the file gets 'full' within 1 second timeout.

src/plugins/ntf_json.c
```diff
@@ -456,6 +456,11 @@ srpntf_json_store(const struct lys_module *mod, const struct lyd_node *notif, co
 
             /* we are done */
             goto cleanup;
+        } else if ((notif_ts->tv_sec == from_ts) && (notif_ts->tv_sec == to_ts)) {
+            srplg_log_errinfo(&err_info, srpntf_name, NULL, SR_ERR_OPERATION_FAILED,
+                "Notification file %" PRId64 "-%" PRId64 " exceeded the maximum size of %d KB.",
+                (int64_t)from_ts, (int64_t)to_ts, SRPJSON_NOTIF_FILE_MAX_SIZE);
```
michalvasko, line 462: We use double indent when continuing a command.

michalvasko:
I suppose you can, not sure what you mean.

src/plugins/ntf_json.c
NLako-6WIND: Maybe there's a way I could launch the style checker on my side? :)

michalvasko: Yes, if you have `uncrustify` installed, just run `make format`.

PR #3587, opened 2025-04-09
ds plugins OPTIMIZE pass req xpaths to load_cb

irfanHaslanded:
load_cb can be xpath aware for operational datastore, and fetch only the required data for the datastore plugins that support it.

Implemented support:
1. redis DS plugin supports this for all datastores.
2. mongo ds plugin supports this except for operational datastore.
  - ietf-origin metadata must be explicitly retrieved for each path.
  - this is not implemented yet, so all data will be retrieved.
3. json ds plugin - unimplemented. ignores passed xpaths.

This has the benefit of optimally loading only the requested data, and not constructing the entire module datatree, if only a small part is being read.

Further, it can reduce the amount of time modules remain read locked, when they have large amount of data, and occasional readers of small data.

### DIFF
--- a/src/plugins/ds_redis.c
+++ b/src/plugins/ds_redis.c
@@ -2641,12 +2641,13 @@ cleanup:
  * @param[in] ctx Libyang context.
  * @param[in] xpaths Array of XPaths.
  * @param[in] xpath_cnt XPath count.
+ * @param[in] oper_ds Flag if the filter is for loading operational data and special handling is needed.
  * @param[out] xpath_filter Final query filter.
  * @return NULL on success;
  * @return Sysrepo error info on error.
  */
 static sr_error_info_t *
-srpds_process_load_paths(struct ly_ctx *ctx, const char **xpaths, uint32_t xpath_cnt, char **xpath_filter)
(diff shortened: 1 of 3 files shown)


src/plugins/ds_redis.c
```diff
@@ -2641,12 +2641,13 @@ srpds_load_and_copy_data(redisContext *ctx, const char *mod_ns_src, const char *
  * @param[in] ctx Libyang context.
  * @param[in] xpaths Array of XPaths.
  * @param[in] xpath_cnt XPath count.
+ * @param[in] discard_items If sysrepo discard-items nodes should be retrieved as well.
```
michalvasko, line 2644: Please just add `(always for operational DS)`. I was a bit confused initially.

michalvasko (approved):
Okay, the changes seem too small for such a change but I assume you have tested it. So just a minor improvement and it can be merged.

src/plugins/ds_redis.c
irfanHaslanded: Changed it so that `srpds_process_load_paths()` functions in both mongo and redis DS plugins are similar.

PR #3589, opened 2025-04-10
mod_info run_cache reduce timeouts

irfanHaslanded:
### mod_info run_cache reduce timeouts

    running ds cache lock needs to be held for the entire operation, only
    when reading the running datastore.

    For all other operations, the necessary data can be quickly duplicated
    and the cache unlocked.

    Without this change, there are a couple of problems:

    1. sr_apply_changes() for the running datastore can fail during store
       operation because the cache could not be updated, because another
       thread is concurrently reading the operational datastore, even from
       potentially unrelated modules.
    2. cache is locked even when oper get subscribers are called to
       fetch operational data.

    - added a test for this scenario, and enabled running cache in
      test_multi_connection for increased testing.

###     modinfo ignore cache update fail during store

    A user facing process, such as gnmi, could be reading data and applying
    changes to the running datastore concurrently with multiple threads.

    A WRITE lock is needed to update the run cache, and may not always be
    available.

    "DONE_ONLY" subscribers can no longer be informed of these changes,
    because although sr_apply_changes() failed, the same changes can no
    longer be retried.

    It can also cause inconsistency when multiple modules are being changed,
    and the cache update fails after some modules were already written into,
    and remaining modules are left with previous data.

    To avoid such problems, it's best to ignore non-fatal problems such as
    cache update failures during sr_modinfo_data_store().

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -2902,6 +2902,12 @@ sr_modinfo_module_data_load(struct sr_mod_info_s *mod_info, struct sr_mod_info_m
     }
 
     if (run_cached_data_cur) {
+        /* CACHE READ LOCK */
+        if ((err_info = sr_rwlock(&conn->run_cache_lock, SR_CONN_RUN_CACHE_LOCK_TIMEOUT, SR_LOCK_READ, conn->cid,
+                __func__, NULL, NULL))) {
+            return err_info;
+        }
+
         /* there are cached data */
         switch (mod_info->ds) {
         case SR_DS_STARTUP:
@@ -2929,6 +2935,10 @@ sr_modinfo_module_data_load(struct sr_mod_info_s *mod_info, struct sr_mod_info_m
                     1, &mod_data);
             break;
         }
+
+        /* CACHE READ UNLOCK */
+        sr_rwunlock(&conn->run_cache_lock, SR_CONN_RUN_CACHE_LOCK_TIMEOUT, SR_LOCK_READ, conn->cid, __func__);
+
         if (err_info) {
             return err_info;
         }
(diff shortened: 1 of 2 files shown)


irfanHaslanded:
I suspect that the run_cache_lock is being held for too long, and can cause deadlocks, when more than 1 module is being read from 2 separate threads, and running datastore changes.

This can be reproduced with the failing test_multi_connection in this commit.
@michalvasko

michalvasko:
So what would you like to happen exactly? I do not think the dead lock can be avoided in this scenario, it creates it. The complexity and freedom of *sysrepo* and its API comes at a price and I am sure there are other use-cases that can result in similar problems and I do not think there is a fix except for forbidding some API combinations. But in this case, you can simply use `SR_OPER_NO_RUN_CACHED` in `test_oper_read_helper()`. The flag description even mentions dead locks.

irfanHaslanded:
I will think about this some more.

michalvasko (commented):
So essentially by adding more locks you solved the problems of the use-case you created?

irfanHaslanded:
Yes, there is more frequent lock and unlock, but overall locks are held for much shorter duration.

The important difference is that during `sr_module_oper_data_update()` and `sr_module_file_data_append()` called from `sr_modinfo_module_data_load()`, the run cache lock is unlocked. These two are the longer duration tasks I found in my use-cases.

Since, this lock is not process shared, but session shared, I don't think this should cause any problems elsewhere.

irfanHaslanded:
Also, the use-case in the test is not very complex IMHO
1. There is a slow (takes longer than 1 second) oper get provider.
2. Another process (connection) has 2 threads reading from the operational datastore of two modules.
3. User makes configuration changes to these two modules (to the running datastore).

michalvasko (approved):

michalvasko:
Okay then, seems fine, thanks.

PR #3595, opened 2025-04-21
compat use __atomic instead of legacy __sync func

irfanHaslanded:
According to [gcc](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html):

> These functions are implemented in terms of the `__atomic` builtins (see Built-in Functions for Memory Model Aware Atomic Operations). They should not be used for new code which should use the `__atomic` builtins instead.


michalvasko (approved):
Okay, it seems it was added in GCC 4.7, which is ancient.

PR #3598, opened 2025-04-23
Coverity fixes - mhr

MartinHerberg:
Fixes of issues from company internal Coverity check.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -657,6 +657,7 @@ sr_file_read(const char *path, char **data)
     sr_error_info_t *err_info = NULL;
     FILE *f = NULL;
     size_t offset, r, size;
+    long ret_ftell;
 
     *data = NULL;
 
@@ -669,8 +670,10 @@ sr_file_read(const char *path, char **data)
 
     /* learn its size */
     fseek(f, 0L, SEEK_END);
-    size = ftell(f);
+    ret_ftell = ftell(f);
     rewind(f);
+    SR_CHECK_INT_GOTO((ret_ftell < 0) || (((unsigned long) ret_ftell) > SIZE_MAX - 1), err_info, cleanup);
+    size = (size_t) ret_ftell;
 
     /* alloc return value */
     *data = malloc(size + 1);
--- a/src/common.c
+++ b/src/common.c
@@ -657,7 +657,6 @@ sr_file_read(const char *path, char **data)
     sr_error_info_t *err_info = NULL;
     FILE *f = NULL;
     size_t offset, r, size;
-    long ret_ftell;
 
     *data = NULL;
 
@@ -670,10 +669,8 @@ sr_file_read(const char *path, char **data)
 
     /* learn its size */
     fseek(f, 0L, SEEK_END);
-    ret_ftell = ftell(f);
+    size = ftell(f);
     rewind(f);
-    SR_CHECK_INT_GOTO((ret_ftell < 0) || (((unsigned long) ret_ftell) > SIZE_MAX - 1), err_info, cleanup);
-    size = (size_t) ret_ftell;
 
     /* alloc return value */
     *data = malloc(size + 1);
--- a/src/executables/sysrepocfg.c
+++ b/src/executables/sysrepocfg.c
@@ -919,7 +919,7 @@ main(int argc, char **argv)
             operation = opt;
             break;
         case 'd':
-            if (!optarg || arg_get_ds(optarg, &ds)) {
+            if (arg_get_ds(optarg, &ds)) {
                 goto cleanup;
             }
             break;
@@ -944,10 +944,6 @@ main(int argc, char **argv)
             xpath = optarg;
             break;
         case 'f':
-            if (!optarg) {
-                error_print(0, "Invalid optarg (NULL pointer)");
-                goto cleanup;
-            }
             if (!strcmp(optarg, "xml")) {
                 format = LYD_XML;
             } else if (!strcmp(optarg, "json")) {
@@ -969,10 +965,6 @@ main(int argc, char **argv)
             opaq = 1;
             break;
         case 'p':
-            if (!optarg) {
-                error_print(0, "Invalid optarg (NULL pointer)");
-                goto cleanup;
-            }
             max_depth = strtoul(optarg, &ptr, 10);
             if (ptr[0]) {
                 error_print(0, "Invalid depth \"%s\"", optarg);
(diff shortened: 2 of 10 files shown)


src/shm_ext.c
```diff
@@ -1920,7 +1920,8 @@ sr_shmext_oper_sub_check(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, const struct ly
     for (i = 0; i < shm_mod->oper_get_sub_count; ++i) {
         valid = 0;
         xp_subs = (sr_mod_oper_get_xpath_sub_t *)(conn->ext_shm.addr + subs[i].xpath_subs);
-        for (j = 0; j < subs[i].xpath_sub_count; ++j) {
+        j = 0;
+        if (j < subs[i].xpath_sub_count) {
```
michalvasko, line 1924: The code is formatted this way to be the same as the other functions for other subscriptions. I do not care that `coverity` complains about it.

src/shm_mod.c
```diff
@@ -220,7 +220,8 @@ sr_shmmod_fill(sr_shm_t *shm_mod, size_t shm_mod_idx, const struct lyd_node *sr_
             name = lyd_get_value(sr_child);
         } else if (!strcmp(sr_child->schema->name, "revision")) {
             /* copy revision */
-            strcpy(smod->rev, lyd_get_value(sr_child));
+            strncpy(smod->rev, lyd_get_value(sr_child), sizeof(smod->rev));
+            smod->rev[sizeof(smod->rev) - 1] = 0;
```
michalvasko, line 224: I think the old code is less confusing and cannot cause any issues, the node value must conform to the YANG restrictions. If you really want to satisfy `coverity`, you can add an assert on the length of the string before the `strcpy` call.

src/sysrepo.c
```diff
@@ -6599,6 +6599,7 @@ sr_change_ly2sr(const struct lyd_node *node, const char *value_str, const char *
     lyd_free_tree(node_dup);
     if (err_info) {
         if (sr_val) {
+            free(sr_val->origin);
```
michalvasko, line 6602: You can just put `sr_free_val` instead of lines 6601 - 6605.

src/utils/sn_common.c
```diff
@@ -1261,6 +1261,8 @@ srsn_create_timer(void (*cb)(void *arg, int *freed), void *arg, const struct tim
     sr_error_info_t *err_info = NULL;
     int r;
 
+    pthread_mutex_lock(&sntimer->lock);
```
michalvasko, line 1264: I do not understand the reason for these locks.

src/utils/subscribed_notifications.c
```diff
@@ -859,6 +859,11 @@ srsn_read_notif(int fd, const struct ly_ctx *ly_ctx, struct timespec *timestamp,
         goto cleanup;
     }
 
+    if (size > UINT32_MAX - 1) {
```
michalvasko, line 862: Same as what I wrote before, an `assert` is fine, a full check redundant and confusing.

michalvasko (changes requested):

src/common.c
```diff
@@ -669,8 +670,10 @@ sr_file_read(const char *path, char **data)
 
     /* learn its size */
     fseek(f, 0L, SEEK_END);
-    size = ftell(f);
+    ret_ftell = ftell(f);
     rewind(f);
+    SR_CHECK_INT_GOTO((ret_ftell < 0) || (((unsigned long) ret_ftell) > SIZE_MAX - 1), err_info, cleanup);
+    size = (size_t) ret_ftell;
```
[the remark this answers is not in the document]
MartinHerberg: I won't change it in that way, because later size is decreased by another size_t value (r) - which would be complained by Coverity again.

src/modinfo.c
```diff
@@ -2049,7 +2049,8 @@ sr_modinfo_module_srmon_datastore(sr_conn_ctx_t *conn, sr_mod_t *shm_mod, sr_dat
         goto cleanup;
     }
 
-    if ((ds_handle->plugin->last_modif_cb(ly_mod, ds, ds_handle->plg_data, &mtime) == 0) && (mtime.tv_sec > 0)) {
+    err_info = ds_handle->plugin->last_modif_cb(ly_mod, ds, ds_handle->plg_data, &mtime);
```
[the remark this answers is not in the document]
MartinHerberg: ok

src/plugins/ntf_json.c
```diff
@@ -114,6 +114,12 @@ srpntf_read_notif(int notif_fd, struct ly_ctx *ly_ctx, struct lyd_node **notif)
         goto cleanup;
     }
 
+    /* size sanity check */
```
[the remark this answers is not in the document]
MartinHerberg: ok

src/common.c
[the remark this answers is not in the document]
michalvasko: These are all such useless checks that do not solve any real issues. You can silence `coverity` however you want but I will not merge this change, it is redundant and hence can only confuse. I care more about code readability that some static code analysis tool issues.

src/shm_mod.c
MartinHerberg: ok

src/executables/sysrepocfg.c
```diff
@@ -919,7 +919,7 @@ main(int argc, char **argv)
             operation = opt;
             break;
         case 'd':
-            if (arg_get_ds(optarg, &ds)) {
+            if (!optarg || arg_get_ds(optarg, &ds)) {
```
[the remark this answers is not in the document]
michalvasko: Yes, because `I`, `X`, `E`, `R`, `N` are all defined as having optional argument, as seen in the string passed to `getopt_long`. Coverity obviously does not understand that but it is a false positive nonetheless, the code will not be changed.

src/sysrepo.c
MartinHerberg: `sr_free_val` is a function a looked for. Makes this piece of code much better. Thank you.

src/utils/sn_common.c
MartinHerberg: Coverity complains that the lock is taken if sntimer->trigger is written in other functions. E.g. `srsn_update_timer()`.

If you think, that the mutex usage here is pointless, we can remove it.

src/utils/subscribed_notifications.c
MartinHerberg: ok

src/common.c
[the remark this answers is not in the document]
MartinHerberg: I will drop this "fix"

src/utils/sn_common.c
MartinHerberg: removed

michalvasko: I got to look at it only now but I think it should be removed. Of course lock is taken in all other cases because in those the structure is accessed from several threads. In this case the structure is being prepared before any thread is accessing it.

michalvasko (approved):

PR #3599, opened 2025-04-24
shm_ext - use allocation table and reduce remap

irfanHaslanded:
### Brief description

There are some limitations of the current Ext SHM design:
1. It is difficult to recover if a process crashes while manipulating
   the hole linked-list and corrupts it.
2. grows SHM size strictly by actual necessary size requiring frequent
   remaps.
3. During system startup time when multiple processes are setting up
   subscriptions, and pushing operational data leading to lock-contention.
4. `ext_lock` is a *system-wide* mutex, and needs to be held everytime
   data is added or removed to the Ext SHM.

Some of these can be fixed by changing the Ext SHM to use an allocation
table instead of the hole linked-list to manage free space, and growing
the Ext SHM in bigger steps than what is strictly needed.

The underlying OS usually maps whole pages, so doing remap for every 40
or so bytes (a single subscription size) is not efficient.

 ### Allocation Table Parameters
The below parameters can be set via cmake flags at build time.
- Ext SHM allocation block size
  #define SR_SHMEXT_BLOCK_SIZE @SYSREPO_EXT_SHM_BLOCK_SIZE@

- Ext SHM Initial block count
  #define SR_SHMEXT_INIT_BLOCK_COUNT @SYSREPO_EXT_SHM_INIT_BLOCK_COUNT@

- Ext SHM Count of blocks to grow by, when out of free space
  #define SR_SHMEXT_GROW_BLOCKS @SYSREPO_EXT_SHM_GROW_BLOCKS@

 ### Preset defaults
1. `SR_SHMEXT_BLOCK_SIZE=64` is chosen as the basic allocation unit.
   When data needs to be stored in the Ext SHM, it will be stored in
   contiguous blocks of this size.

2. `SR_SHMEXT_INIT_BLOCK_COUNT=64` is chosen as the initial block
   count for the Ext SHM. All blocks except the first one are free to
   use. The first block is not used, to allow for easily check for
   `offset == 0` as empty/uninitialized data offset.

3. `SR_SHMEXT_GROW_BLOCKS=64` is chosen as the number of blocks to add
   each time available space is not enough. This means the Ext SHM will
   grow by 4KB, and prevent frequent remapping.

 ### Allocation Table
 The allocation table is located at the tail end of the Ext SHM. This
 allows for easy remapping as the existing data is never moved. Only the
 table is moved on remap.

```C
typedef ATOMIC16_T sr_shmext_block_usage_t;
typedef struct {
   pthread_mutex_t lock;
   uint32_t block_count;
   sr_shmext_block_usage_t used[];
} sr_shmext_table_t;
```
The number of blocks allocated including the current block are stored in
the `used` field. As this is 16 bits long, the maximum blocks which can
be allocated to a single chunk is 65535.

 ### Shrinking
 When 2 * SR_SHMEXT_GROW_BLOCKS or more free space exists in the Ext SHM
 at the tail end, Ext SHM is shrunk, to save space.

 ### Locking
 Ext SHM WRITE lock is only needed in two cases:
 1. Growing the Ext SHM, after the current space is exhausted.
 2. Debug printing Ext SHM contents (to prevent other threads from
	concurrently changing contents).

 In all other cases, such as adding/removing subscriptions or oper push
 data, only a READ lock is needed.

 To allow this, ext_lock is upgraded to a full read-write lock. If
 ext_lock is held, it is always held in the same mode as the connection
 ext remap lock.

- table lock
sr_shmext_try_reserve() can reserve a chunk without any locks, by using
atomic operations. But there is a data race possible, where two threads
concurrently try to reserve adjacent blocks, and neither succeeds.

For example, T1 and T2 want to reserve 3 blocks, and T1 grabs B1, and T2
grabs B2.

`B1, B2, B3, B4, B5, B6`
`T1, T2, B3, B4, B5, B6`
and then
`B1, T2, T1, B4, B5, B6`

To prevent this, a table lock is used, for reserving chunks.
table lock is not needed for releasing chunks.

 ### Advantages
1. Robust Ext SHM, not corrupted by linked-list after a crash.
2. Concurrent Ext SHM manipulation possible.
  - Only block reserve happens under exclusive mutex lock.
  - Unsubscribe, and storing data all happen concurrently.


irfanHaslanded:
For instance
With previous code, nearly 19% of the time is spent in `sr_shm_remap` for the tests.

```C
    18.86%     0.12%  test_multi_conn  libsysrepo.so.7.33.4  [.] sr_shm_remap
```

With this change
```C
     2.97%     0.05%  test_multi_conn  libsysrepo.so.7.33.4  [.] sr_shm_remap
```

With these changes remap expense is reduced to under 3%.

michalvasko:
Alright, briefly looking at the changes it seems fine. We never got to optimize this because it worked well enough. The benchmark is nice (despite the tests certainly remapping the SHM more often than during normal usage), can you also compare the memory allocated?

irfanHaslanded:
With these changes, from my `test_multi_connection`, the max ext SHM size was 51112.
Before these changes, the max ext SHM size was 58824.

But these seem to be varying based on fragmentation, and where holes are left.

I need to see a more realistic use-case, and how Ext SHM size varies there.

irfanHaslanded:
@michalvasko I can push an update, because I have a more recent version of this, if interested.

michalvasko:
Please do, I thought you have given up on this PR. If not, please finish it so that it can ultimately be merged. It is also fine if it takes some more time.

PR #3606, opened 2025-05-12
sub ntf BUGFIX allow recovery user on on-change notifications

jeremie6wind:
on-change notifications were sent only to users defined in the nacm rules. Grant access to the recovery user.


michalvasko (approved):
Thanks, seems correct.

michalvasko:
Just now checked and fixed another special case, when the NACM is turned off.

PR #3610, opened 2025-05-16
change_sub OPTIMIZE write event only if needed

irfanHaslanded:
## change_sub OPTIMIZE write event only if needed
When change subscribers are notified at a given priority, there
is a bottleneck of the sub_shm `SR_LOCK_WRITE_URGE` lock, for each
subscriber that processes the event, in order to update the progress of
the event processing.

Specifically, the following items are updated as relevant:
1. subscriber_count - remaining subscribers yet to process the event.
2. if there is data to be sent back (for an UPDATE/RPC event etc.)
3. event errors.
4. if there are no remaining subscribers, then final result.

The most common of these is scenario 1.
It is possible to use atomic operations to update `subscriber_count`,
while holding the sub_shm READ lock.

This allows for event processing to continue in parallel without
stopping all other subscribers, each time a subscriber finishes
processing.

For the final subscriber to process the event, the `WRITE` lock is
readily available, as there are no other readers.

---

### perf tests - add test for change_sub scale
There are no performance tests for changes being made with change
subscriptions present, which is the most common case.

### DIFF
--- a/src/shm_sub.c
+++ b/src/shm_sub.c
@@ -823,7 +823,7 @@ sr_shmsub_notify_write_event(sr_sub_shm_t *sub_shm, sr_cid_t orig_cid, uint32_t 
     ATOMIC_STORE_RELAXED(sub_shm->request_id, request_id);
     ATOMIC_STORE_RELAXED(sub_shm->event, event);
     ATOMIC_STORE_RELAXED(sub_shm->priority, priority);
-    sub_shm->subscriber_count = subscriber_count;
+    ATOMIC_STORE_RELAXED(sub_shm->subscriber_count, subscriber_count);
     sub_shm->operation_id = operation_id;
 
     /* remap if needed */
@@ -1752,7 +1752,7 @@ sr_shmsub_change_notify_change(struct sr_mod_info_s *mod_info, const char *orig_
                 continue;
             }
 
-            nsub->sub_shm->subscriber_count = subscriber_count;
+            ATOMIC_STORE_RELAXED(nsub->sub_shm->subscriber_count, subscriber_count);
             nsub->pending_event = 1;
             pending_events = 1;
         }
@@ -1946,7 +1946,7 @@ sr_shmsub_change_notify_change_done(struct sr_mod_info_s *mod_info, const char *
                 continue;
             }
 
-            nsub->sub_shm->subscriber_count = subscriber_count;
+            ATOMIC_STORE_RELAXED(nsub->sub_shm->subscriber_count, subscriber_count);
             nsub->pending_event = 1;
             pending_events = 1;
         }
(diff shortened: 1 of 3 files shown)


irfanHaslanded:
### **Test Summary**
With these changes, the number of futex syscalls seems to reduce by nearly 30%, and the number of unnecessary wakes (or errors) is reduced significantly.

---

Running only the `test_many_oper_change_subs` under strace, with below diff
```diff
$ git diff tests/
diff --git a/tests/perf.c b/tests/perf.c
index bc698d4b..142455ce 100644
--- a/tests/perf.c
+++ b/tests/perf.c
@@ -987,6 +987,8 @@ sysrepo_destroy(void)
 }

 struct test tests[] = {
+    {"many oper change_subs", setup_empty_oper, test_many_oper_change_subs, teardown_empty},
+#if 0
     {"get tree", setup_running, test_get_tree, teardown_running},
     {"get item", setup_running, test_get_item, teardown_running},
     {"get tree hash", setup_running, test_get_tree_hash, teardown_running},
@@ -1009,6 +1011,7 @@ struct test tests[] = {
     {"modify an item cached", setup_running_cached, test_item_modify, teardown_running},
     {"remove an item", setup_running, test_item_remove, teardown_running},
     {"remove an item cached", setup_running_cached, test_item_remove, teardown_running},
+#endif
 };
```
---
## with these changes 0.467227 s (60% of system calls time spent in futex)

```sh
$ make -j sr_clean test_clean all > /dev/null 2>&1 && strace -c  ./tests/sr_perf 100 1 2>&1 | grep -v '0.00'

| Options

  Data set size      : 100
  Each test executed : 1 time


| Performance tests

  JSON DS file
 ___________________________________________________________________
|                                                                   |
| test name                          |         time |    comparison |
|___________________________________________________________________|
|                                                                   |
| many oper change_subs ............ |   0.467227 s |       1.000 x |
|___________________________________________________________________|


All comparisons refer to how many times faster (green) or slower (red) the current plugin is compared to the first plugin.

 % time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 59.86    0.066825          55      1194         4 futex
------ ----------- ----------- --------- --------- ----------------

```
---
## without these changes 0.606427 s (84% of system calls time spent in futex)

```sh
$ make -j sr_clean test_clean all > /dev/null 2>&1 && strace -c  ./tests/sr_perf 100 1 2>&1 | grep -v '0.00'

| Options

  Data set size      : 100
  Each test executed : 1 time


| Performance tests

  JSON DS file
 ___________________________________________________________________
|                                                                   |
| test name                          |         time |    comparison |
|___________________________________________________________________|
|                                                                   |
| many oper change_subs ............ |   0.606427 s |       1.000 x |
|___________________________________________________________________|


All comparisons refer to how many times faster (green) or slower (red) the current plugin is compared to the first plugin.

 % time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 84.16    0.195312         107      1825       205 futex
------ ----------- ----------- --------- --------- ----------------
```

michalvasko (approved):
Nice and makes sense, thanks.

PR #3615, opened 2025-05-21
rwlock BUGFIX dead reader deletion

irfanHaslanded:
a dead recursive reader must be completely removed and not just the `read_count` decremented, because it could have held the lock recursively.


michalvasko (approved):
Right, thanks.

PR #3616, opened 2025-05-23
properly recover from process crash during READ sr_rwunlock()

irfanHaslanded:
### Brief
simplify the complex memmove logic for reader_del to keep readers and
read_count arrays consistent.

 ### Problem description
Although memmove() is MT-safe, it is not an atomic operation.
Also if the process crashes after memmove() of `readers` but before
memmove() `read_count`, the rwlock is not recoverable.

 ### Solution
Swap the last reader with the reader being deleted. This means that
there are only 2 swaps necessary to complete the reader deletion.

With this, there is only one data race possible:
The last reader CID can be present in duplicates, at the location of the
just deleted reader and at the end, concurrently.

But this is easily recoverable as implemented in
sr_rwlock_reader_recover().

 ### Example:
reader_cid(read_count) [2(1), 7(1), 4(1), 5(3)]

CID 7 unlocks with the following steps:

Step 1. decr read_count
  [2(1), 7(0), 4(1), 5(3)]

Step 2. store last CID read_count under CID 7
  [2(1), 7(3), 4(1), 5(3)]

Step 3. change CID 7 to CID 5
  [2(1), 5(3), 4(1), 5(3)]

Step 4. zero last CID
  [2(1), 5(3), 4(1), 0(3)]

Step 5. zero last CID count
  [2(1), 5(9), 4(1), 0(0)]

Specifically a duplicate reader at the end of the readers array is
present if process CID 7 dies just after Step 3. So, as part of
recovering the rwlock, we should also check for this condition and
handle it properly.

Also, it is important to maintain the correct order while deleting the
CID from the readers list.
For example, if reader CID is updated before read_count, we may be left
with
  [2(1), 5(0), 4(1), 5(3)]
after a crash, which is not easy to recover as CID 5 is alive, but it's
read_count is reported as 0.

## modinfo BUGFIX duplicate reader in srmon ds_lock

There is a data race with unlocked access of rwlock data for populating
sysrepo monitoring module lock data in
`sr_modinfo_module_srmon_locks_ds()`.

To accurately report locks in sysrepo monitoring, The only correct
solution would be to quickly cache the entire readers array with the
rwlock mutex held and unlock, and then construct the sysrepo-monitoring
tree using the cached data.

But even this comes with the problem of having to acquire and release a
lot of locks, if there are a lot of modules, and is not ideal.
Also, if any lock is WRITE locked, then it leads to delays.

So, the simplest solution is to ignore duplicate readers
(which can happen rarely) by passing the `LYD_NEW_PATH_UPDATE`
flag to `sr_lyd_new_path()`.

## add tests for rwlock reader limit and recovery

There are no tests covering the scenarios where a rwlock readers list is
full.

If the readers array gets corrupted, it can make the system unusable.
Adding these tests now to cover this gap.

also add a test for sr_unsubscribe retry on first lock acq failure if
the callback thread was busy.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -2255,9 +2255,7 @@ cleanup:
 static void
 sr_rwlock_reader_del_(sr_rwlock_t *rwlock, uint32_t i, int dead)
 {
-    uint32_t move_count;
-    sr_cid_t last_reader;
-    uint8_t last_read_count;
+    uint32_t j;
 
     /* there should be at least 1 read_count or it must be a dead CID */
     assert(rwlock->read_count[i] || dead);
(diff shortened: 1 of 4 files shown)


michalvasko (approved):

michalvasko:
Thanks, seems fine.

PR #3657, opened 2025-08-07
Adding a couple of dir macros

mruprich:
To avoid leaving some directories behind when the package is removed, each directory in the owned path should be explicitly named in the %files section as explained here:

https://docs.fedoraproject.org/en-US/packaging-guidelines/UnownedDirectories/


michalvasko:
Looks fine, thanks.

PR #3658, opened 2025-08-07
Coverity_fixes_jsc-se

jsc-se:
Fix for Coverity issues (company internal).


michalvasko:
Fine, but this is a false-positive report, those should not be catered to in the code.

Issue #3688, opened 2025-10-21
Segfault in sr_shmext_notif_sub_remove_dead


michalsieron:
(this issue most probably obsoletes https://github.com/CESNET/netopeer2/issues/1762. At least the backtrace is now correct)

## Used versions
netopeer2: 2.4.1
libnetconf2: 3.7.1
sysrepo: 3.6.11
libyang: 3.12.2

Cross-compiled on x86-64 for aarch64.

## What happened
`netopeer2-server` segfaults in libsysrepo right after starting.

Below is the backtrace:
```
(gdb) bt
#0  0x0000007f81844130 in sr_shmext_notif_sub_remove_dead (conn=<optimized out>, shm_mod=<optimized out>) at /usr/src/debug/sysrepo/3.6.11/src/shm_ext.c:1318
#1  sr_shmext_notif_sub_add (conn=conn@entry=0x557c7a2b50, shm_mod=shm_mod@entry=0x7f80c085c8, sub_id=sub_id@entry=82, xpath=xpath@entry=0x0, evpipe_num=11) at /usr/src/debug/sysrepo/3.6.11/src/shm_ext.c:1340
#2  0x0000007f8180b0a4 in _sr_notif_subscribe (session=session@entry=0x557ca6d880, mod_name=<optimized out>, xpath=0x0, start_time=0x0, stop_time=stop_time@entry=0x0, callback=callback@entry=0x0, tree_callback=tree_callback@entry=0x7f818646b0 <srsn_sn_rpc_subscribe_cb>, private_data=private_data@entry=0x557ca716c0, opts=opts@entry=0, subscription=subscription@entry=0x557ca716d0) at /usr/src/debug/sysrepo/3.6.11/src/sysrepo.c:7577
#3  0x0000007f81818850 in sr_notif_subscribe_tree (session=session@entry=0x557ca6d880, module_name=<optimized out>, xpath=<optimized out>, start_time=<optimized out>, stop_time=stop_time@entry=0x0, callback=callback@entry=0x7f818646b0 <srsn_sn_rpc_subscribe_cb>, private_data=private_data@entry=0x557ca716c0, opts=opts@entry=0, subscription=subscription@entry=0x557ca716d0) at /usr/src/debug/sysrepo/3.6.11/src/sysrepo.c:7638
[...]

PR #3689, opened 2025-10-21
shm_ext: Fix potential segfault

michalsieron:
In recent change (#3509) in shm_ext.c, some functions started acquiring ext_lock. In couple of them adaptation was required, which moved reading of `conn->ext_shm.addr` into the block where lock is already acquired.

It seems sr_shmext_notif_sub_remove_dead was missed and there is now a risk of segfault.

Simply moving the calculation of `notif_subs` address should be enough to fix it.

Should fix: #3688

### DIFF
--- a/src/shm_ext.c
+++ b/src/shm_ext.c
@@ -1306,7 +1306,7 @@ sr_shmext_notif_sub_remove_dead(sr_conn_ctx_t *conn, sr_mod_t *shm_mod)
 {
     uint32_t i = 0;
     sr_error_info_t *err_info = NULL;
-    sr_mod_notif_sub_t *notif_subs = (sr_mod_notif_sub_t *)(conn->ext_shm.addr + shm_mod->notif_subs);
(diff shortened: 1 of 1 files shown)


michalvasko (approved):

michalsieron:
@michalvasko kinda related question. What is the reason for `sr_shm_remap` to remap even when called with `new_shm_size=0` (return current size)? I am not really familiar with SHM related stuff, so I don't know what are use cases for remapping with the same size and other parameters.

michalvasko:
> What is the reason for sr_shm_remap to remap even when called with new_shm_size=0 (return current size)?

I am not sure I understand but generally the SHM segments can be changed by any processes so they must always check it has not been changed and fix their own mapping, in case they were.

> I am not really familiar with SHM related stuff

Then my well-meant advice is to stay away :)

michalsieron:
>the SHM segments can be changed by any processes so they must always check it has not been changed and fix their own mapping

yeah, that does make sense :+1:

PR #3696, opened 2025-10-30
sr_lycc_store_context write to filename not fd

irfanHaslanded:
# sr_lycc_store_context write to filename not fd

Context lock has become significantly more complex when a printed
context is in use with caches.
The caches in question are:
1. running datastore cache
2. oper poll subscription cache
3. oper push data cache (in the sessions).

We need to track all the users of the three caches.

The main problem is that caches held in all processes and connections
need to be flushed before the context can change.
There is only one way to flush all the caches - have a
`flush_cache_rpc` and a thread to process it.

 ### Printed context
Printed context is written into `sr_ctx` SHM file, and accessed via a
file descriptor. The same filename is used by every process to directly
reuse the printed context.

 ### Context switch
A system-wide WRITE context lock is acquired, stopping all other
operations. The open file descriptor is overwritten with new content.
So, the old context immediately disappears for all processes, and old
caches can no longer be flushed.

 ### Solution
To prevent this, the new context is written into a temporary SHM file
and renamed to `sr_ctx`. The old fd (and old context) is still available
to all other processes. When a new context lock is acquired, other
processes detect the changed context, and first flush their caches, and
then close the old context fd, and open the new `sr_ctx` file.

Once all processes close the fd for the old context file, it will be
removed from memory by the operating system.

---

## context update track caches privately
sr_shmmain_oper_push_cache_flush() is used to flush the process oper
cache, avoiding the need for complex CID, PID tracking before updating
the shared libyang context.

---

## sr_conn_list_s track connection instead of CID

Tracking connection pointers instead of CID allows us to fetch other
information in the connection, such as the oper caches, which is needed
to flush them on context change.

---

## hold context lock correctly for oper ds changes
context lock is missing in some cases, and in other cases being unlocked
too soon, causing occasional test failures.

--- 
 ## add test for ly_ctx change races
Add a test which changes the context in one thread by constantly installing and removing modules, and another thread which uses oper push data caches.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -538,10 +538,11 @@ sr_error_info_t *sr_path_mod_shm(char **path);
 /**
  * @brief Get the path of the ctx SHM.
  *
+ * @param[in] tmp if a temporary filename should be created.
  * @param[out] path Created path. Should be freed by the caller.
  * @return err_info, NULL on success.
  */
-sr_error_info_t *sr_path_ctx_shm(char **path);
+sr_error_info_t *sr_path_ctx_shm(int tmp, char **path);
 
 /**
  * @brief Get the path of the external SHM.
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -316,7 +316,7 @@ sr_connect(const uint32_t opts, sr_conn_ctx_t **conn_p)
 
         /* first connection on the system, no need to hold a write lock,
          * print the context for other connections to reuse */
-        sr_lycc_store_context(conn, &sr_yang_ctx.ly_ctx_shm, sr_yang_ctx.ly_ctx);
+        sr_lycc_store_context(&sr_yang_ctx.ly_ctx_shm, sr_yang_ctx.ly_ctx);
     }
 
     SR_LOG_INF("Connection %" PRIu32 " created.", conn->cid);
@@ -1609,7 +1609,7 @@ _sr_install_modules(sr_conn_ctx_t *conn, const char *search_dirs, const char *da
     }
 
     /* perform the update of datastore and SR data, update schema-mount contexts */
-    if ((err_info = sr_lycc_update_data(conn, &cc_info, &sr_yang_ctx.mod_shm, &sr_run_cache, &sr_oper_cache))) {
+    if ((err_info = sr_lycc_update_data(conn, &cc_info, &sr_yang_ctx.mod_shm))) {
         goto error;
     }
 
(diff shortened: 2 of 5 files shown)


src/common.h
```diff
@@ -538,10 +538,11 @@ sr_error_info_t *sr_path_mod_shm(char **path);
 /**
  * @brief Get the path of the ctx SHM.
  *
+ * @param[in] tmp if a temporary filename should be created.
```
michalvasko: Wrong param order

irfanHaslanded:
> There is still a race which is causing an occasional failure in unit-tests:
> 
> ```c
> test_multi_connection: /home/irfan/opensource/sysrepo/src/modinfo.c:4053: sr_modinfo_push_oper_mod_update_cache: Assertion `!sess->oper_push_mods[i].cache' failed.
> 539.019525 [3248480][140737347475136][DBG]: Applying "running" datastore changes success.
> 539.019500 [3248480][140737339082432][INF]: Connection 6 destroyed.
> [Thread 0x7ffff71a46c0 (LWP 3248484) exited]
> 
> Thread 4 "test_multi_conn" received signal SIGABRT, Aborted.
> [Switching to Thread 0x7ffff69a36c0 (LWP 3248485)]
> __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
> 44      ./nptl/pthread_kill.c: No such file or directory.
> A debugging session is active.
> ```
> 
> checking the contents of the cache, it looks like it is valid.
> 
> ```
> <l1 xmlns="urn:test" xmlns:or="urn:ietf:params:xml:ns:yang:ietf-origin" or:origin="or:intended">
>   <k>key2</k>
>   <v or:origin="or:unknown">1</v>
> </l1>
> ```
> 
> `sr_modinfo_get_oper_data()` is not consuming the cache as expected. I suspect this is happening because `mod->shm_mod->oper_push_data_count` or `shm_mod` is in a race here. Still investigating.

This should be fixed with https://github.com/sysrepo/sysrepo/pull/3698

src/sysrepo.c
```diff
@@ -5164,6 +5191,9 @@ _sr_discard_oper_changes(sr_session_ctx_t *session, const char *module_name, int
         sr_modinfo_erase(&mod_info);
         memset(&mod_info, 0, sizeof mod_info);
 
+        /* CONTEXT READ UNLOCK to allow upgrade */
+        sr_lycc_unlock(session->conn, SR_LOCK_READ, 0, __func__);
+
```
michalvasko: Invalid extra code.

michalvasko (approved):

michalvasko:
There were issues with CONTEXT locks, I tried to simplify and fix them.

PR #3698, opened 2025-11-01
oper push data BUGFIX session stop

irfanHaslanded:
## Background
Information about sessions which have push operational data for a module
is stored in the Ext SHM. This is not removed until the session is
stopped for performance. Only the `has_data` flag is updated if
the session has no data, as it's likely that it may push new data again.

When a session pushes operational data to a module, it remembers this in
session->oper_push_mods array and read by sr_modinfo_collect_oper_sess().

 ## Locking for sr_apply_oper_changes() during sr_session_stop()
oper push data is protected by shmmod data locks. a WRITE lock is needed
to remove it from Ext SHM by calling `sr_shmext_oper_push_del()`.

Normally `sr_modinfo_collect_oper_sess()` is called to learn which
modules should be added into mod_info for locking.
This skips any modules without any session push data, because there are
no changes for them.
So, these modules are not locked.
This breaks sr_shmext_oper_push_del() which expects module WRITE lock.

 ## Solution
For sr_session_stop(), all modules in the oper_push_mods must be
collected, even if they have no data so that they can be properly
locked, and Ext SHM entries removed.

### DIFF
--- a/src/modinfo.c
+++ b/src/modinfo.c
@@ -323,7 +323,7 @@ cleanup:
 }
 
 sr_error_info_t *
-sr_modinfo_collect_oper_sess(sr_session_ctx_t *sess, const struct lys_module *ly_mod, struct sr_mod_info_s *mod_info)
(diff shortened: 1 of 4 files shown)


michalvasko (approved):
It seems fine and I trust that you know what you are doing looking into all this.

PR #3699, opened 2025-11-03
use oper_edit_fetched flag to check staged changes

irfanHaslanded:
For `sr_edit_batch()` and `sr_discard_oper_changes()`, presence of `session->dt[session->ds].edit` was being checked to verify that there aren't already any staged changes.

With operational data, this is not true. For example if all pushed data are being deleted, `session->dt[session->ds].edit` is NULL. The correct way to check it is by using the flag
`session->oper_edit_fetched`.

All other data edit functions already use this flag correctly.

### DIFF
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -4145,7 +4145,7 @@ sr_edit_batch(sr_session_ctx_t *session, const struct lyd_node *edit, const char
         }
     }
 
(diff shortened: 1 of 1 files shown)


irfanHaslanded:
I think I have found the reason for the test failure, https://github.com/sysrepo/sysrepo/pull/3700

PR #3700, opened 2025-11-03
tests BUGFIX apply changes locked write timing

irfanHaslanded:
when test_apply_changes is run under valgrind, there is risk of starvation of the thread calling sr_lock().
Use a barrier wait to let both threads start, and sleep in the unlocked thread each iteration to allow the locking thread to acquire the lock more frequently.


michalvasko (approved):
Fine, I have already reduced the number of iterations to prevent the test failures.

PR #3709, opened 2025-11-20
skip context lock in subscription event callbacks

irfanHaslanded:
subscription event callbacks are processed by calling `sr_subscription_process_events()`, which already acquires a context lock for the entire duration.

callbacks usually need to acquire context to construct libyang data trees, so additional recursive context lock is acquired when not needed.

More importantly, it prevents application code losing track of acquired context locks, and missing to unlock when errors occur.

### DIFF
--- a/src/sysrepo.c
+++ b/src/sysrepo.c
@@ -487,6 +487,11 @@ sr_session_acquire_context(sr_session_ctx_t *session)
         return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


michalvasko (approved):
Well, okay, but I am not sure there is much value in optimizing the context lock. But since in this case it is simple enough, fine.

PR #3716, opened 2026-01-05
tests: Format time_t as long long

bgermann:
time_t can be 64 bits wide on a 32 bit system (see Y2k38 problem). Format time_t as long long instead of long to address this.


michalvasko:
Thanks, should be fine, I have just changed the target branch to `devel`, we do not accept PRs into `master`. It will be merged on the next release.

Issue #3346, opened 2024-07-09
support SYS_futex in riscv32 arch


satomhxl:
Hello, I'm testing sysrepo-2.10.1 in riscv32 environment recently, and I found a problem in compilation.
The riscv32 architecture doesn't support SYS_futex system call.
```
| /build/tmp/work/riscv32-poky-linux/sysrepo/2.10.1+gitAUTOINC+7a8ebf3523-r0/git/src/sr_cond/sr_cond_futex.c: In function 'sys_futex_wait':
| /build/tmp/work/riscv32-poky-linux/sysrepo/2.10.1+gitAUTOINC+7a8ebf3523-r0/git/src/sr_cond/sr_cond_futex.c:72:20: error: 'SYS_futex' undeclared (first use in this function); did you mean 'SYS_tee'?
|    72 |     return syscall(SYS_futex, uaddr, futex_op, expected, timeout, NULL, FUTEX_BITSET_MATCH_ANY);
|       |                    ^~~~~~~~~
|       |                    SYS_tee
| /build/tmp/work/riscv32-poky-linux/sysrepo/2.10.1+gitAUTOINC+7a8ebf3523-r0/git/src/sr_cond/sr_cond_futex.c:72:20: note: each undeclared identifier is reported only once for each function it appears in
| /build/tmp/work/riscv32-poky-linux/sysrepo/2.10.1+gitAUTOINC+7a8ebf3523-r0/git/src/sr_cond/sr_cond_futex.c: In function 'sys_futex_wake':
| /build/tmp/work/riscv32-poky-linux/sysrepo/2.10.1+gitAUTOINC+7a8ebf3523-r0/git/src/sr_cond/sr_cond_futex.c:85:20: error: 'SYS_futex' undeclared (first use in this function); did you mean 'SYS_tee'?
|    85 |     return syscall(SYS_futex, uaddr, FUTEX_WAKE, waiter_count, NULL, NULL, 0);
|       |                    ^~~~~~~~~
|       |                    SYS_tee
```

And I can pass the compilation if I change SYS_futex to SYS_futex_time64, as said in [this link](https://lore.kernel.org/all/20210910043102.3616198-2-alistair.francis@opensource.wdc.com/).
Can you consider to fix this issue?

Thank you very much.

jktjkt:
I'm not the maintainer, but I think that the usual "patches welcome" principle applies here, especially if you have access to that platform.

michalvasko:
[...]

PR #3737, opened 2026-02-13
cmake UPDATE use sr_cond_futex only if SYS_futex syscall is available

BlockoS:
This is a follow-up to https://github.com/sysrepo/sysrepo/issues/3346

Checking for the presence of `linux/futex.h` does not guarantee that futexes can be used.
That is why an extra check for `SYS_futex` symbol was added.


michalvasko (approved):

michalvasko:
No problem, I am just wondering what platform has a Linux futex and no syscalls.

BlockoS:
As far as I can tell, some RISC-V 32 emulators/simulators (qemu_riscv32 and spike_riscv32).

PR #3743, opened 2026-03-10
cmake UPDATE specify sysrepo as C project

flxzt:
I noticed that sysrepo requires a CXX compiler.
Since it is a C based project, only a C compiler is needed.
This updates the CMake config to reflect this.


michalvasko (approved):
Right, thanks.

PR #3770, opened 2026-06-03
Sysrepo notifd example + fixes

Roytak:


examples/notifd/notifd_client.c
```diff
@@ -0,0 +1,720 @@
...
+
+    mt_name = (hdr->media_type < 4) ? media_type_names[hdr->media_type] : "unknown";
+
+    printf("\n--- Notification ---\n");
+    printf("  Path:          %s\n", path);
+    printf("  Publisher ID:  %" PRIu32 "\n", hdr->publisher_id);
+    printf("  Message ID:    %" PRIu32 "\n", hdr->message_id);
+    printf("  Media Type:    %s\n", mt_name);
+
+    if (!lyd_print_mem(&payload, notif, LYD_JSON, LYD_PRINT_SHRINK)) {
```
michalvasko: Why use `LYD_PRINT_SHRINK`? Makes it less human-readable.

examples/notifd/notifd_client.c
```diff
@@ -0,0 +1,720 @@
...
+ * @brief Pending message being reassembled from segments.
+ */
+typedef struct {
+    uint32_t publisher_id;      /**< publisher ID */
+    uint32_t message_id;        /**< message ID */
+    uint8_t media_type;         /**< media type from first segment */
+    segment_buffer_t *segments; /**< array of segment buffers */
+    uint16_t total_segments;    /**< total segments (0 if unknown) */
+    uint16_t received_count;    /**< number of received segments */
+    time_t first_received;      /**< timestamp of first segment */
```
michalvasko: As discussed before, using `time_t` often causes possible issues while `struct timespec` fixes those. In this case I can imagine a DoS by sending lots of segmented messages in 1 second that will cause local resources to be used up.

examples/notifd/README.md
```diff
@@ -0,0 +1,125 @@
...
+    </receivers>
+  </subscription>
+</subscriptions>
+```
+
+This example XML data is stored in `example_sn_data.xml`. To apply it to the running
+datastore using `sysrepocfg`:
+
+```bash
+$ sysrepocfg -I ../examples/notifd/example_sn_data.xml
```
michalvasko: `-I` replaces the whole datastore with the new configuration. Using `-E` merges it into the existing configuration, so that is what you want, I think.

michalvasko (changes requested):

michalvasko (approved):

PR #3775, opened 2026-07-10
sysrepo bugfix and add docker setup for DS plugins

Humblesaw:
Size of the modules structure is padded from 84 to 88 bytes. During memcpy into the internal structure, these padding bytes overwrite the is_schema_yang field. The subsequent detection only sets the field to 1, never to 0, so a non-zero padding value causes file paths to be misinterpreted as inline YANG data. Explicitly set all the fields to
avoid unwanted behavior.


jktjkt (changes requested):
I would prefer that instead of relying on `memcpy` and some out-of-band knowledge that these structs should "look similar to each other", to simply copy the fields which are relevant. That way, the code is much more robust.

src/sysrepo.c
```diff
@@ -1839,12 +1839,16 @@ sr_install_modules2(sr_conn_ctx_t *conn, const sr_install_mod_t *modules, uint32
 
     /* copy all the items */
     for (i = 0; i < module_count; ++i) {
-        memcpy(&new_mods[i], &modules[i], sizeof *modules);
+        new_mods[i].schema_path = modules[i].schema_path;
```
michalvasko, line 1844: Mention why `memcpy` is not used.

Humblesaw: Done

README.md
```diff
@@ -353,6 +353,10 @@ To enable authentication via a username and password, set `REDIS_USERNAME` and `
 create a corresponding user with sufficient rights, and do not forget to enforce the authentication on the server (see [official Redis documentation](https://redis.io/docs/latest/commands/auth/)).
 For more information on how the plugin works, please refer to the [plugin documentation](doc/sr_plugins.dox).
 
+### Docker setup
```
michalvasko: Please put into a separate README in the `docker` directory.

src/sysrepo.c
michalvasko: The issue is not that the structures have a different size (obviously, since one has much more members) but that even though the structures are compatible, padding makes them actually incompatible.

michalvasko (changes requested):

README.md
Humblesaw: Done

src/sysrepo.c
Humblesaw: Done

src/sysrepo.c
```diff
@@ -1839,12 +1839,18 @@ sr_install_modules2(sr_conn_ctx_t *conn, const sr_install_mod_t *modules, uint32
...
+        new_mods[i].module_ds = modules[i].module_ds;
+        new_mods[i].owner = modules[i].owner;
+        new_mods[i].group = modules[i].group;
+        new_mods[i].perm = modules[i].perm;
 
         /* detect schema format */
-        if (strchr(modules[i].schema_yang, '{')) {
-            new_mods[i].is_schema_yang = 1;
-        }
+        new_mods[i].is_schema_yang = strchr(modules[i].schema_yang, '{') ? 1 : 0;
```
michalvasko: Please remove the change, it is not needed after the other changes.

michalvasko (changes requested):

michalvasko (approved):

PR #3791, opened 2026-08-15
log UPDATE configurable log level for logging callback

fredrik-persson:
A callback set with sr_log_set_cb() was called for every message with no way to say which levels it wants. sr_shmext_print() also treats a registered callback as a request for debug output, so it built a full ext SHM dump on every subscription change even when the callback discarded it, and sr_log()/srplg_log() formatted every message before the level was checked.

Add sr_log_set_cb_level()/sr_log_get_cb_level() and skip formatting messages no destination wants. Defaults to SR_LL_DBG, so existing callers are unaffected.

Rationale: we register a callback that ignores SR_LL_DBG, but have to pay the cost of building every debug message anyway (mostly noticeable in the ext SHM dumps). Found it in an AddressSanitizer build, but it costs without sanitizers too.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -37,6 +37,7 @@ sr_log_level_t sr_stderr_ll = SR_LL_NONE;   /**< stderr log level */
 sr_log_level_t sr_syslog_ll = SR_LL_NONE;   /**< syslog log level */
 int syslog_open;                            /**< Whether syslog was opened */
 sr_log_cb sr_lcb;                           /**< Logging callback */
+sr_log_level_t sr_lcb_ll = SR_LL_DBG;       /**< logging callback log level */
 
 /**
  * @brief String error list.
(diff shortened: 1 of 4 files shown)


src/log.h
```diff
@@ -52,6 +53,17 @@ extern sr_log_cb sr_lcb;             /**< logging callback */
...
+/**
+ * @brief Check whether any log destination is interested in a message of some level.
+ *
+ * Mirrors the delivery conditions in ::sr_log_msg(), letting callers skip work
+ * (formatting a message, collecting data to print) that would be discarded.
+ *
+ * @param[in] ll Message log level.
+ * @return Whether the message would be logged anywhere.
+ */
+int sr_log_ll_wanted(sr_log_level_t ll);
```
michalvasko: Please rename to `sr_log_msg_hidden()` and will obviously return a negated value.

michalvasko (changes requested):
Okay, thanks for the contribution, just a minor issue. Once fixed, it can be merged.

src/log.c
```diff
@@ -150,7 +157,7 @@ sr_log_msg(int plugin, sr_log_level_t ll, const char *msg)
     }
 
     /* logging callback */
-    if (sr_lcb) {
+    if (sr_lcb && ll <= sr_lcb_ll) {
```
michalvasko: Separate boolean expression needs to be in parentheses for formatting to pass tests.

michalvasko (approved):

PR #3795, opened 2026-08-25
Notifd fixes + srsn new API

Roytak:
Fix multiple issues in sysrepo-notifd and introduce new API `srsn_read_dispatch_del` to avoid data races with SRSN dispatch thread.


src/utils/sn_common.c
```diff
@@ -1162,6 +1173,41 @@ srsn_dispatch_add(int fd, void *cb_data)
     return err_info;
 }
 
+sr_error_info_t *
+srsn_dispatch_del(int fd)
```
michalvasko, line 1177: Should probably generate an error if `fd` is not found.

michalvasko (changes requested):

src/utils/sn_common.c
Roytak: The FD not being found is the expected outcome of correct usage here, not a caller error. Clarified this in the docs.

michalvasko (approved):

michalvasko (approved):

PR #3804, opened 2026-09-01
Refactor test_notifd + NCC notif before SR_EV_DONE

Roytak:
Refactor test_notifd's harness and racy tests.

`netconf-config-change` notification is now sent before triggering the `SR_EV_DONE` event for the subscribers.
Therefore other subscribers now see `netconf-config-change` published slightly earlier, before change-done callbacks in other processes return. Sysrepo never guaranteed ordering between a notification delivered to one process and a change callback in another, but still this is a behavioral change.
Motivation: avoid `subscription-started` + `netconf-config-change` emitted by `sysrepo-notifd` to the receiver being created.


src/sysrepo.c
```diff
@@ -4706,17 +4706,23 @@ sr_changes_notify_store(struct sr_mod_info_s *mod_info, sr_session_ctx_t *sessio
...
     }
 
-    /* publish "done" event, all changes were applied */
-    if ((err_info = sr_shmsub_change_notify_change_done(mod_info, sid, orig_name, orig_data, timeout_ms, err_info2))) {
-        goto cleanup;
+    /* generate the netconf-config-change notification before the "done" event, the changes are
+     * already committed at this point; otherwise a notification subscription created by a "done"
+     * callback would receive this change, which predates it */
+    if (session) {
+        notif_err_info = sr_modinfo_generate_config_change_notif(mod_info, session);
```
michalvasko: I have used `tmp_err` for this purpose before and no need to put the merge into cleanup, it can be right after this function.

tests/test_notifd.c
```diff
@@ -604,119 +641,249 @@ receive_notification_ext(int sockfd, const struct ly_ctx *ly_ctx, int timeout_ms
...
 
+/**
+ * @brief Check whether the test is running under valgrind.
+ *
+ * @return 1 if it is, 0 otherwise.
+ */
 static int
-receive_notification(int sockfd, const struct ly_ctx *ly_ctx, struct lyd_node **notif,
-        udp_notif_header_t *header)
+running_with_valgrind(void)
```
michalvasko, line 656: Is this really necessary? Why not always use the higher timeout?

michalvasko (changes requested):

tests/test_notifd.c
Roytak: Done for the waits that expect something - those return as soon as the notification arrives. The scaling stays only on `QUIET_TIMEOUT_MS`, which is the opposite case: it asserts that *nothing* arrives, so the whole window is consumed on every run. There are ~65 of those per run and scaling them unconditionally takes `test_notifd` from 25 s to ~117 s.

src/sysrepo.c
Roytak: Done, but had to use `tmp_err` for both the calls to avoid a leak.

